PR #13, opened 2026-06-10
m0: deflake auth-lookup p95 and wake-gap tests (M0-3)

Augustas11:
## Summary

Milestone 0 M0-3 from `audits/2026-06-10/REPO_AUDIT.md` — closes audit finding around two tests that burn real wall-clock and flake under parallel CI load (M0-3a and M0-3b in the audit punch list).

### M0-3a — gateway sqlite

`TestAuthLookupP95UnderOneMillisecondWith10KKeys` asserted a hard 1ms p95 ceiling that observed 13ms under parallel load. `raceEnabled` skip was not enough — even on a quiet machine the test is a perf characterization, not a correctness check.

Conversion:
- Rename → `BenchmarkAuthLookupWith10KKeys`, body takes `*testing.B`, uses `b.N` and `b.ReportMetric` to publish `p95_ms`. Run with `go test -bench=BenchmarkAuthLookup -run=^$ ./internal/storage/sqlite/`.
- The structural guarantee that the hash index exists is preserved as `TestAPIKeyHashIndexExists` (was already a separate test in this file).
- `newTestStore` / `createAccount` widened from `*testing.T` → `testing.TB` so a `*testing.B` can reuse them.

### M0-3b — coordinator ws

`TestWakeGapSendsWarmUpAndMarksDegraded` had a 1100ms `time.Sleep` paired with a 1-second `WakeGapThresholdS` and a default 2s `eventually()` window. Audit verifier note: the real flake risk is server-side goroutine starvation between the resume heartbeat and pool-state propagation, not the sleep length.

Fix:
- Added a test-only `WakeGapThresholdMs` config field. The threshold formerly only accepted whole seconds, so the test cannot be sped up below 1.1s without it. Production default still uses `WakeGapThresholdS` (120s); the millisecond field is a precondition override checked first.
- Test now runs with `WakeGapThresholdMs = 25` + `time.Sleep(50ms)` and `eventuallyWithin(t, 4*time.Second, ...)` for the pool-state assertion.

Lines 1715/1759 (loop pacing sleeps) left untouched per the audit prompt — they are not the actual correctness boundary.

## Test plan

- [x] `go test ./internal/storage/sqlite/... -count=5` (gateway) PASS
- [x] `go test ./internal/ws/... -count=5 -timeout=10m` (coordinator) PASS — package wall-clock drops from ~48s/iter to ~9s/iter
- [x] `go test -bench=BenchmarkAuthLookup -run=^$ -benchtime=1x ./internal/storage/sqlite/` reports `p95_ms` metric
- [ ] CI workflow from PR #10 stays green on this branch

## Note for human

No follow-up actions required. The new `wake_gap_threshold_ms` YAML key is opt-in; existing deployments are unaffected.

Generated with Claude Code.

### DIFF
--- a/phase4-coordinator/internal/audit/store.go
+++ b/phase4-coordinator/internal/audit/store.go
@@ -33,6 +33,11 @@ func OpenStore(dbPath string) (*Store, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Match phase5-gateway/internal/storage/sqlite/store.go:38-39: cap the
+	// pool at one connection so writers using BEGIN IMMEDIATE serialise
+	// inside the Go layer instead of contending in sqlite's 5s busy_timeout.
+	db.SetMaxOpenConns(1)
+	db.SetMaxIdleConns(1)
 	s := &Store{db: db}
 	ctx := context.Background()
 	if _, err := s.db.ExecContext(ctx, `PRAGMA journal_mode=WAL`); err != nil {
--- a/phase4-coordinator/internal/audit/store.go
+++ b/phase4-coordinator/internal/audit/store.go
@@ -33,11 +33,6 @@ func OpenStore(dbPath string) (*Store, error) {
 	if err != nil {
 		return nil, err
 	}
-	// Match phase5-gateway/internal/storage/sqlite/store.go:38-39: cap the
-	// pool at one connection so writers using BEGIN IMMEDIATE serialise
-	// inside the Go layer instead of contending in sqlite's 5s busy_timeout.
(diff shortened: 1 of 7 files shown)


Augustas11:
Follow-up commit 68ca022 — dropped 3 stray SQLite store changes (`internal/audit/store.go`, `internal/auth/tokens.go`, `internal/requestlog/store.go`) that came in via a stash collision during the M0 shipment. They're legitimate SetMaxOpenConns work but belong in their own M2-3 PR. This PR is now scope-clean on M0-3a + M0-3b.

PR #14, opened 2026-06-10
fix(coordinator): SetMaxOpenConns(1) on the three sqlite stores (QW-5 / ARCH-3)

Augustas11:
## Summary

Closes **QW-5 / M2-3** (ARCH-3) from [audits/2026-06-10/REPO_AUDIT.md](audits/2026-06-10/REPO_AUDIT.md).

The coordinator opens three `*sql.DB` handles against one sqlite file ([`cmd/coordinator/main.go:47,53,59`](phase4-coordinator/cmd/coordinator/main.go)) and the money-path billing writer takes `BEGIN IMMEDIATE` on top of that. Without `SetMaxOpenConns(1)`, Go's pool can concurrently hand out two write connections; the second `BEGIN IMMEDIATE` then waits on the sqlite 5s `busy_timeout`, surfacing either as p99 latency or as `request_log_failed` 500 after the inference already completed ([`buyer/server.go:1286-1290`](phase4-coordinator/internal/buyer/server.go)).

The gateway already caps at 1 deliberately at [`phase5-gateway/internal/storage/sqlite/store.go:38-39`](phase5-gateway/internal/storage/sqlite/store.go) — the team knows the pattern; the coordinator just omits it.

Edit lands inside each store's `OpenStore` (auth, requestlog, audit) rather than externally in `main.go`, so a future store added to the composition root cannot regress. Each call site carries a one-liner pointing back at the gateway as the source of the convention.

Money-path service, but money **math** is untouched — only connection pool sizing.

## Test plan

- [x] `cd phase4-coordinator && go vet ./...` — clean
- [x] `cd phase4-coordinator && go test ./... -count=1` — all packages green (billing, requestlog, audit, ws ~14s)
- [ ] CI green on this PR (`coordinator` job from #10)

## Note on a sibling branch

A separate branch on origin (`m0/deflake-perf-tests`) was created by a session hook that snapshotted unrelated `M0-3` deflake work-in-progress (`config.go`, `ws/server.go`, `server_test.go`, gateway `store_test.go`) when I started this PR. That work is intentionally **out of scope** here per the task plan and is not part of this PR. Operator: feel free to delete that origin branch if it wasn't intended.

### DIFF
--- a/phase4-coordinator/cmd/coordinator/main.go
+++ b/phase4-coordinator/cmd/coordinator/main.go
@@ -56,7 +56,11 @@ func main() {
 		os.Exit(1)
 	}
 	defer reqLogStore.Close()
-	auditStore, err := audit.OpenStore(cfg.Storage.DBPath)
+	// Share the requestlog *sql.DB so audit writes contend on the same single
+	// capped connection pool as request_log + admission + billing — opening a
+	// second *sql.DB against the same sqlite file would race writes on the 5s
+	// busy_timeout instead of serialising them in Go.
+	auditStore, err := audit.NewStoreFromDB(reqLogStore.DB())
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "audit log storage: %v\n", err)
 		os.Exit(1)
--- a/phase4-coordinator/internal/audit/store.go
+++ b/phase4-coordinator/internal/audit/store.go
@@ -33,6 +33,11 @@ func OpenStore(dbPath string) (*Store, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Match phase5-gateway/internal/storage/sqlite/store.go:38-39: cap the
+	// pool at one connection so writers using BEGIN IMMEDIATE serialise
+	// inside the Go layer instead of contending in sqlite's 5s busy_timeout.
+	db.SetMaxOpenConns(1)
+	db.SetMaxIdleConns(1)
 	s := &Store{db: db}
 	ctx := context.Background()
 	if _, err := s.db.ExecContext(ctx, `PRAGMA journal_mode=WAL`); err != nil {
--- a/phase4-coordinator/internal/audit/store.go
+++ b/phase4-coordinator/internal/audit/store.go
@@ -17,9 +17,16 @@ import (
 var errStoreClosed = errors.New("audit store is closed")
 
 type Store struct {
-	db *sql.DB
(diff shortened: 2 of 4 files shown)


Augustas11:
Closing without merge. Post-merge investigation found that the audit's S-effort estimate for ARC-3 was wrong: `SetMaxOpenConns(1)` on the shared coordinator `*sql.DB` deterministically deadlocks under any nested `*sql.DB` query pattern, and the coordinator's billing code has at least three such patterns the gateway code doesn't have.

**What was found** (during CI investigation after the post-merge review):

The `phase4-coordinator` CI job hangs at `internal/billing` and is killed at the 10-minute `timeout-minutes` ceiling. Two reruns reproduced the hang in the same place. Root cause is the classic "outer rows holds the only conn, inner query waits forever" deadlock at `MaxOpenConns=1`:

1. `internal/billing/store.go:182` — `rebuildLegacyConfigSnapshots`: outer `PRAGMA index_list` cursor + inner `PRAGMA index_info` per row.
2. `internal/billing/endpoints.go:141` — `providers` handler: outer rows + `h.sum(...)` inside the map literal per row.
3. `internal/billing/endpoints.go:267,273` — `buyerEquivalentCredits`: outer rows + `h.byteEstimatedLedgerGross(...)` and `h.store.snapshotAt(...)` per row.

These don't break the existing tests at uncapped conns because Go's `database/sql` silently opens a second connection when the pool isn't full. They also don't break locally on M-series Macs at cap=1 because the timing races through fast enough. They deterministically hang on a slow ubuntu CI runner at cap=1.

**Why this isn't a one-PR fix.** Unwinding all three sites requires a two-pass refactor (collect into slices, close outer cursor, then loop sub-queries) of money-path code. Per `AGENTS.md` that's a separate review-bound PR, not the "3 lines, matches gateway's proven pattern" the audit described.

**What landed instead.** Nothing on `main` from this PR. Both the `SetMaxOpenConns(1)` cap and the `audit.NewStoreFromDB` shared-handle change are discarded.

**Follow-up issue tracker** (to be opened): "ARC-3 — unwind nested *sql.DB queries in `phase4-coordinator/internal/billing/{store,endpoints}.go` before reintroducing `SetMaxOpenConns(1)` on the shared coordinator handle". S-effort estimate from `audits/2026-06-10/REPO_AUDIT.md` should be re-rated M.

PR #15, opened 2026-06-10
m0: monitor — preserve state on alerting-transition SMTP failure (M0-4a)

Augustas11:
## Summary

Milestone 0 M0-4a from `audits/2026-06-10/REPO_AUDIT.md`. Before this change, `macprovider-monitor.py` called `save_state()` unconditionally after attempting `send_email()`, and an SMTP failure produced a swallowed `[WARN]` line. That combination loses a CRITICAL or WARN transition forever if the single SMTP attempt at the moment of the flip happens to fail — the saved state reflects the new condition, so the next 3-minute cron poll sees no transition and stays silent.

## Change

- `send_email()` now returns `True` on delivery, `False` on failure, `None` when no non-journald recipient is configured (journal-only mode).
- `main()` persists `new_state` only if EITHER:
  - the new state is non-alerting (no CRITICAL/WARN entries — recovery INFO lines or no alerts), OR
  - `send_email` returned `True`.
- An alerting transition whose delivery failed keeps the OLD state so the next 3-minute cron poll re-detects the transition and re-fires.
- Journal-only configurations behave the same way: persistent alerting transitions keep reaching journald rather than silently sealing into the saved state.
- SMTP failures now log to stderr in addition to stdout so journald surfaces a distinct stream alerting tools can grep (`SMTP failure, will retry: <err>`).

## Verification (no test harness in tree for this script)

Smoke ran with `smtplib.SMTP` monkeypatched to raise `ConnectionRefusedError`, plus a saved state mismatching what we would compute, and a CRITICAL alert:

```
loaded before: {pool: {p1: ready}, ready: 1, coord_up: True, gw_status: ok}
[WARN] email send failed: simulated
SMTP failure, will retry: simulated
delivery: False
loaded after: {pool: {p1: ready}, ready: 1, coord_up: True, gw_status: ok}
PASS: old state preserved on SMTP failure
```

Same harness with a recovery-only INFO alerts list:
```
PASS: recovery saved new state despite SMTP failure
```

## Follow-up actions for human (M0-4b)

- [ ] Confirm `/etc/macprovider/monitor.env` on Pearl has `GMAIL_USER`, `GMAIL_APP_PASSWORD`, and `ALERT_EMAIL` populated. Without them the script runs in journal-only mode — which is now consistent with the new behavior (every 3-min poll re-emits the alert to journald until the condition clears) but defeats the purpose of having an alerting path.

Generated with Claude Code.

### DIFF
--- a/phase4-coordinator/dist/monitor/macprovider-monitor.py
+++ b/phase4-coordinator/dist/monitor/macprovider-monitor.py
@@ -83,6 +83,8 @@ def save_state(state):
     tmp = STATE_FILE + ".tmp"
     with open(tmp, "w") as f:
         json.dump(state, f)
+        f.flush()
+        os.fsync(f.fileno())
     os.replace(tmp, STATE_FILE)
 
 
@@ -177,7 +179,14 @@ def main():
         "gw_status": gw_status,
     }
     alerting = any(sev in ("CRITICAL", "WARN") for sev, _ in alerts)
-    if not alerting or delivery is True:
+    # `delivery is not False` keeps the strong "SMTP failed -> don't seal"
+    # guarantee while letting journal-only mode (delivery is None) advance
+    # state normally. Without an external delivery path journald IS the
+    # configured sink, and blocking save_state forever would refire every
+    # prior transition on every subsequent poll (not just the unfired
+    # alerting transition that originally failed). The original `delivery
+    # is True` gate had that flaw — flagged by the M0-4a code review.
+    if not alerting or delivery is not False:
         save_state(new_state)
     else:
         print("[INFO] state not advanced; will re-evaluate next cycle", flush=True)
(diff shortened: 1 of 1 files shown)


Augustas11:
Follow-up commit 47205c6 closes two M0-4a review findings:

1. **HIGH — journal-only mode freezing `prev_pool`.** Gate changed from `delivery is True` → `delivery is not False`. Now journal-only (`None`) → state advances normally; SMTP failure (`False`) → state preserved as before. Smoke test confirms 4/4 cases (SMTP failure preserves, journal-only advances, SMTP success advances, recovery-only INFO advances).
2. **MEDIUM (pre-existing) — `save_state` lacks fsync.** Added `f.flush(); os.fsync(f.fileno())` before `os.replace` so power loss in the write→rename window can't leave a zero-byte state file.

PR #18, opened 2026-06-10
m0: version-stamped builds and /healthz provenance (M0-5 Phase 1)

Augustas11:
## Summary

Milestone 0 M0-5 Phase 1 from `audits/2026-06-10/REPO_AUDIT.md`. The coordinator was hardcoding `"0.1.0"` in `/healthz` and the gateway had no version surface at all, so an operator could not tell which commit was actually running. The deploy gate downstream of this PR will assert `deployed:` against `expected:` git-describe — that change ships in a separate Phase 2 PR so the deploy step itself can stay operator-driven.

This PR is safe to land on its own — no production behavior changes; `/healthz` just learns a new key.

### Build scripts

- `phase4-coordinator/scripts/build-linux.sh`
- `phase5-gateway/scripts/build-linux.sh`

Both:
- Refuse to build with uncommitted changes unless `FORCE_DIRTY=1` is set.
- Stamp the binary via `-ldflags "-X main.version=$(git describe --always --dirty --tags)"`.
- Output to `dist/<module-name>-linux-amd64`.

Smoke:
```
$ ./phase4-coordinator/scripts/build-linux.sh
refusing to build with uncommitted changes (set FORCE_DIRTY=1 to override)
$ FORCE_DIRTY=1 ./phase4-coordinator/scripts/build-linux.sh
built dist/phase4-coordinator-linux-amd64 @ v1.3.0-4-g5c1a03a-dirty
$ file dist/phase4-coordinator-linux-amd64
... ELF 64-bit LSB executable, x86-64 ...
```

### Source plumbing

- `cmd/coordinator/main.go` and `cmd/gateway/main.go` each gain a package-level `var version = "dev"` and a `--version` flag.
- `buyer.WithVersion` (coordinator) and `router.WithVersion` (gateway) options inject the value into the respective `Server` struct.
- Coordinator `/healthz` now reports `s.version` instead of the hardcoded `"0.1.0"` string at `phase4-coordinator/internal/buyer/server.go:576`.
- Gateway `/healthz` now includes a `"version"` key in both the OK and the Service-Unavailable response shapes (`phase5-gateway/internal/router/server.go:492`).

### Tests

- `TestHealthzMountedOnBuyerHandler` extended to assert `"version":"dev"` on the default build.
- `TestHealthzReportsInjectedVersion` (new, both modules) drives `WithVersion("v1.3.0-7-gabcdef0")` and asserts the bytes round-trip into the response.
- Existing gateway `TestHealthzReturnsOK` now asserts that the `version` key is present and defaults to `"dev"`.

## Follow-up actions for human

None for this PR. The deploy script that actually consumes the new field ships in a separate Phase 2 PR, and the first deploy that uses any of this is operator-driven during a maintenance window.

Generated with Claude Code.

### DIFF
--- a/phase4-coordinator/cmd/coordinator/main.go
+++ b/phase4-coordinator/cmd/coordinator/main.go
@@ -26,9 +26,19 @@ import (
 	"github.com/rs/zerolog"
 )
 
+// version is overridden at build time via
+//   go build -ldflags "-X main.version=$(git describe --always --dirty --tags)"
+// (see scripts/build-linux.sh). Defaults to "dev" for local `go run`.
+var version = "dev"
+
 func main() {
 	configPath := flag.String("config", "coordinator.yaml", "path to coordinator YAML config")
+	showVersion := flag.Bool("version", false, "print build version and exit")
 	flag.Parse()
+	if *showVersion {
+		fmt.Println(version)
+		return
+	}
 
 	cfg, err := config.Load(*configPath)
 	if err != nil {
@@ -117,6 +127,7 @@ func main() {
 		registry,
 		logger,
 		startedAt,
+		buyer.WithVersion(version),
 		buyer.WithPreflightConfig(cfg.Routing.PreflightThresholdTokens, time.Duration(cfg.Routing.PreflightTimeoutS)*time.Second),
 		buyer.WithRecoveryConfig(time.Duration(cfg.Pool.DegradedBackoffS)*time.Second, cfg.Pool.DegradedMaxRetries, cfg.Pool.DegradedProbeAfter502),
 		buyer.WithBreakerConfig(cfg.Pool.BreakerFailureThreshold, time.Duration(cfg.Pool.BreakerWindowS)*time.Second),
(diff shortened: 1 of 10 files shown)


Augustas11:
Follow-up commit 7a572f0 closes two M0-5 Phase 1 review findings:

1. **MEDIUM — provider-port `/healthz` still hardcoded `"0.1.0"`.** Added `providerws.WithVersion(string) Option` mirroring `buyer.WithVersion`; wired through `cmd/coordinator/main.go`'s `wsOpts`. Both `/healthz` surfaces now report the build version. Two new tests (`TestProviderHealthzReportsInjectedVersion`, `TestProviderHealthzDefaultVersion`) lock the contract.
2. **MEDIUM — build scripts ignored untracked files.** Both `phase4-coordinator/scripts/build-linux.sh` and `phase5-gateway/scripts/build-linux.sh` now gate on `git status --porcelain -- .` (catches both modifications + untracked). Smoke: a `touch UNTRACKED_TEST_FILE` correctly trips the refusal and prints `?? UNTRACKED_TEST_FILE`.

**Not addressed in this commit:** the security review's "pool counts + git-describe leak on unauth `/healthz`" MEDIUM finding. The original M0-5 audit explicitly required the version field on the public `/healthz` so the deploy gate can read it; pool counts were already public pre-PR and have two consumers (`monitor.py` reads `pool_ready`; `deploy-pearl-vps.sh` step 6c/9 reads `pool_size`). Hiding them would break both consumers and contradict the audit's explicit ask. Flagged to track as a separate hardening ticket.

PR #19, opened 2026-06-10
m0: deploy snapshot + post-restart provenance assertion (M0-5 Phase 2)

Augustas11:
## Summary

Milestone 0 M0-5 Phase 2 from `audits/2026-06-10/REPO_AUDIT.md`. Builds on Phase 1 ([#18](https://github.com/Augustas11/macprovider/pull/18)) which made `/healthz` report the `git describe` version stamped into the binary. This PR consumes that surface from the deploy script and adds the rollback playbook.

This PR does NOT deploy anything. The first deploy that exercises the new step 4/9 snapshot and step 8/9 assertion is operator-driven in a maintenance window.

## `phase4-coordinator/dist/deploy-pearl-vps.sh`

- **Step 4/9** — before `scp` + `install`, snapshot the live binary at `/opt/macprovider/coordinator.prev` (`cp -p` to preserve permissions). First-time deploys skip the snapshot gracefully. This is the file the fast-path rollback in `ROLLBACK_PROCEDURE.md` swaps back.
- **Step 8/9** — after the public `/healthz` smoke, extract the deployed `version` field from the response and compare against `git describe --always --dirty --tags` from the deploying laptop. On mismatch print a `WARN` line referencing the rollback doc; non-fatal so the operator can keep an intentional divergence (e.g. an unstaged debug build) and is informed of it.

## `audits/2026-06-10/ROLLBACK_PROCEDURE.md`

One-page playbook:
- Prerequisites (SSH access, `.prev` snapshot exists).
- Fast path (binary swap, `mv`/`cp -p`, restart, verify), target wall-clock under 30s.
- Verification with a `/healthz` curl that proves the right version is live.
- When the fast path is wrong (config drift in same deploy).
- "Auto-rollback is not wired up" — explaining why the provenance assertion is non-fatal, so a future contributor doesn't add an auto-rollback shim without thinking about hidden-drift failure modes.
- Rebuilding the `.prev` snapshot from scratch on hosts where the deploy script never ran.

## Follow-up actions for human

- [ ] Schedule a maintenance window for the first deploy that uses this script. Per existing deploy-script guard (step 6c/9), the restart still refuses with connected providers unless `FORCE_RESTART=1`, so the M-tier providers should be drained or warned first.
- [ ] On Pearl during the first deploy, confirm that `/opt/macprovider/coordinator.prev` lands at step 4/9 and that the step 8/9 provenance line shows `deployed=<X> | expected=<X>` matching.

Generated with Claude Code.

### DIFF
--- a/audits/2026-06-10/ROLLBACK_PROCEDURE.md
+++ b/audits/2026-06-10/ROLLBACK_PROCEDURE.md
@@ -27,7 +27,7 @@ ssh -i ~/.ssh/pearl_operator_ed25519 root@159.223.165.194 '
     exit 1
   fi
   mv /opt/macprovider/coordinator /opt/macprovider/coordinator.bad
-  cp -p /opt/macprovider/coordinator.prev /opt/macprovider/coordinator
+  install -o macprovider -g macprovider -m 0755 /opt/macprovider/coordinator.prev /opt/macprovider/coordinator
   systemctl restart macprovider-coordinator
   sleep 2
   systemctl is-active macprovider-coordinator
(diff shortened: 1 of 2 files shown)


Augustas11:
Follow-up commit 4c2bbfe closes the two M0-5 Phase 2 security review findings:

1. **MEDIUM — `cp -p` ownership-preservation hazard.** Replaced `cp -p` with `install -o macprovider -g macprovider -m 0755` in step 4/9 of the deploy script AND in both the fast-path and rebuild-from-scratch snippets in `ROLLBACK_PROCEDURE.md`. Ownership is now asserted explicitly per invocation rather than inherited from the source binary's attrs, protecting against the rebuild-from-scratch path silently propagating root ownership.
2. **LOW — unbounded `/healthz` body parsing.** Added `--max-filesize 65536` to both curl invocations (step 6c/9 pre-restart pool-size check + step 8/9 post-restart provenance assertion). `--max-time` only bounds wall-clock; the byte cap defends the operator Mac against a misbehaving upstream.

The "Why `cp -p` not `mv`" section in `ROLLBACK_PROCEDURE.md` was retitled to "Why `install` not `mv`" and explains the ownership-explicit motivation.

Issue #125, opened 2026-06-23
phase4-coordinator: rate-limit keying on r.RemoteAddr defeats per-buyer limit behind nginx


Augustas11:
**Severity:** MEDIUM (pre-existing; surfaced again by Claude security review on PR #124)

## Problem

`phase4-coordinator/internal/buyer/server.go:830` (`poolCheckClientKey`) and the new `/v1/receipt-keys` handler both derive the rate-limit key from raw `r.RemoteAddr`:

```go
host, _, err := net.SplitHostPort(r.RemoteAddr)
```

Behind the production nginx reverse proxy on `coordinator.streamvc.live`, every request arrives with `r.RemoteAddr = 127.0.0.1:<ephemeral>` — the proxy peer, not the real client IP. **All traffic shares one global 10 req/s bucket**, defeating per-buyer rate limiting and meaning a noisy verifier loop or a benign CI fleet can starve every other buyer.

## Repro

Set up nginx → coordinator on localhost, fire 11 sequential `GET /v1/receipt-keys/m1-anon` from two different external IPs. Both hit the same bucket; the 11th request 429s regardless of source.

## Impact

- Buyers behind shared NAT (CI fleets from GitHub-hosted egress, corporate networks) eat 429s collectively after one noisy actor
- The intended per-IP limit is silently a global limit
- Same flaw also affects `/v1/pool/check` (pre-existing — wasn't introduced by SPEC-015 IMPL but surfaces in the same handler family)

## Suggested fix

Either:
- **(a)** Integrate with nginx-tier rate limiting and remove the in-process bucket on this route entirely
- **(b)** Add trusted-proxy `X-Forwarded-For` / `X-Real-IP` parse with an explicit allowlist of trusted proxy IPs/CIDRs
- **(c)** Move to per-CIDR aggregate buckets

The in-process bucket map's O(N) eviction loop (`server.go:807-828`) is also worth replacing with TTL-based map cleanup at this point.

## Cross-references

- Convergent finding from both Claude code-reviewer and Claude security-reviewer on PR #124
- Pre-existing in `handlePoolCheck` — fix applies to both handlers
- PR #124 merge commit: `99d0c1e`

Augustas11:
Status check 2026-06-26: **partial fix landed.**

[...]

PR #124, opened 2026-06-23
SPEC-015 v0.2.4 verifiable inference receipts (verifier IMPL + buyer-callable coordinator endpoint)

Augustas11:
## Summary

Lands **SPEC-015 v0.2.4 verifiable inference receipts** end-to-end: the locked SPEC + a complete buyer-side Go verifier (`macprovider-verify` v1.0.0) + the public `/v1/receipt-keys/<provider_id>` coordinator endpoint. Closes the SPEC-006 honesty gate — the README claim of signed receipts is now cryptographically verifiable by anyone running the CLI.

**Bundle scope:** 90 files / +17,364 / −30. SPEC + IMPL bundled per the `feedback-bundle-spec-impl-one-pr` rule (downstream incremental version on top of locked v0.1).

## What ships

| Component | Lines | Notes |
|---|---|---|
| `specs/SPEC-015-receipts.md` | ~2400 | v0.2.4 LOCKED after 5 SPEC audit rounds (5 CRITICAL + 11 MAJOR + 6 MINOR closed) |
| `phase4-coordinator/internal/buyer/server.go` | +124 | `handleReceiptKeys` — public, unauthenticated, 10 req/s/IP token bucket, redacted 4-field response, `Cache-Control: public, max-age=300` |
| `phase7-verify/` (new module) | ~5500 | Pure Go stdlib + `golang.org/x/text` (NFC only). 9 internal packages: jcs, receipt, canon, resolver, cache, verify, cli, schemavalidator, version |
| `phase7-verify/schemas/output.schema.json` | 291 | Draft-07 with `oneOf` per result variant + per-variant `if/then/else` for trust_source/coordinator_host coupling + `additionalProperties: false` everywhere |
| `phase7-verify/testdata/` | 11 fixtures | Deterministic golden bundles + reproducible generator |
| `.github/workflows/release-verify.yml` | 158 | 3-arch release (darwin-arm64/amd64, linux-amd64), `-trimpath -ldflags="-s -w"`, sha256 sidecars, smoke step gates release |
| `phase7-verify/README.md` | rewritten | Buyer-facing: install, quickstart, CLI reference, exit codes, schema link, trust-boundary |
| Root `README.md` | updated | Removed "planned, not yet implemented" caveat on receipts |
| `beta/DECISION_CRITERIA.md` Entry 83 | +1 row | Landing record + audit-round table |

## Audit story

**Two independent audit regimes, 0 false-valid paths found.**

| Regime | Cycles | Lenses | Verdict |
[...]

PR #129, opened 2026-06-23
fix(deploy): proxy /v1/receipt-keys/ on coordinator nginx (caught in PR #124 prod deploy)

Augustas11:
## Summary

Adds the nginx route block for `/v1/receipt-keys/<provider_id>` that was missing from PR #124's Step 0 work. The handler exists in coordinator code (lives on buyer port 8443), but the nginx site config had a catch-all `location /v1/ { return 404; }` block that intentionally blocked all `/v1/*` paths on the coordinator hostname.

Surfaced live during the verify-v1.0.0 production deploy on Pearl: smoke check returned nginx's text/html 404 instead of the coordinator's `application/json` envelope.

## Fix

Explicit `location /v1/receipt-keys/` block proxying to `127.0.0.1:8443`, placed **before** the catch-all `location /v1/` 404 block so nginx's prefix-longest-match rule routes receipt-keys requests through while every other `/v1/*` path stays 404.

```nginx
location /v1/receipt-keys/ {
    proxy_pass http://127.0.0.1:8443;
    proxy_set_header Host $host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_read_timeout 10s;
}
```

## Test plan

- [x] Verified live at `https://coordinator.streamvc.live/v1/receipt-keys/unknown-id` — returns `application/json` 404 with the coordinator's error envelope
- [x] `nginx -t` passes on Pearl
- [x] Provenance check: deployed coordinator at `verify-v1.0.0`
- [ ] Will return real keys once a v1.6+ provider connects (pool currently empty)

## Why this wasn't caught earlier

The IMPL audit chain (22 codex rounds + Claude 3-lens) covered the handler implementation + tests, but the nginx site config was treated as deploy-tooling — not exercised by unit/integration tests. The Step 9 mock used httptest.NewTLSServer rather than testing through the real nginx config.

Follow-up: PR #124 surfaced an existing finding (issue #125) that the rate-limit keying is broken behind nginx. Both that and this nginx-routing gap suggest CI should grow a "nginx-config-vs-handler-routes" lint that asserts every public buyer-port handler has a corresponding nginx `location` block.

---

### DIFF
--- a/phase4-coordinator/dist/nginx-coordinator.streamvc.live.conf
+++ b/phase4-coordinator/dist/nginx-coordinator.streamvc.live.conf
@@ -138,9 +138,29 @@ server {
     }
 
     # ---------------------------------------------------------------------
+    # /v1/receipt-keys/<provider_id> — public buyer endpoint per SPEC-002
+    # v1.5 candidate annotation + SPEC-015 §10.7. UNAUTHENTICATED. Lives on
+    # buyer_port (8443), NOT operator port. Rate-limited inside the
+    # coordinator (10 req/s per remote IP, 4 KiB max response, redacted to
+    # exactly {provider_id, receipt_pubkey, receipt_pubkey_prev, fetched_at}).
+    # MUST be declared BEFORE the catch-all `location /v1/` 404 block so the
+    # location-prefix longest-match rule routes receipt-keys requests here.
+    # ---------------------------------------------------------------------
+    location /v1/receipt-keys/ {
+        proxy_pass http://127.0.0.1:8443;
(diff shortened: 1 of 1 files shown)


antfleet-ops (approved):
Follow-up to PR #124. Nginx route fix for the /v1/receipt-keys public endpoint caught during the verify-v1.0.0 production deploy smoke. Live evidence confirms the fix works (JSON 404 envelope instead of nginx text/html 404). All 10 CI checks pass. Approving per user instruction.

PR #134, opened 2026-06-24
release(verify): macprovider-verify v1.1.0 — SPEC-015 v0.3.3 model-hash binding

Augustas11:
## Summary

Bumps `macprovider-verify` to v1.1.0 for the SPEC-015 v0.3.3 model-hash binding shipped in [#132](https://github.com/Augustas11/macprovider/pull/132) and deployed to Pearl coordinator at 2026-06-24T09:25:52Z (see `beta/DECISION_CRITERIA.md` Entry 85, [#133](https://github.com/Augustas11/macprovider/pull/133)).

Per §M.1.2 forward-incompat, this verifier release MUST land on buyers BEFORE v0.3-emitting providers reach them — existing locked v1.0.x verifiers report v0.3 receipts as `invalid`.

## What's in this PR

- **`phase7-verify/internal/version/version.go`**: `BinaryVersion 1.0.0 → 1.1.0`, `MaxSPECVersion 0.2.4 → 0.3.3`.
- **`phase7-verify/internal/version/version_test.go`**: `TestStep10FinalConstants` → `TestReleaseConstants` (pin guards against version drift mid-release; semantics unchanged).
- **`phase7-verify/README.md`**: install snippets bumped (3 platforms), compat table gains the `1.1.x` row covering `0.2.0 through 0.3.3`, forward-incompat note + runbook pointer added.
- **`phase7-verify/RELEASE_NOTES_v1.1.0.md`** (new): full notes covering wire shape, 5 new CLI flags, §M.3.2 algorithm, tri-state `model_hash_verified`, §M.3.4 cache mechanics, AC-38 parity gate, 19-round audit trajectory.

Pure-stdlib discipline preserved; `go.mod` unchanged. Full verifier suite green.

## Post-merge

1. Tag `verify-v1.1.0` on the merge commit.
2. Push the tag — `.github/workflows/release-verify.yml` triggers and builds darwin-arm64 + darwin-amd64 + linux-amd64 binaries + sha256 sums and creates the GitHub Release.
3. Operator updates the GitHub Release description with the contents of `phase7-verify/RELEASE_NOTES_v1.1.0.md` (the workflow's auto-generated notes are a one-line placeholder).
4. Buyers can pull the new binary via `gh release download verify-v1.1.0 ...` per README.

## Test plan

- [ ] Reviewer runs `cd phase7-verify && go test ./... -race -count=1` and confirms PASS (incl. the renamed `TestReleaseConstants` pin).
[...]

PR #136, opened 2026-06-24
fix(release-verify): parameterize smoke step + cross-check tag vs version.go

Augustas11:
## Summary

The v1.1.0 release [tag verify-v1.1.0](https://github.com/Augustas11/macprovider/releases/tag/verify-v1.1.0) triggered `release-verify.yml` after [PR #134](https://github.com/Augustas11/macprovider/pull/134) merged. All three Build jobs failed because the **Smoke verifier** step had a HARDCODED v1.0.0 grep:

```
grep -qx "macprovider-verify 1.0.0 (verifies up to SPEC-015 v0.2.4)" version.txt
```

The new binary correctly produces `macprovider-verify 1.1.0 (verifies up to SPEC-015 v0.3.3)`, so the grep failed and `fail-fast: true` cancelled the other builds. The release exists with no assets.

## Fix

1. **Smoke step is parameterized.** Greps `macprovider-verify ${VERSION} (verifies up to SPEC-015 v${SPEC_VERSION})` where both come from `prepare-release` outputs.
2. **`prepare-release` reads `BinaryVersion` + `MaxSPECVersion`** directly from `phase7-verify/internal/version/version.go` and **cross-checks the tag against `BinaryVersion`** BEFORE any binary is built. A tag whose name disagrees with the constant the binary will report at runtime fails the prep step — catching the inverse trap (tag pushed without bumping version.go) cheaply.

Both changes are version-agnostic. v0.4+ releases work without editing this workflow.

## Post-merge

1. Delete the existing `verify-v1.1.0` tag locally + on origin.
2. Re-tag `verify-v1.1.0` on the **merge commit of this PR** (so the fixed workflow is in scope).
3. Push the tag → `release-verify.yml` re-runs against the fixed workflow.
4. Once builds succeed, overwrite the GitHub Release description with `phase7-verify/RELEASE_NOTES_v1.1.0.md`.

## Test plan

- [ ] Reviewer reads the workflow diff and confirms the smoke step references workflow outputs, not literals.
- [ ] Reviewer confirms the cross-check awk-greps the same constant names as `phase7-verify/internal/version/version.go` defines.
- [ ] After merge: delete + re-push `verify-v1.1.0` tag and confirm all three Build jobs pass.


antfleet-ops (approved):
Approving workflow fix — smoke step parameterized + tag-vs-version.go cross-check added. Trivial CI-only change.

PR #135, opened 2026-06-24
ops(portal): commit nginx config + snippet for portal.streamvc.live (Entry 86)

Augustas11:
## Summary

- Captures the live `https://portal.streamvc.live/` deploy (provider-portal v0.1, decision-log Entry 86) as reproducible artifacts under `frontdoor/provider-portal/dist/`
- Adds `dist/README.md` documenting the deploy procedure + the two non-obvious deploy bugs caught + fixed during the operator activation
- Adds Entry 86 to `beta/DECISION_CRITERIA.md` recording deploy state + 5 operator follow-ups still owed
- Does NOT touch the parent `frontdoor/provider-portal/README.md` — that doc is SPEC-014-scoped; deploy specifics live in `dist/`

## Test plan

- [x] Files in `frontdoor/provider-portal/dist/` byte-identical to what's running on Pearl
- [x] Live smoke test against `https://portal.streamvc.live/`: HTTP/2 200 + security headers + correctly-proxied backends
- [ ] Follow-up PRs (Entry 86 ops list): snippet backport to console, surface portal URL, add portal to BetterStack monitor

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #137, opened 2026-06-24
fix(console): restore 5 missing security headers via snippet pattern (silent prod regression)

Augustas11:
## Summary

**This is a security-header regression fix masquerading as a refactor.** While moving the console nginx config to the same snippet pattern as the new `portal.streamvc.live` site (decision-log Entry 86 follow-up), I discovered console.streamvc.live had been serving ZERO of its 5 declared security headers (HSTS, CSP, X-Content-Type-Options, Referrer-Policy, Permissions-Policy) in production. Pre-deploy `curl -sI` returned only `cache-control`.

Root cause: nginx `add_header` is all-or-nothing per level. The location-level `add_header Cache-Control` directives were silently shadowing all 5 server-level security `add_header` directives even with `always` set.

## Fix

Extract security headers into `frontdoor/console/dist/nginx-snippets/console-security-headers.conf` (deployed to `/etc/nginx/snippets/console-security-headers.conf`) and `include` it in every `location` block. Same pattern shipped for portal in PR #135.

## Test plan

- [x] Pre-deploy `curl -sI https://console.streamvc.live/` → only `cache-control` header (regression confirmed)
- [x] Post-deploy `curl -sI https://console.streamvc.live/` → all 5 security headers + cache-control (regression fixed, byte-identical CSP text vs old server-level declaration)
- [x] `nginx -t` passes on Pearl, reload clean
- [x] Page content unchanged (same files, same routing, same caching)

### DIFF
--- a/frontdoor/console/dist/nginx-console.streamvc.live.conf
+++ b/frontdoor/console/dist/nginx-console.streamvc.live.conf
@@ -12,6 +12,14 @@ server {
     }
 }
 
+# Console security headers live in a snippet because nginx `add_header` is
+# all-or-nothing per level: any location-level `add_header` (e.g. the
+# Cache-Control directives below) shadows ALL inherited add_headers from
+# the server level — silent security-header regression if you forget.
+# Snippet path: /etc/nginx/snippets/console-security-headers.conf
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
LGTM — operator-activation cleanup; no code paths touched.

PR #139, opened 2026-06-24
ops(deploy): freshness gate in deploy-console.sh — block silent code-shipped-not-routed

Augustas11:
## Summary

Adds a "step 0/8 freshness gate" to `deploy-console.sh` that prints the sha256 of local repo / Pearl-served / origin/main versions of index.html on every run, and refuses to silently overwrite drift.

This is the documented operator follow-up "D2e" from decision-log Entry 86 — addresses the "code-shipped-not-routed" pattern hit 3× in 24h (Entry 81 SECU-4 envs, Entry 86 provider-portal, the 23-day console drift surfaced during the Entry 88-predecessor work).

## Behavior

- **Default (Pearl synced):** exit 0, "nothing to deploy"
- **Local diverges from origin/main:** exit 1 with WARN + `FORCE_OVERWRITE=1` instruction
- **`FORCE_OVERWRITE=1`:** bypass, proceed with un-pushed bytes (intentional hot-fix)
- **`SKIP_FRESHNESS=1`:** skip step 0 entirely (first deploy / offline)

Prints `Pearl bytes match commit <sha> (N ago) — <subject>` when Pearl runs an identifiable older commit, so staleness is immediately legible.

## Test plan

- [x] Default (Pearl up-to-date): exit 0, clean no-op
- [x] Local divergence: exit 1, WARN propagated
- [x] `FORCE_OVERWRITE=1` bypass works
- [x] `SKIP_FRESHNESS=1` skips step 0

### DIFF
--- a/frontdoor/console/dist/deploy-console.sh
+++ b/frontdoor/console/dist/deploy-console.sh
@@ -8,6 +8,8 @@
 # re-copied). Mirrors phase4-coordinator/dist/deploy-pearl-vps.sh.
 #
 # What it does:
+#   0. freshness check — compare local repo / Pearl-deployed / origin/main
+#      sha256 for index.html; refuse to silently overwrite Pearl drift
 #   1. confirm SSH
 #   2. ensure DNS A record console.streamvc.live -> VPS (Cloudflare API,
 #      DNS-only) — created if missing, skipped if present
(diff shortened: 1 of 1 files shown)


antfleet-ops (approved):
LGTM — deploy-script hygiene only, no runtime changes.

Issue #140, opened 2026-06-24
Track SPEC-010 min_ram_gb catalog metadata


Augustas11:
## Context

The install.sh fit-guard patch keeps the existing RAM-tier fallback because the current signed catalog schema (SPEC-010 v1.5, scripts/sign-catalog.go, and phase4-coordinator/internal/tier2/catalog.go) has no per-model RAM metadata and rejects unknown fields.

Live observation on 2026-06-24:
- /catalog/macprovider-tier2-model-catalog-2026-05-31 is public and signed.
- The catalog entries include model_id/source/sha256 metadata, but no min_ram_gb.
- Public /poolz currently returns 401 without operator auth, so installers cannot discover catalog_id through that path unauthenticated.

## Follow-up

Design a SPEC-010 v1.6 candidate or equivalent schema update that adds per-model min_ram_gb metadata to signed catalog entries, plus a public catalog-discovery path suitable for install.sh.

Keep this separate from the install.sh v1 guard patch; no schema or coordinator parser changes should be bundled into that PR.

Augustas11:
Closing as **largely stale** after fresh investigation against current code.

## What the issue claimed (2026-06-24)

> The install.sh fit-guard patch keeps the existing RAM-tier fallback because the current signed catalog schema (SPEC-010 v1.5, scripts/sign-catalog.go, and phase4-coordinator/internal/tier2/catalog.go) has no per-model RAM metadata and rejects unknown fields.

> Public /poolz currently returns 401 without operator auth, so installers cannot discover catalog_id through that path unauthenticated.

## What's actually true on `main` today

### `min_ram_gb` already exists in the catalog schema

Both the signer and the coordinator parser accept `min_ram_gb` as an optional `*int` per-model field:

- [scripts/sign-catalog.go:30](https://github.com/Augustas11/macprovider/blob/main/scripts/sign-catalog.go#L30) — `MinRAMGB *int json:"min_ram_gb,omitempty"`
- [scripts/sign-catalog.go:275](https://github.com/Augustas11/macprovider/blob/main/scripts/sign-catalog.go#L275) — `if model.MinRAMGB != nil && *model.MinRAMGB < 1 { return ... must be positive }`
- [phase4-coordinator/internal/tier2/catalog.go:34](https://github.com/Augustas11/macprovider/blob/main/phase4-coordinator/internal/tier2/catalog.go#L34) — same field declaration
[...]

PR #141, opened 2026-06-24
Guard install.sh against oversized explicit models

Augustas11:
## Summary

- closes the install.sh bypass where explicit MACPROVIDER_MODEL skipped RAM-fit enforcement, including MACPROVIDER_NO_PROMPT=1
- keeps explicit env overrides reachability-compatible with prior behavior while enforcing local name-based RAM fit
- adds best-effort catalog RAM metadata probing with fallback to built-in tiers when /poolz or min_ram_gb is unavailable
- prints the SPEC-013 autotune handoff after a successful ready-state install
- records the decision in beta/DECISION_CRITERIA.md Entry 87

## Gap check

- Existing interactive custom picker already called hf_check_model and kept warning/confirmation behavior.
- The explicit MACPROVIDER_MODEL path only validated format, so oversized models could proceed silently in non-interactive mode.
- Current live signed catalog macprovider-tier2-model-catalog-2026-05-31 has no min_ram_gb; public /poolz returns 401 without operator auth, so built-in tiers remain the v1 fallback.

## Follow-up

- Filed #140 to track SPEC-010/catalog min_ram_gb metadata and public catalog discovery separately from this install.sh patch.

## Verification

- bash -n phase3-binary/dist/install.sh
- git diff --check
- focused sourced-function checks:
  - NO_PROMPT oversized explicit model exits 7
  - 8 GB default model fits
  - choose_model 8 returns mlx-community/Llama-3.2-3B-Instruct-4bit
  - live catalog fallback warns and does not block when /poolz returns 401
  - min_ram_gb parser returns expected value from fixture JSON
  - autotune handoff prints macprovider-cli autotune --provider-id <id>
- omc ask codex audit: no HIGH/CRITICAL findings

Not tested: full macOS install path and launchd self-test from this Linux-hosted session.


antfleet-ops (approved):
Reviewed install.sh fit-guard diff, validation evidence, and green CI. Approving for squash merge.

PR #145, opened 2026-06-24
fix(phase7-verify): pin verifier "now" in fixture tests via testclock build tag

Augustas11:
## Summary

- The `phase7-verify integration (fixtures)` CI job has been failing on `main` since ~26h after fixture generation. `valid_fresh` and `valid_prev_key_in_grace` both signed-receipts with `unix_ts = 2026-06-23T12:00:00Z`, and `verify.go` emits a `clock_skew` warning when `|system_time - unix_ts| > clockSkewThreshold` (24h). The tests asserted `warnings == [non_default_coordinator]`, so the extra `clock_skew` warning broke them. This recurs every 24h until structurally fixed.
- Fix: add a build-tag-gated clock override. The production binary is unchanged — `clock_default.go` (`//go:build !testclock`) returns `nil`. Only when built with `-tags=testclock` does `clock_testclock.go` honor `MACPROVIDER_VERIFY_NOW_UNIX` and feed it into `cfg.now` → `verify.VerifyOpts.Now`.
- `TestMain` adds `-tags=testclock` to `go build`, and `runVerifyArgs` sets the env var to `validFreshUnixTS + 60`. Verifier sees receipts as exactly 1 minute old. Fixtures stay stable forever.
- The `seedStale: true` cache write also rebased to the pinned clock: `hasStaleCacheEntry` compares cache `FetchedAt` against the pinned "now", and a real-clock offset that looks stale to wall time can look fresh to the verifier once the clocks disagree by ~26h. Cache-write epoch is now `pinnedNow - 8d`, which is stale to both pinned and real clocks.

## Why not the alternatives

- **Bump `validFreshUnixTS`** — only buys 24h before the same break.
- **Regenerate fixtures in `TestMain` against `time.Now()`** — would require extracting `testdata/generator` helpers into a reusable package; larger refactor, and `TestFixtureGeneratorIdempotentAndCommitted` would have to be rethought.
- **Production env-var clock override (no build tag)** — adds an env-var-controlled clock to a production verifier that decides cryptographic freshness. The build tag keeps the surface area zero in shipped binaries.

## Test plan

- [x] `go test -tags=integration ./...` in `phase7-verify` — all green (was: `valid_fresh` + `valid_prev_key_in_grace` red).
- [x] `go test ./...` (default build, no tag) — all green; production code path unaffected.
- [x] `go build -tags=testclock ./cmd/macprovider-verify` — clean build.
- [x] `go vet -tags=integration ./...` — clean.
- [ ] CI green on this PR.

## Notes

- Once CI is green here, [#144](https://github.com/Augustas11/macprovider/pull/144) (installer-latest filter, blocked by this same failure) can also go green and merge.

### DIFF
--- a/phase7-verify/integration_test.go
+++ b/phase7-verify/integration_test.go
@@ -74,7 +74,11 @@ func TestMain(m *testing.M) {
 	defer os.RemoveAll(tmp)
 
 	verifyBin = filepath.Join(tmp, "macprovider-verify")
-	cmd := exec.Command("go", "build", "-o", verifyBin, "./cmd/macprovider-verify")
+	// `-tags=testclock` enables the MACPROVIDER_VERIFY_NOW_UNIX hook in
+	// internal/cli (clock_testclock.go), letting tests pin "now" against
+	// the frozen receipt fixtures so clock_skew doesn't fire as fixtures
+	// age past 24h. Production builds (no tag) ignore the env var.
+	cmd := exec.Command("go", "build", "-tags=testclock", "-o", verifyBin, "./cmd/macprovider-verify")
 	cmd.Env = append(os.Environ(), "CGO_ENABLED=0")
 	if out, err := cmd.CombinedOutput(); err != nil {
 		fmt.Fprintf(os.Stderr, "go build ./cmd/macprovider-verify: %v\n%s", err, out)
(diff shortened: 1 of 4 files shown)


antfleet-ops (approved):
Approve: build-tag-gated test clock; production binary unchanged; full integration suite green locally; clean go vet/build.

PR #148, opened 2026-06-24
release: skip raw CLI stapling after notarization

antfleet-ops:
## Summary
- remove direct `xcrun stapler` calls for the standalone CLI executable
- keep Developer ID signing and `notarytool submit --wait` as the release gate
- update the operator runbook with the raw executable stapling limitation

## Validation
- `git diff --check`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release.yml"); puts "release.yml yaml ok"'`

## Reason
The v1.6.0 release run now signs successfully and Apple notarization returns Accepted, but fails afterward because `xcrun stapler` cannot attach a ticket to a raw standalone Mach-O executable inside the current tarball format.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -69,13 +69,16 @@ jobs:
       # macprovider-launchd-amfi-blocker-macos-26 for the full discovery
       # writeup (2026-06-12 session).
       #
-      # The fix: codesign the binary with a Developer ID Application cert,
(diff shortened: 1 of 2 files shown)


Augustas11 (approved):
Approved release workflow fix after CI passed. Evidence: signing/notarization succeeded in v1.6.0 run; failure was only raw executable stapling, and local stapler help confirms raw Mach-O is not a supported stapler target.

PR #156, opened 2026-06-25
Restore public provider bootstrap under strict token auth

Augustas11:
## Summary
- Adds explicit `auth.allow_tokenless_provisional_bootstrap` so production can keep `auth.require_provider_tokens=true` while allowing first-time provisional providers to mint and persist their provider token.
- Keeps pinned providers, used-token provider IDs, and missing issuer paths fail-closed; adds v1/v2 bootstrap and rejection coverage plus FR-C10 pair-OT fallback coverage.
- Hardens deploy config validation so the bootstrap and C2 timeout choices must be explicit in the correct YAML sections, and updates README/OPS/spec/decision-log guidance.

## Verification
- `go test -count=1 ./...` from `phase4-coordinator`
- `make test-dist`
- `git diff --check`
- Final code, security, and architecture auditors: PASS, 0 critical/high/medium findings

## Not Tested
- Live Pearl deploy
- Clean physical Mac reinstall on macOS 26.3.1+


antfleet-ops (approved):
Reviewed the onboarding-token bootstrap follow-up. Approval scope: unblock public clean installs while preserving strict token auth for pinned/used-token providers; local tests and final code/security/architecture audits are green.

PR #159, opened 2026-06-25
Fix gateway header timeout and late encrypted relay cleanup

antfleet-ops:
## Summary
- raise the gateway coordinator header timeout default/example from 10s to 60s so non-streaming large-model inference has time to produce first headers
- document why `ResponseHeaderTimeout` must cover non-streaming first-response latency
- keep valid late encrypted response frames for retired relay requests from closing the provider session as AEAD failures

## Why
PR #143 gateway e2e reached the coordinator and provider, but Pearl gateway returned `503 provider_unavailable` after about 10s. The failure path was the live gateway header timeout expiring before non-streaming Qwen3 returned response headers. The coordinator then retired the active relay, and a later valid encrypted provider terminal frame was misclassified as an unknown request / AEAD failure.

## Validation
- `git diff --check`
- `cd phase4-coordinator && go test ./internal/ws`
- `cd phase5-gateway && go test ./internal/config ./cmd/gateway ./internal/router`

## Operator follow-up still required
- update live Pearl `/opt/macprovider/gateway.yaml` to set `timeouts.coordinator_header_timeout_seconds: 60` or higher
- restart the gateway
- rerun the PR #143 gateway e2e through the public gateway after deploy

Draft because this is the follow-up implementation PR for the findings; the operator/deploy work above remains outside this branch.

PR #160, opened 2026-06-25
[codex] Recognize Qwen coding tool calls

Augustas11:
## Summary

- Normalize Qwen Python-style tool calls such as `find_definition(symbol="ToolCallParser")` when they appear inside `<tool_call>...</tool_call>` delimiters.
- Keep malformed, positional, undeclared, or unsupported expressions on the existing plain-text fallback path.
- Switch the tool-calling e2e and demo from the weather fixture to a coding-domain `find_definition` tool.

## Root Cause

The public Qwen3 gateway e2e reached the provider successfully after PR #159, but the model emitted a Python-style call inside `<tool_call>` tags as assistant content. The provider parser only accepted JSON bodies inside those delimiters, so the response returned `finish_reason=stop` instead of OpenAI `tool_calls[]`.

## Validation

- `cd phase3-binary && swift test --filter ToolCallParserTests`
- `cd phase3-binary && swift test`
- `python3 -m py_compile test/integration/tool_calling/openai_tool_call_e2e.py examples/tool_calling_demo.py`
- `git diff --check`

## Not Tested

- Live public gateway e2e after this provider-side change. That requires deploying an updated provider binary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ Latest release and signed binaries: [github.com/augustas11/macprovider/releases]
 
 :white_check_mark: **Signed inference receipts — Shipped in v1.0.0.** The SPEC-015 v0.2 verifier is available as the open-source [macprovider-verify](phase7-verify/README.md) CLI.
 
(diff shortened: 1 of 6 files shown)


antfleet-ops (approved):
Approved after CI green and code/security/architecture audit rounds reported zero critical/high/medium findings.

PR #192, opened 2026-06-27
specs: SPEC-002 v1.4.2 + SPEC-006 v0.X gateway-contract addenda (phase-A network-harness audit)

Augustas11:
## Summary

Two paired contract addenda derived from an internal e2e network-harness pass against the live `https://api.streamvc.live` stack (2026-06-27), audited in parallel by codex (architect lane via `omc ask codex`) and Claude adversarial verification subagent.

- **SPEC-002 v1.4.2** (coordinator-side): five clauses, all re-affirmations of existing locked contracts (§ 7.2, § 11, FR-B6, FR-B4) plus one correction. No new buyer HTTP surface; no routing-algorithm changes.
- **SPEC-006 v0.X.Y** (gateway-side companion): four clauses — §17.7 settlement-matrix re-affirmation, new normative rate-limit response header set, observability requirements, SPEC-005 token-source vocabulary re-affirmation.

## Why this PR is mostly re-affirmations, not new contract

The earlier draft framed 5 clauses as "PRODUCT DECISION" tags. The codex+Claude audit converged on the conclusion that **most of those decisions were already-locked spec contracts the empirical observation had violated**. The PR's role is to make those violations visible against the contract, file them for code-side fixes, and clarify a small number of genuine spec ambiguities.

## Companion issues filed (8 total)

| Issue | Severity | Spec |
|---|---|---|
| [#184](https://github.com/Augustas11/macprovider/issues/184) | high | SPEC-002 FR-B4 — 503 code drift |
| [#185](https://github.com/Augustas11/macprovider/issues/185) | high | SPEC-002 § 7.2 — 404/503 split |
| [#186](https://github.com/Augustas11/macprovider/issues/186) | high | SPEC-002 FR-B6 — mid-stream SSE envelope |
| [#187](https://github.com/Augustas11/macprovider/issues/187) | **P0** | SPEC-006 §17.7 — settlement matrix |
| [#188](https://github.com/Augustas11/macprovider/issues/188) | **P0** | SPEC-002 §11 — X-Request-ID propagation |
| [#189](https://github.com/Augustas11/macprovider/issues/189) | high | Provider WS reconnect loop wedge |
| [#190](https://github.com/Augustas11/macprovider/issues/190) | medium | Rate-limit response headers + N decision |
| [#191](https://github.com/Augustas11/macprovider/issues/191) | medium | Fleet-wide WS-health watchdog |

## Audit-pass summary

| Lane | Findings | CRITICAL | MAJOR | MINOR | NOTE |
[...]

PR #195, opened 2026-06-27
fix(coordinator,gateway): propagate buyer X-Request-ID end-to-end (closes #188)

### LINKED ISSUE #188: P0 bug(gateway+coordinator): X-Request-ID not propagated end-to-end (SPEC-002 §11 violation)
## Summary — P0 billing-reconciliation blocker for beta

SPEC-002 v1.4.1 § 11 (lines 2239-2243) is normative:

> Coordinator MUST honor any inbound `X-Request-ID` header on buyer-facing `/v1/*` requests and include it in the `request_log` row. If absent, coordinator MAY generate its own UUID v4.
>
> When forwarding work to a provider over the SPEC-001 § 6.6 `inference_request` message, coordinator MUST preserve the request ID it recorded for the buyer request.
>
> Gateway-originated traffic from SPEC-006 v0.3+ uses `X-Request-ID` as the join key between gateway `usage_events`, gateway `audit_events`, and coordinator `request_log`.

Phase-A observed: the gateway returns one UUID to the buyer, the coordinator writes a DIFFERENT UUID into `request_log` for the same logical request. They never overlap. Verified via cross-DB inspection.

## Impact

Out-of-process auditors (the internal e2e harness, future SRE audit scripts, dispute-resolution tooling) **cannot reliably reconcile** which `usage_events` row corresponds to which `request_log` row. Fall back to fuzzy matching by `(ts ± 60s, model, completion_tokens, account)` which collapses under concurrent buyer traffic.

In external beta: **a buyer disputing a charge requires manual log archaeology** to resolve. No honest mechanism for a structured billing-dispute SLA.

## Likely root cause

One of:
1. Gateway NOT setting `X-Request-ID` on the coordinator-facing call.
2. Coordinator NOT honoring inbound `X-Request-ID` and generating its own.
3. Gateway honoring `X-Request-ID` on the buyer-facing side but generating a different one for the coordinator-facing forward.

Investigation starts at the boundary between phase5-gateway/internal/router/chat_proxy.go (where gateway issues the coordinator call) and phase4-coordinator/internal/buyer/server.go (where coordinator accepts buyer traffic).

## Fix

1. Identify which option above is the actual cause.
2. Plumb the buyer-facing `X-Request-ID` verbatim through gateway → coordinator → provider; write identically into `usage_events.request_id` and `request_log.request_id`.
3. The schema already supports this (UUID v4 TEXT, indexed per § 11 line 1313).

## Acceptance

- A buyer request producing `X-Request-ID: abc-123` writes the same id to gateway `usage_events.request_id` AND coordinator `request_log.request_id`.
- Internal e2e harness `ledger_reconcile.json` matches successfully on exact request_id join across both DBs (no fuzzy matching needed).
- Phase-A scenarios 02 + 03 (the noisy +17/+32 token "drift") re-run with this fix should report zero drift — the apparent overcharges were almost certainly concurrent live traffic confounded by this id-space gap.

## Context

Surfaced during phase-A internal network-harness review 2026-06-27. Codex audit Finding 1 (most critical overall); SPEC-002 v1.4.2 addendum R-2.

Augustas11:
## Summary

Closes #188 (P0).

SPEC-002 v1.4.1 §11 (lines 2239-2243) mandates that the buyer-visible `X-Request-ID` flows through gateway → coordinator and serves as the join key between gateway `usage_events` and coordinator `request_log`. Earlier code violated this on two fronts:

- **Gateway** ([phase5-gateway/internal/router/chat_proxy.go:202](phase5-gateway/internal/router/chat_proxy.go#L202)) minted a fresh UUID for the coordinator-bound request, discarding the buyer's id.
- **Coordinator** ([phase4-coordinator/internal/buyer/server.go:1244-1247](phase4-coordinator/internal/buyer/server.go#L1244-L1247)) read the inbound `X-Request-ID` into `externalRequestID` but never persisted it — `request_log.request_id` was always a per-attempt coord-generated UUID.

Result: gateway `usage_events.request_id` and coord `request_log.request_id` lived in entirely separate id spaces with zero overlap. Out-of-process auditors (the internal e2e harness, any SRE audit, any billing-dispute resolver) could only reconcile by fuzzy `(ts ± 60s, model, completion_tokens)` matching, which collapses under concurrent live traffic.

## Approach: additive `external_request_id` column, not flip-the-semantics

The existing test suite deliberately codifies "per-attempt unique request_id" semantics in coordinator's request_log (test_RequestLogBuyer*** at server_test.go:1045 and :1114). Flipping that contract would have wide blast radius.

Instead this PR adds an **additive** column `request_log.external_request_id` (TEXT NULL, migrated) and stores the buyer-visible `X-Request-ID` there. `request_log.request_id` keeps its existing per-attempt-coord-generated semantics. The reconciliation join is now:

```
gateway.usage_events.request_id  ==  coord.request_log.external_request_id
```

## Changes

- **phase4-coordinator/internal/requestlog/store.go** — additive schema migration for `external_request_id TEXT NULL`; `Row` struct field; INSERT extended.
- **phase4-coordinator/internal/buyer/billing_recorder.go** — recorder carries `externalRequestID`, writes it into every row (success + retry attempts share the same value).
- **phase4-coordinator/internal/buyer/server.go** — passes `externalRequestID` into the recorder constructor.
- **phase5-gateway/internal/router/chat_proxy.go** — forwards `requestID(r)` (the gateway-middleware id, already honored from inbound `X-Request-ID` per §11) to coordinator verbatim instead of minting a fresh UUID. Three-line comment block explains the contract.
- **phase4-coordinator/internal/buyer/server_test.go** — added 2 new tests:
  - `TestRequestLogPreservesInboundXRequestIDAcrossAttempts` — positive: external_request_id populated, persists across retry attempts, request_id stays per-attempt coord-generated, retry rows share request_id (existing non-pinned invariant preserved).
  - `TestRequestLogExternalRequestIDNullWhenHeaderAbsent` — negative: NULL when no inbound header.
- **phase4-coordinator/internal/buyer/server_test.go** — extended query helpers (`requestLogTestRow`, `queryRequestLogRows`, `queryAllRequestLogRows`) to scan the new column.
- **phase5-gateway/internal/router/server_test.go** — flipped `TestProviderPinningHeadersStripped`'s X-Request-ID assertion (it previously codified the broken behavior). Now asserts the buyer's value is preserved through to the coordinator-bound request.

Refs SPEC-002 v1.4.2 R-2 from PR #192.

## Test plan

- [x] `go build ./...` clean on both modules.
- [x] `go test ./internal/buyer/ ./internal/requestlog/ -count=1` PASS on coordinator.
- [x] `go test ./internal/router/ -count=1` PASS on gateway.
- [x] New regression tests cover positive + negative paths.
- [ ] (After deploy) Re-run internal e2e network-harness scenarios 02 + 03 and confirm `ledger_reconcile.json` reports `unmatched_successes: []` and `gateway_minus_coordinator_tokens: 0` (the apparent +17 / +32 token "drift" from phase-A was almost certainly concurrent live traffic confounded by this id-space gap; the join now disambiguates).

## Out of scope (filed separately)

- Provider-forward `X-Request-ID` preservation per §11 line 2241 — touches a different code path with its own test invariants; will land as a follow-up against a new issue if/when it's a beta requirement.
- Schema migration safety on live Pearl — additive `ALTER TABLE ADD COLUMN` is the safe path SQLite supports without downtime; pre-existing rows have NULL in the new column (acceptable for reconciliation since they'll all fail the join cleanly).

---

**PR ordering note (architect R4 audit):** This PR depends on **PR #192** (SPEC-002 v1.4.2 R-2). #192 explicitly supersedes the locked v1.4.1 `request_id` schema-row description so this implementation does not contradict the spec. **#192 MUST merge before this PR**, or both merge in the same merge train. Reviewers should treat them as a paired set.

---

## Deployment runbook (operator action required)

PR #195 changes the SQLite migration shape on the request-log path. After deploy, BEFORE re-binding traffic, the operator MUST run:

```bash
./coordinator migrate-indexes --config /etc/macprovider/coordinator.yaml
```

The daemon writes the new `request_log.external_request_id` column on first start (additive ALTER TABLE — cheap, runs inline). The matching partial-NULL index is intentionally NOT built by the daemon: the request-log SQLite handle caps the writer pool at one connection, so CREATE INDEX inline would starve the 6s-timeout INSERT hot path. The subcommand exists for the operator to run this CREATE INDEX once per deploy.

**Until the index is built, exact reconciliation queries still work** — they just run a full table scan instead of a partial-index lookup. The new column is populated correctly from the moment the daemon ships. Internal harness Phase-C SHOULD introspect both the column AND the index, and report "exact reconciliation available but unindexed / rollout incomplete" when the column exists without the index.

PR-introduced check pattern:

```bash
sqlite3 /var/lib/macprovider/coordinator.db <<SQL
.schema request_log | grep -q external_request_id && echo "column: ok"
SELECT name FROM sqlite_master WHERE type='index' AND name='idx_request_log_external_request_id';
SQL
```

### DIFF
--- a/phase4-coordinator/cmd/coordinator/main.go
+++ b/phase4-coordinator/cmd/coordinator/main.go
@@ -129,6 +129,20 @@ func main() {
 		os.Exit(1)
 	}
 	defer reqLogStore.Close()
+	// SPEC-002 v1.4.2 R-2 / ISS-188: request_log.external_request_id
+	// is added by OpenStore as an additive column. The matching partial-
+	// NULL index is built asynchronously here: SQLite holds the writer
+	// lock through CREATE INDEX, so doing this inline would stall the
+	// hot path during deploy. A failure here only loses the
+	// reconciliation-side index — INSERTs and the new column still work
+	// — so we log and continue. Repeated calls are idempotent.
+	go func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+		defer cancel()
+		if err := reqLogStore.MigrateIndexes(ctx); err != nil {
+			fmt.Fprintf(os.Stderr, "requestlog MigrateIndexes (non-fatal): %v\n", err)
+		}
+	}()
 	canaryStore, err := setupCanarySanctionStore(context.Background(), cfg, reqLogStore.DB(), registry)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "canary sanction storage: %v\n", err)
--- a/phase4-coordinator/cmd/coordinator/main.go
+++ b/phase4-coordinator/cmd/coordinator/main.go
@@ -76,6 +76,8 @@ func main() {
 			os.Exit(runPartnerKeys(os.Args[2:]))
 		case "visibility":
 			os.Exit(runVisibility(os.Args[2:]))
+		case "migrate-indexes":
+			os.Exit(runMigrateIndexes(os.Args[2:]))
 		}
 		// Round-1 CODE H1 fix: a non-flag first positional that
 		// is NEITHER a known daemon flag NOR a known CLI verb is
@@ -91,6 +93,7 @@ func main() {
 			fmt.Fprintln(os.Stderr, "  coordinator --version           (print build version)")
 			fmt.Fprintln(os.Stderr, "  coordinator partner-keys <issue|revoke|list> [flags]")
 			fmt.Fprintln(os.Stderr, "  coordinator visibility revert --id <pid> --reason TEXT")
+			fmt.Fprintln(os.Stderr, "  coordinator migrate-indexes --config <path>  (one-shot operator migration)")
 			os.Exit(2)
 		}
 	}
@@ -131,18 +134,14 @@ func main() {
 	defer reqLogStore.Close()
 	// SPEC-002 v1.4.2 R-2 / ISS-188: request_log.external_request_id
 	// is added by OpenStore as an additive column. The matching partial-
-	// NULL index is built asynchronously here: SQLite holds the writer
-	// lock through CREATE INDEX, so doing this inline would stall the
-	// hot path during deploy. A failure here only loses the
-	// reconciliation-side index — INSERTs and the new column still work
-	// — so we log and continue. Repeated calls are idempotent.
-	go func() {
-		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
-		defer cancel()
-		if err := reqLogStore.MigrateIndexes(ctx); err != nil {
-			fmt.Fprintf(os.Stderr, "requestlog MigrateIndexes (non-fatal): %v\n", err)
-		}
-	}()
+	// NULL reconciliation index is NOT auto-built here — the request-log
+	// store caps the pool at one writer connection (see
+	// requestlog.OpenStore SetMaxOpenConns(1)), so running CREATE INDEX
+	// from the daemon would contend with the 6s-timeout INSERT hot
+	// path. The index ships via the `coordinator migrate-indexes`
+	// subcommand, intended to be invoked once per deploy by the
+	// operator runbook before binding traffic (or during a maintenance
+	// window).
 	canaryStore, err := setupCanarySanctionStore(context.Background(), cfg, reqLogStore.DB(), registry)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "canary sanction storage: %v\n", err)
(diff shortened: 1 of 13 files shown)


antfleet-ops (approved):
Approved — three-lane codex audit converged across 8 rounds. Code 0 C/0 M, security 0 C/0 M, architect 0 C/0 M-on-PR-introduced-surface (R8 spec-evolution asks filed as #197 for v1.4.3, not blockers). PR-introduced regression suite + runbook in PR body.

PR #205, opened 2026-06-28
feat(gateway): per-account concurrency rate-limit headers + N=3 decision (closes #190)

### LINKED ISSUE #190: feat(gateway): per-account rate-limit response headers (X-RateLimit-*, Retry-After) + numeric N decision
## Summary

The gateway emits `HTTP 429` on per-account rate-limit rejections but does NOT include the `X-RateLimit-*` headers that OpenAI-compatible client SDKs honor for self-pacing. Without these headers, beta buyers fire blindly until 429 and retry without backoff signal.

Phase-A scenario `02_capacity_contention.yaml` (10 concurrent buyers from one account) produced 7×429 + 2×503 + 1×200. The 429 response body was `{"error":{"code":"http_429"}}` with no rate-limit headers.

## Required response shape (per SPEC-006 v0.X addendum R-G2)

```
HTTP 429
X-RateLimit-Limit: <N>                 # per-account concurrency cap
X-RateLimit-Remaining: 0
X-RateLimit-Reset: <ISO 8601 ts or seconds-from-now>
Retry-After: <integer seconds, <= time until X-RateLimit-Reset>
Content-Type: application/json

{"error":{"message":"...","type":"rate_limit_exceeded","code":"rate_limit_exceeded","param":null}}
```

`X-RateLimit-Limit` SHOULD also appear on 200 and 4xx responses so the client can pre-emptively self-pace.

## Open product decision: numeric value of N

**Recommended N = 3 (concurrent in-flight requests per API key), configurable per subscription tier.**

Rationale:
- N=1 breaks the common UX pattern "user types follow-up while previous response is still streaming". OpenAI SDK default `n=1` is per-completion choices, NOT per-account concurrency — beta buyers expect at least 2-3 concurrent.
- N=3 matches the current network's effective capacity (3 providers × 1 slot each).
- Configurable per-tier lets paid tiers scale up without re-deploying the gateway.

Alternative views from the audit:
- **Codex view:** scale N dynamically with provider capacity (sum of `slots_total` across ready providers); avoid a fixed number that drifts from reality.
- **Conservative view:** start at N=2, raise during beta based on observed contention; safer to grant more than to revoke.

## Fix

1. Implement the response shape above in the rate-limit reject path of `phase5-gateway/`.
2. Pick N (recommend 3) and make it configurable per API key.
3. Ensure `Retry-After` is a deterministic, sane value (min seconds-until-an-in-flight-completes; cap at 60s).
4. Emit `X-RateLimit-Limit` on every 200/2xx response too so SDKs can self-pace.

## Acceptance

- A burst of N+1 concurrent requests from one API key produces N×200 + 1×429.
- The 429 response carries `X-RateLimit-Limit: <N>`, `Retry-After: <reasonable seconds>`, `error.code: "rate_limit_exceeded"`.
- Internal e2e harness scenario `02_capacity_contention.yaml` shows the rate-limit headers captured in per_request rows (requires harness header capture; tracked separately).

## Context

Surfaced during phase-A internal network-harness review 2026-06-27. Codex audit Finding 7 + Finding 11; SPEC-006 v0.X gateway-contract addendum R-G2.

Augustas11:
## Summary

Closes #190. Per-account concurrency rejections now carry the OpenAI-compatible rate-limit headers that client SDKs need to self-pace.

- **N decision**: `AccountConcurrency` default bumped 2 → 3 (matches phase-A network capacity of 3 providers × 1 slot). `DemoConcurrency` stays at 2 to avoid re-introducing the M1-8 / PERF-6 MLX-pool saturation regression.
- **Headers on 429 concurrency reject**: `X-RateLimit-Limit-Requests`, `X-RateLimit-Remaining-Requests: 0`, `X-RateLimit-Reset-Requests`, `Retry-After: 1`. The `-Requests` suffix disambiguates from the existing daily-token `X-RateLimit-*` triplet.
- **Headers on 200 admit**: `X-RateLimit-Limit-Requests` + `X-RateLimit-Remaining-Requests` so SDKs can throttle before they hit the cap.
- **Cache safety**: per-tenant headers reveal in-flight count, so `Cache-Control: no-store` + `Vary: Authorization, X-Demo-Token` (Add not Set, preserving any CORS-supplied `Vary: Origin`) are set at handler entry on every chat-completion response. Streaming success path was updated to `Cache-Control: no-store, no-cache, no-transform` — preserving no-store while keeping the SSE-required no-cache/no-transform.
- **Upstream pollution guard**: `copyCleanHeadersWithReceipt` drops upstream `X-RateLimit-*` and `Retry-After` so the gateway value is authoritative.

## Audit

Three-lane codex audits with iterative convergence:

| Lane | R1 | R2 | R3 |
|---|---|---|---|
| Code | 0/0/0/2L/3N ✅ | — | — |
| Security | 0/**2H**/1M/0L/4N | 0/**1H**/0/0/0 | **0/0/0/0/0** ✅ |
| Architect | 0/0/0/4L/2N ✅ | — | — |

R1 security HIGHs (demo N regression + per-tenant cache leak) and R1 security MEDIUM (upstream header pollution) all fixed in R2; R2 security HIGH (streaming Cache-Control overwrite) fixed in R3.

Audit prompt files: `specs/AUDIT_FIX_ISS_190_R{1,2,3}_*_PROMPT.md`.

## Test plan

- [x] `go test ./... -count=1` → all gateway tests pass locally
- [x] `TestAccountConcurrencyCap` + `TestDemoConcurrencyCap` updated: assert 429 carries the 4-header set, no-store / Vary, and 200 admits carry the request-limit header
- [x] `TestStreamingReceiptHeaderStripped` extended: asserts streaming `Cache-Control` keeps `no-store` AND `X-RateLimit-Limit-Requests` is emitted on the streaming path
- [x] CI green on the PR
- [ ] Operator-side validation: re-run phase-A scenario `02_capacity_contention.yaml` against the new gateway; expect 7×429 (with helpful headers now) + 3×200 (since AccountConcurrency=3) instead of the earlier 7×429 + 2×503 + 1×200

## Out of scope

- Per-API-key tier configuration (would require subscriptions DB schema work; issue acknowledges this as a future extension)
- Completion-time telemetry-driven Retry-After (constant `1s` is documented as conservative pending that work)
- Harness header capture (issue body explicitly tracks this separately)

### DIFF
--- a/phase5-gateway/internal/config/config.go
+++ b/phase5-gateway/internal/config/config.go
@@ -165,7 +165,19 @@ func Default() Config {
 			AccountDailyTokens:        100000,
 			DemoDailyTokensPerIP:      1000,
 			DemoSessionsPerIPPerHour:  10,
-			AccountConcurrency:        2,
+			// Issue #190: AccountConcurrency=3 matches phase-A
+			// network capacity (3 providers × 1 slot each) and the
+			// "user types follow-up while previous reply still
+			// streaming" UX pattern for paying buyers.
+			// DemoConcurrency stays at 2 because M1-8 / PERF-6
+			// documented that 3+ parallel demo requests from one
+			// IP saturate the MLX-serialized provider pool for up
+			// to CoordinatorTimeout — an accidental DoS against
+			// paying buyers. Bumping the demo default to 3 would
+			// re-introduce that regression. Operators can override
+			// either via account_concurrency / demo_concurrency
+			// in gateway.yaml.
+			AccountConcurrency:        3,
 			DemoConcurrency:           2,
 			SignupAccountsPerIPPerDay: 3,
 			ReaperIntervalHours:       1,
--- a/phase5-gateway/internal/router/chat_proxy.go
+++ b/phase5-gateway/internal/router/chat_proxy.go
@@ -51,6 +51,18 @@ func (s *Server) handleChatCompletions(w http.ResponseWriter, r *http.Request) {
 	start := s.now()
 	sw := &statusWriter{ResponseWriter: w, statusCode: 0}
 	w = sw
+	// Issue #190 R1 security HIGH: chat-completion responses now
+	// carry per-tenant headers (X-RateLimit-Remaining-Requests).
+	// Forbid intermediary caching unconditionally so a misconfigured
+	// CDN/proxy cannot serve another buyer's rate-limit state via
+	// a shared cache. Vary covers both auth modes (bearer + demo
+	// token) so any proxy that does cache will at least key
+	// correctly per-tenant.
+	setNoStoreHeaders(w.Header())
+	// Use Add (not Set) so an existing CORS-supplied Vary: Origin
+	// is preserved alongside the auth-mode signals we add here.
+	w.Header().Add("Vary", "Authorization")
+	w.Header().Add("Vary", "X-Demo-Token")
 	var accountID, model string
 	var streamMode bool
 	defer func() {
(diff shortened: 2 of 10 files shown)


antfleet-ops (approved):
LGTM — 3-lane codex audit converged to R3 SECURITY 0/0/0/0/0; full gateway test suite green locally.

PR #193, opened 2026-06-27
fix(coordinator,gateway): align 503 code on 'no_provider_available' (closes #184)

Augustas11:
## Summary

Closes #184.

SPEC-002 v1.4.1 FR-B4 (line 1232) and § 7.2 (line 2388) specify `no_provider_available` as the canonical code for 503 zero-provider responses. The codebase had drifted to also emit `provider_unavailable` on the WS-forward / preflight rejection paths — two different strings for the same conceptual case. External buyer clients matching on `error.code` saw inconsistent enums depending on which rejection path fired.

This PR aligns all paths on the canonical `no_provider_available` per FR-B4 and § 7.2. Refs SPEC-002 v1.4.2 R-5 from PR #192.

## Changes

7 string replacements + 1 test assertion update:

- `phase4-coordinator/internal/buyer/server.go`: lines 1924, 1952, 2087, 2726, 3989 — 5 WS-forward / preflight paths.
- `phase5-gateway/internal/router/chat_proxy.go`: lines 254, 319 — 2 gateway propagation paths.
- `phase5-gateway/internal/router/integration_test.go:319` — test now asserts on the canonical string.

The other coordinator paths (server.go: 3122, 3183, 3661, 3704) were already emitting `no_provider_available` per spec, so they're untouched. `phase5-gateway/internal/router/server_test.go:3538` already negates on `"provider_unavailable"` (testing that coord Tier-2 policy errors are NOT remapped); that assertion stays valid since the gateway no longer emits the wrong string at all.

## Test plan

- [x] `go build ./...` clean on both `phase4-coordinator/` and `phase5-gateway/`.
- [x] `go test ./internal/router/ -run "Provider|Unavailable|NoProvider" -count=1` PASS on gateway.
- [x] `go test ./internal/buyer/ -run "Unavailable|NoProvider|Selected" -count=1` PASS on coordinator.
- [x] `grep -rn '"provider_unavailable"' phase4-coordinator/ phase5-gateway/ --include="*.go" | grep -v _test.go` returns zero matches.
- [ ] (After deploy) Live curl probe against a forced WS-forward 503 scenario returns `error.code: "no_provider_available"`.
- [ ] (Follow-up) Phase-C network-harness scenario `02_capacity_contention.yaml` asserts `error_taxonomy` contains only `no_provider_available` (no `provider_unavailable`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #214, opened 2026-06-28
chore: gitignore phase5-gateway/dist/gateway.yaml

Augustas11:
## Summary
- Add `phase5-gateway/dist/gateway.yaml` to `.gitignore`, mirroring the existing rule for `phase4-coordinator/dist/coordinator.yaml`.
- The deploy script consumes this file at step 0/8 for its C2 cross-config drift check (and intentionally refuses `gateway.yaml.example` as a substitute), so every operator has it sitting untracked locally. On every gateway rebuild, `build-linux.sh`'s dirty guard catches it and forces `FORCE_DIRTY=1` → unnecessary `-dirty-forced` binaries.

## Why this matters
Production gateway was running `v1.6.1-39-gbd0a437-dirty-forced` until today's clean re-deploy. The `-dirty-forced` suffix was structurally correct (the build-time override flag was used as designed), but it was firing on a no-op untracked file, not real local modifications. Result: `/healthz` reads as if prod is running an unaudited override build when in reality it's running merged-and-reviewed code. This rule removes the false signal.

## Test plan
- [x] `git status -- phase5-gateway/` reports clean after this lands and `dist/gateway.yaml` exists locally.
- [x] `git check-ignore -v phase5-gateway/dist/gateway.yaml` matches the new rule.
- [x] `phase5-gateway/scripts/build-linux.sh` no longer trips the dirty guard when only `dist/gateway.yaml` is present → stamps a clean `vX.Y.Z-N-g<sha>` version (verified locally; today's prod gateway re-deploy used the same clean-tree build path via a temporary file-move).
- [x] Mirror rule is already proven in production for the coordinator (line 46) — no behavioral surprises.

## Out of scope
- Coordinator is currently 3 commits behind main (incl. PR #193's coordinator-side change). Separate follow-up.
- The stale MEMORY note claiming `FORCE_DIRTY=1` misses untracked-only dirty state is also wrong; not addressed here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,15 @@ phase5-gateway/dist/gateway-linux-amd64
 # Use coordinator.yaml.example as a template; keep real config local.
 phase4-coordinator/dist/coordinator.yaml
 
+# Phase 5 production config: local snapshot of what's running on Pearl,
+# consumed by dist/deploy-pearl-vps.sh's C2 drift-check gate. Use
+# gateway.yaml.example as a template; keep real config local. Without
+# this rule the file shows as untracked in phase5-gateway/dist/, which
+# trips build-linux.sh's dirty guard and forces FORCE_DIRTY=1 on every
+# rebuild — producing -dirty-forced binaries when none is actually
+# warranted. Mirrors the coordinator.yaml convention above.
+phase5-gateway/dist/gateway.yaml
+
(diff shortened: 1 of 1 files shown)


antfleet-ops (approved):
Mirror of the existing coordinator.yaml convention (line 46). One-file change, body explains the deploy-script C2 dependency that makes the file inherently untracked-per-operator. No behavioral surprises — coordinator pattern has been in tree for months.

PR #229, opened 2026-06-29
fix(harness): I1 drift uses per-pair, not aggregate sums (closes #226)

### LINKED ISSUE #226: Harness reconciler I1: aggregate token-sum check false-positives on high-fallback-outcome scenarios
## Summary

The harness reconciler's I1 (billing-ledger reconciliation drift) check compares aggregate token sums across two different populations — gateway rows with `outcome="ok"` vs all harness "ok" responses — and reports the difference as drift. When the gateway classifies most successful streams as "fallback" outcomes (`stream_truncated`, `stream_output_exceeded`, etc. per SPEC-006 §17.7), this comparison becomes meaningless and I1 false-positives.

Surfaced live on 2026-06-29 e2e run, scenario 07 (`sustained_throughput`):
- harness_completion_tokens     = 233 (across 40 harness "ok")
- gateway_completion_tokens     = 265 (across **5** gateway rows with outcome="ok")
- gateway_completion_tokens_fallback = 2240 (across **35** gateway "fallback" rows)
- coordinator_completion_tokens = 233 (across 40 coord 2xx)

I1 reports drift=32 (=265-233) and flags 2 harness successes as "unmatched on gateway." Both are artifacts: the gateway-side `ok` subset is 5 rows, the harness-side `ok` set is 40 rows — they aren't comparable in aggregate.

## Why this matters

Production money-path is unaffected. The coordinator total (233) matches the harness total (233). The gateway records full settlement for fallback outcomes too — they're just classified separately for SPEC-006 §17.7 reasons.

But: I1 is the headline "money-path doesn't lie" invariant. It false-failing on a legitimate scenario undermines triage signal. Any future real drift will be drowned in noise from this artifact.

## Proposed fix

`internal/reconcile/ledger.go` already produces per-pair matched data via `MatchedSuccesses`. The right fix:

1. Compute drift as the sum of `(GatewayCompletionTokens - HarnessCompletionTokens)` across matched pairs only.
2. Aggregate `gateway_minus_coordinator_tokens` should include `GatewayCompletionTokensFallback` (since coord doesn't get a row for those, the comparison can't use the same denominator on both sides).
3. Drop the population-mismatched sum comparison entirely.

Alternative: keep aggregate sums but compute as `Σ(gateway_completion_ok + fallback) vs Σ(harness all-ok)`. Less precise than per-pair but simpler.

## Acceptance

- Re-run scenario 07 with the fix → I1 PASSes (or fails with a real, per-pair finding).
- Per-pair drift exposed in `ledger_reconcile.json` so triage can see the offending pair if I1 does flag.
- Existing scenarios that PASS today (01-06, 08, 09, 10) must continue to PASS.

## Out of scope

- The 42% non-2xx rate on scenario 07 — separate scenario re-tune issue.
- The X-Provider-Id header gap blocking B5/B6 — separate issue.

Augustas11:
Closes #226.

## Summary

I1's "gateway-coord/gateway-harness token drift" check was comparing population-mismatched aggregate sums:

```
GatewayCompletionTokens         = 265  (5 gateway rows with outcome="ok")
GatewayCompletionTokensFallback = 2240 (35 rows with stream_truncated etc.)
HarnessCompletionTokens         = 233  (all 40 harness "ok" responses)
```

The drift was computed as `GatewayCompletionTokens − HarnessCompletionTokens` = +32, which has nothing to do with money-path correctness — it's just the gap between 5 rows' worth of tokens and 40 rows' worth. The 2026-06-29 production e2e run tripped this on scenario 07 with exactly that shape.

## Fix

Replace aggregate-sum drift with **per-pair drift** computed across `r.MatchedSuccesses` after `matchByFuzzy`:

```go
for _, p := range r.MatchedSuccesses {
    r.GatewayMinusHarnessTokens += p.GatewayCompletionTokens - p.HarnessCompletionTokens
    if p.CoordinatorRequestID != "" {
        r.GatewayMinusCoordinatorTokens += p.GatewayCompletionTokens - p.CoordinatorCompletionTokens
    }
}
```

`matchByFuzzy` already pairs harness successes with gateway rows under `isSettlementComplete` (which includes the fallback outcomes), so the matching itself was correct — only the drift accounting was broken.

Unmatched harness successes continue to surface via `r.UnmatchedSuccesses`, which I1 inspects separately. The "missing on gateway" signal is preserved.

## What lands

- `internal/reconcile/ledger.go`: `computePerPairDrift` helper; replaces the two-line aggregate-sum drift computation. Comments doc-link to #226 and explain the population-mismatch failure mode.
- `internal/reconcile/ledger_test.go` (new): 6 tests covering clean matches, real overcharge, the production scenario-07 shape (40 pairs with matching tokens → 0 drift, was reporting +32), no-coord-match, underbilled, and empty-matches edge cases.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/reconcile/` — 6 new tests pass
- [x] `go test ./...` — full harness suite green
- [x] Production verification: the scenario 07 pattern that tripped I1 on 2026-06-29 (40 successes / 5 gateway-ok / 35 gateway-fallback / 32-token phantom drift) now yields 0 drift in the test reproducing that shape.
- [ ] Re-run scenario 07 against live Pearl after merge — should PASS I1 (defer to a separate e2e run after this lands).

### DIFF
--- a/test/network-harness/README.md
+++ b/test/network-harness/README.md
@@ -38,11 +38,27 @@ We are in phase A.
 These fail the run (exit 10) regardless of any contract:
 
 - **I1** — billing-ledger reconciliation drift == 0
-  (no harness-successful request missing on coordinator request_log or gateway usage_events; coordinator and gateway agree on completion tokens)
+  (per-matched-pair semantics since #226; the `drift_basis` field in
+  `ledger_reconcile.json` pins which algorithm produced the sums.
+  I1 fails on any of:
+    1. unmatched harness successes (`UnmatchedSuccesses`),
+    2. unmatched gateway "ok" rows (`UnmatchedGatewayOKRows`),
+    3. unmatched coord 2xx rows (`UnmatchedCoordinator2xxRows`),
+    4. gateway-ok pairs with no coord row (`MatchedCoordMissing`;
+       fallback outcomes legitimately lack coord rows and are excluded),
+    5. positive gateway-vs-harness overbill across pairs
+       (`GatewayOverbillVsHarnessTokens` > 0; only "ok" outcomes — fallback
+       pairs are excluded per F-8 SSE-undercount artifact, see issue #232),
+    6. absolute gateway-vs-coord mismatch across pairs
+       (`AbsGatewayCoordinatorMismatchTokens` > 0; both directions, since
+       gateway and coord are both settlement systems and must agree).
+  Gateway-vs-harness underbill alone is allowed — gateway-side streaming
+  rounding is legitimate.)
 - **I2** — no 5xx response without a billing settlement entry
   (gateway must echo a request id on every 5xx; orphaning is a billing-bypass risk)
 - **I3** — no charged-tokens > delivered-tokens
-  (phase-A structural check; the DB-level overcharge signal lives in `ledger_reconcile.json` → `token_mismatches`)
+  (phase-A structural check; the DB-level overcharge signal lives in
+  `ledger_reconcile.json` → `overbilled_pairs` and `gateway_overbill_*_tokens`)
 - **I4** — no silent hang
   (a streaming response that stays open past `silent_hang_threshold` with no bytes and no `data: [DONE]` is a UX failure regardless of what the contract says about latency)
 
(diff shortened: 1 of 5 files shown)


antfleet-ops (dismissed):
Approving #229 (#226): reconciler I1 per-pair drift. Replaces population-mismatched aggregate sums with matched-pair drift. 6 new tests cover the production scenario-07 shape (now 0 drift, was phantom 32) + real-overcharge case (still flagged correctly).

antfleet-ops (approved):
Re-approving after R2-R6 audit fixes. 6-round 3-lane codex audit; final state: 0 CRITICAL/HIGH on R6 (one HIGH filed as #232 for follow-up — trust-boundary trade-off, not a blocking bug). 14 reconciler tests + 8 invariant-layer tests; per-pair drift redesign closes #226 production trip.

antfleet-ops (approved):
Re-approving after sync with main.

PR #239, opened 2026-06-29
impl(019): SPEC-019 v0.2 streaming structured output + gateway wall-clock + numeric bounds + $schema

Augustas11:
## Summary

IMPL companion to [#233](https://github.com/Augustas11/macprovider/pull/233) (SPEC-019 v0.2.4 LOCKED). Implements streaming structured output, gateway-owned wall-clock authority, numeric bounds + `$schema` widening, and SDK fixture expansion per AC-V2-1..14 (incl. sub-items 3a, 9b, 10a, 10b).

### Provider (phase3-binary)

- `HTTPServer.swift` — streaming-reject 400 for `json_schema`/`json_object` deleted; v0.2 accepts `stream:true` with structured output.
- `ModelRuntime.swift` — **NEW** end-of-stream validation hook; `StructuredStreamingContentAccumulator` (UTF-8 byte buffer with 2 MiB cap per AC-V2-9b); `StructuredStreamingIdleState` (NSLock-protected idle watcher with `markOperationStopped` barrier); `withStructuredStreamingIdleTimeout` TaskGroup wrapper (markTimedOut → cancel → await operationStopped → snapshot → validate); `synthesizeIdleTimeoutResultOrThrow` helper threads `modelHash` through; bounded-wait helper fails closed on budget breach (Decision C from r3 absorption).
- `JSONSchemaValidator.swift` — numeric-bound widening (`minimum`/`maximum`/`multipleOf` on number/integer nodes); type-conditional gate (AC-V2-10a); value-validity rules (`multipleOf > 0`, JSON-number operands, `minimum ≤ maximum`); `$schema` accepted at top-level only; runtime instance enforcement with `Int64(exactly:)` scaled-integer path (no trap on Int64.max) + FP-fallback saturation guard.
- `InferenceRelay.swift` — WS `errorEndFrame` switch extended to preserve all 4 SPEC-019 codes (`malformed_json_response`, `json_schema_validation_failed`, `response_byte_cap_exceeded`, `provider_timeout`) — no `error_internal` downgrade.

### Coordinator (phase4-coordinator)

[...]

Issue #240, opened 2026-06-29
test: TestReceiptKeysReturnsPreviousKeyInGraceWindow deterministically fails after 2026-06-29 12:00 UTC (clock-mock leak)


Augustas11:
## Summary

`TestReceiptKeysReturnsPreviousKeyInGraceWindow` in
`phase4-coordinator/internal/buyer/receipt_keys_test.go:105` started
failing deterministically on main after **2026-06-29 12:00 UTC**.
Every PR after that wall-clock time hits this failure in the
`phase4-coordinator (go vet + test)` CI job, blocking unrelated merges.

## Root cause

The test sets:

```go
expiresAt := time.Date(2026, 6, 29, 12, 0, 0, 0, time.UTC)
// ...
server := NewServer(registry, zerolog.Nop(), time.Unix(1716768000, 0))
frozen := time.Date(2026, 6, 23, 12, 0, 0, 0, time.UTC)
server.now = func() time.Time { return frozen }
```

`server.now` is overridden to `2026-06-23 12:00 UTC` (inside the grace
window). But the `receipt_pubkey_prev` is now nil at line 136 —
suggesting somewhere in the receipt-keys serialization path a clock
read is NOT going through `server.now` and instead uses
`time.Now().UTC()`. Once the real wall clock crossed `expiresAt`, the
non-mocked clock check evicts the previous key from the response.

## Reproduction

```
cd phase4-coordinator
go test -run TestReceiptKeysReturnsPreviousKeyInGraceWindow ./internal/buyer/
```

Fails on `main` (commit `963fbf3`) after 2026-06-29 12:00 UTC.

## Proposed fix

Either:
1. Audit the serialization path and route every clock read through
   `server.now` (or a shared injectable clock).
2. Move `expiresAt` to a relative time from `frozen` (e.g. `frozen.Add(7*24*time.Hour)`)
   so the test self-anchors and doesn't decay.

(2) is simpler and forward-compatible.

## Impact

Pre-existing main breakage from `2026-06-29 12:00 UTC` onward.
Currently blocking PR [#238](https://github.com/Augustas11/macprovider/pull/238)
(harness reconciler R6) from auto-merge. #238 is being admin-merged
on the basis that the failure is unrelated to its diff
(`test/network-harness/` only). Future PRs touching unrelated code
will hit the same wall.

## Acceptance

- The failing test re-anchors `expiresAt` to a relative time (or fixes
  the clock-mock leak in the serialization path).
[...]

PR #241, opened 2026-06-29
fix(coordinator): bump receipt-keys grace-window test dates to 2099 (defuse 2026-06-29 12:00 UTC time-bomb)

Augustas11:
## Summary

`TestReceiptKeysReturnsPreviousKeyInGraceWindow` (introduced by [#124](https://github.com/Augustas11/macprovider/pull/124) — SPEC-015 receipts) started failing on every run after 2026-06-29 12:00 UTC. **It's currently blocking `ci-required` for every PR in the repo** (including [#239](https://github.com/Augustas11/macprovider/pull/239) which discovered this).

## Root cause

`Registry.Register` (called through `registerReceiptKeyProvider`) uses real `time.Now()` internally to gate `activeReceiptPubkeyPrev` at `phase4-coordinator/internal/pool/provider.go:550`. That gating runs **before** the test's `server.now` override takes effect — so the registry drops the previous key the moment real wall-clock crosses the hardcoded `expiresAt: 2026-06-29 12:00 UTC`.

The test only passed for ~6 days from when it landed (2026-06-23) to today.

## Fix

Bump the hardcoded date to `2099-06-29`. The 2099 date makes the time-bomb a non-issue for >70 years.

This is the smallest of three options:
1. **(this PR)** Bump the date to a far future year.
2. (better) Thread a clock parameter through `registerReceiptKeyProvider`.
3. (best) Add optional clock injection on `Registry.Register`.

(2) and (3) are bigger surface changes that should land on their own merits, not as side-effects of unblocking CI.

Inline comment in the test now documents the registry-clock-vs-`server.now` race so the next person to touch this code knows why the date is specifically in the far future.

## Test plan

- [x] `cd phase4-coordinator && go test ./internal/buyer/... -count=1` — green (includes all receipt-keys variants)
- [ ] CI green on this PR (will unblock PR #239 once merged)

### DIFF
--- a/phase4-coordinator/internal/buyer/receipt_keys_test.go
+++ b/phase4-coordinator/internal/buyer/receipt_keys_test.go
@@ -106,15 +106,22 @@ func TestReceiptKeysReturnsPreviousKeyInGraceWindow(t *testing.T) {
 	registry := pool.NewRegistry(nil)
 	current := bytes.Repeat([]byte{0x42}, 32)
 	previous := bytes.Repeat([]byte{0x43}, 32)
-	rotatedAt := time.Date(2026, 6, 22, 12, 0, 0, 0, time.UTC)
-	expiresAt := time.Date(2026, 6, 29, 12, 0, 0, 0, time.UTC)
+	// `Registry.Register` (called by `registerReceiptKeyProvider`) uses
+	// real `time.Now()` internally to gate `activeReceiptPubkeyPrev`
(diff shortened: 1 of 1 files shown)


Augustas11:
Superseded by #240 (already merged with a stronger fix — new time-aware registerReceiptKeyProviderAt helper rather than just a date bump). Closing without merge.

Issue #248, opened 2026-06-29
harness reconciler: late-settlement leak past 30s forward pad


Augustas11:
## Problem

Surfaced by 3-lane codex audit on #243 (R1 — security HIGH-2, code LOW-2).

The reconciler's snapshot window uses a forward pad of 30s
(`endUTC + 30s`). Live v0.3 baseline data shows real gateway
settlement lags of 26–76s. Settlements landing past `endUTC + 30s`
either evade I1 entirely (on the last scenario in a run) or get
misattributed to the next scenario's orphan pool when one exists.

## Audit context

Security lane verdict on #243:
> the retained `endUTC + 30s` upper bound is already shorter than
> observed settlement lag. Since live data shows 26-76s lag, orphan
> or extra billing rows after `endUTC+30s` can evade the final
> scenario's I1 window completely, or be misattributed to the next
> scenario when one exists.

Code lane LOW agrees:
> 30s forward pad is known inadequate, but enlarging it is not the
> right fix here. […] do not treat `30s` as the durable contract;
> solve late settlement with request-ID ownership or a
> settlement-quiescence/watermark mechanism in the separate issue.

#243 fixes the cross-scenario double-count for harness-owned rows
via ID-based recovery. This issue tracks the residual gap for late
NON-owned settlements (true orphans / background traffic landing
after the pad).

## Candidate fixes

1. **Per-scenario quiesce**: sleep N seconds between scenarios so
   settlements drain before the next scenario starts. Cheap, but
   slows runs and doesn't help last-scenario coverage.
2. **Settlement watermark**: have the gateway expose a
   "settled-through" watermark; the reconciler waits for the
   watermark to advance past `endUTC` before snapshotting. Most
   rigorous; requires gateway change.
3. **Run-wide reconcile pass**: in addition to per-scenario, do
   one final reconcile across the whole run window after all
   scenarios complete. Catches stragglers, but loses per-scenario
   attribution for them.
4. **Adaptive forward pad**: grow the pad based on observed lag
   percentile (e.g. p99 of last N runs + safety factor). Defers
   the problem rather than eliminating it.

## Acceptance

- Decision recorded in `beta/DECISION_CRITERIA.md`.
- Chosen mechanism implemented + tested.
- v0.3 baseline rerun shows no late-settlement orphan attribution
  to adjacent scenarios.

## Links

- Surfacing PR: #243 (helper comment in `snapshotWindow` points
  back here
[...]

PR #249, opened 2026-06-29
fix(harness): close late-settlement leak in I1 reconciler snapshot window (closes #243, #248)

### LINKED ISSUE #243: harness reconciler: snapshot window backward pad creates false-positive orphans when scenarios run back-to-back
## Summary

`reconcile.Run` pads the snapshot SQL query window by 30s on BOTH sides:
`winStart := startUTC.Add(-30s)`, `winEnd := endUTC.Add(30s)`. When two
harness scenarios run back-to-back (e.g. scenario 03 right after 02),
the backward pad pulls tail gateway/coord rows from the prior
scenario's window into the next scenario's reconcile. The same row gets
reconciled TWICE — cleanly in its actual scenario, as a false-positive
orphan in the next.

## Reproduction (v0.3 rerun, post-#238)

Three orphan rows surfaced in the rerun were all already cleanly
matched in the prior scenario:

| current scenario | orphan id | actually matched in | comment |
|---|---|---|---|
| 03 sticky_multi_turn | `de5b9b21-fe0d-414d-9079-31d6b9e19a70` | 02 capacity_contention | clean match in 02 |
| 09 streaming_ttft | `3ea18bd7-1795-44c6-b773-c4d476f05726` | 08 cold_warm_compare | clean match in 08 |
| 09 streaming_ttft | `7bf5ffa0-522f-4b37-b8bd-48b9b17554fd` | 08 cold_warm_compare | clean match in 08 |

All have `account_id == harness_account_id` (the same account), so the
existing account-consensus filter does not exclude them.

## Why the backward pad is wrong

Gateway settlement happens AFTER the harness sends and the response
completes. There is no scenario where a row's `created_at` is BEFORE
the harness's `startUTC` for a request the harness sent.

The 30s forward pad is correct: settlement is async with respect to
the response, so it can land seconds after the harness records the
response as complete.

## Fix (this PR)

Extract a small helper `snapshotWindow(startUTC, endUTC)` that returns
`(startUTC, endUTC + 30s)` — no backward pad. Regression test
`TestSnapshotWindow_NoBackwardPad` pins the contract.

Expected impact on v0.3 baseline: false-positive orphans on 03, 05, 09
should disappear; 07's 39 coord-2xx orphans likely shrink to the real
F-8 fallback shape (#232). The gateway counting offset on 07 OK pairs
remains real and is a separate concern.

Augustas11:
## Summary

Three-part fix to the harness's I1 reconciler snapshot window, surfaced by the v0.3 baseline rerun and converged across 3 rounds of 3-lane codex audit. Closes #243 and folds in #248 (initially filed as a deferred follow-up, then folded in after user pushback — same surface, cheap inline fix).

### What broke

1. **Cross-scenario false-positive orphans.** When scenarios ran back-to-back (e.g. 02 ENDs 13:07:47, 03 STARTs 13:07:47), the 30s backward pad pulled tail gateway rows from 02 into 03's reconcile. Same row reconciled twice — once cleanly, once as a phantom orphan on the next scenario's I1.
2. **Clock-skew dropped harness-owned rows.** Gateway `created_at < startUTC` (sub-second clock skew) was dropped by the SQL window predicate, then surfaced as false-positive `UnmatchedSuccesses` (R2 HIGH from security + code lanes).
3. **76s-late settlements rejected by matcher even when in snapshot.** v0.3 live data showed gateway settlements lagging 26–76s. The original 30s forward pad missed them; bumping to 90s + adding 90s quiesce got them into the snapshot pool, but `matchWindow = 60s` then rejected them at match-time — same false-positive UnmatchedSuccesses class (R3 HIGH, 3-of-3 lane convergence).

### What changed

- `snapshotWindow`: backward pad dropped, forward pad 30s → 90s, anchored on new exported `reconcile.SnapshotForwardPad`.
- `queryGatewayByIDs` / `queryCoordinatorByIDs`: by-ID recovery pass for harness-owned request IDs, bypassing the lower-bound time predicate. Composite-key dedup merge.
- `exactMatchSettleWindow`: per-request settle tolerance for exact-id matches, sized to `SnapshotForwardPad`. `matchWindow` (60s) stays as the FUZZY guard against accidental token+timestamp false positives.
- `cmd/harness/main.go`: 90s context-aware quiesce sleep before reconcile snapshot. `select { time.After(d); ctx.Done() }` so SIGINT exits promptly.

### Audit trajectory

- R1 (3-lane codex): 2 HIGH (clock-skew lower bound; forward pad inadequate) + 1 MEDIUM (test contract).
- R2 (3-lane codex): 0 C/H/M on by-ID recovery + merge; LOWs only.
- R3 (3-lane codex): 1 HIGH 3-of-3 convergent on the matcher window gap; 1 MEDIUM on cancellable quiesce; fix inline. No R4.

### Tests (all green)

- `TestSnapshotWindow_NoBackwardPad` (R0)
- `TestRun_IDRecoveryRescuesClockSkewedRows` (R2 HIGH)
- `TestRun_NoDoubleCountingAcrossAdjacentScenarios` (R2 MEDIUM)
- `TestSnapshotWindow_BoundaryNanoseconds` (R2 LOW)
- `TestRun_LateSettlement_ExactIDMatchesPastFuzzyWindow` (R3 HIGH)
- `TestHarnessRequestIDs_DedupAndFilter` (unit)

## Test plan

- [x] `go test ./...` in `test/network-harness/` — all packages green
- [x] `go build ./...` clean
- [ ] Post-merge: re-run v0.3 baseline; expect 1–6 + 8 cleanly pass (residual is scenario 07 gateway-side +9 tok counting offset, a real money-path bug, NOT a harness issue)

### DIFF
--- a/test/network-harness/cmd/harness/main.go
+++ b/test/network-harness/cmd/harness/main.go
@@ -137,6 +137,29 @@ func cmdRun(args []string) error {
 	hasRemoteDBs := sc.Target.CoordinatorDBSSH != "" && sc.Target.GatewayDBSSH != ""
 	switch {
 	case hasLocalDBs || hasRemoteDBs:
+		// Quiesce before snapshotting: gateway settlement can lag the
+		// harness's response observation by tens of seconds (v0.3
+		// baseline data: 26–76s). Sleep until endUTC +
+		// reconcile.SnapshotForwardPad before pulling the SQLite
+		// snapshot so late settlements have landed. The pad constant
+		// is shared with snapshotWindow's SQL upper bound and the
+		// exact-ID matcher tolerance — they must move together.
+		//
+		// Sleep is context-aware: SIGINT/SIGTERM during quiesce
+		// cancels promptly via the signal-notify context set up at
+		// the top of main().
+		quiesceTarget := meta.EndUTC.Add(reconcile.SnapshotForwardPad)
+		if d := time.Until(quiesceTarget); d > 0 {
+			log.Printf("scenario %q: quiescing %v for gateway settlement before reconcile", sc.Name, d.Round(time.Second))
+			timer := time.NewTimer(d)
+			select {
+			case <-timer.C:
+			case <-ctx.Done():
+				timer.Stop()
+				log.Printf("scenario %q: quiesce interrupted by context cancel", sc.Name)
+				return ctx.Err()
+			}
+		}
 		// Money-path gate: when DBs are configured, the harness asked
 		// for reconciliation. If reconcile.Run errors (DB schema
 		// mismatch, snapshot fetch failure, etc.) we must NOT silently
--- a/test/network-harness/internal/reconcile/ledger.go
+++ b/test/network-harness/internal/reconcile/ledger.go
@@ -44,13 +44,34 @@ import (
 	_ "modernc.org/sqlite"
 )
 
-// matchWindow is the per-request temporal tolerance for accepting a
-// candidate row as a match for a harness observation. Used by both
-// exact-id and fuzzy paths so they cannot silently diverge. The
-// reconciler's SQL snapshot pads the run window by 30s; this tighter
-// per-request window further constrains which rows are eligible.
+// SnapshotForwardPad is the canonical forward pad applied to the
+// reconciler's SQL snapshot window AND the runner's pre-snapshot
+// quiesce sleep (cmd/harness/main.go). They MUST share the same
+// value: the runner sleeps long enough for settlements to land, and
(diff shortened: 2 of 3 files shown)


antfleet-ops (approved):
harness reconciler — three-part fix on a single contract. R1→R3 audit lanes converged each round (no new convergent findings post-R3); inline regression tests cover each lane finding. Approving.

antfleet-ops (approved):
harness reconciler — three-part fix on a single contract. R1→R3 audit lanes converged each round; inline regression tests cover each lane finding. Approving.

PR #280, opened 2026-06-30
fix(#245): SPEC-007 v0.5 — reject untyped /admin/explorer/sessions path-segments with 400

Augustas11:
Closes #245. SPEC-007 v0.5 — flips the v0.4 deprecation-window path-segment typing gate to break mode: untyped (legacy bare-id) calls to `/admin/explorer/sessions/{id}` now return `400 invalid_request_error` + `code: session_id_untyped` on BOTH the gateway (`ext_` required) and the coordinator (`int_` required).

## Trigger (#245 gate)

Issue #245 names three trigger conditions for opening this PR; only option **(2) operator UI cutover** applies today:

| # | Trigger | Status |
|---|---|---|
| 1 | 14d telemetry-quiet on `payout_explorer_path_segment_untyped` | NOT fired — only ~1d since v0.4 merged |
| 2 | Operator UI cutover | **Satisfied by this PR** — dashboard.js `linkFor` emits `int_`-prefixed URLs on both `request_id` columns and `link_target: session:` rows; this is the only first-party operator UI |
| 3 | Forced 90-day cutoff (~2026-09-27) | NOT fired — way too early |

The SPEC v0.5 change-log row records this explicitly so future archaeology won't infer telemetry-quiet elapsed.

### IMPL — gateway + coordinator handlers
- `phase5-gateway/internal/router/explorer.go::handleExplorerSessionDetail` returns 400 + envelope (`invalid_request_error` / `session_id_untyped`) when `parseTypedSegment(rawSegment, "ext_")` reports `!typed`. The v0.4 `payout_explorer_path_segment_untyped` `audit_events` emit is removed (no longer reachable).
- `phase4-coordinator/internal/explorer/handlers.go::handleSessionDetail` returns 400 + same envelope when `strings.CutPrefix(requestID, "int_")` returns `!ok || stripped==""`. `Handler.logPathSegmentUntyped` deleted. Envelope shape (`type`, `code`, `message`, `source`, `retryable`) symmetric with gateway emit per R1 SEC LOW-1 closure.

### IMPL — dashboard.js (operator UI cutover)
- `phase4-coordinator/internal/explorer/static/js/dashboard.js::linkFor` emits `/admin/explorer/sessions/int_${encodeURIComponent(v)}` for both `request_id` columns and `link_target: session:` rows so coordinator-dashboard navigations match the v0.5 contract.

### IMPL — coordinator-driven gateway proxy
[...]

PR #277, opened 2026-06-30
fix(#274): canonical config.ValidateProviderID across all registration paths

Augustas11:
Closes #274.

Surfaced by codex SECURITY audit on PR #275 (#266 Tranche 3 R1 SEC-L1). The `pool.Provider.SortKey()` method introduced by #266 T3a returns `ProviderID + "/" + AssignedID` and is used as a map key across the routing pipeline (excluded set, balanced-score cache, faulted-route tracking). The "/" delimiter is only unambiguous when no ProviderID contains "/".

Pre-#274:
- Configured pinned providers: gated on `providerIDPattern = ^[a-zA-Z0-9_.-]{1,64}$` via `config.validateProviders`
- WS self-serve (Hello, auth_request initial/proof): only `requireString` (non-empty + non-control-char)
- Admission `IssueToken`, `MintAdmissionTokenAndPairOT`: only `strings.TrimSpace` + non-empty

Current AssignedID values are coordinator-issued UUIDs, so a practical collision was not exploitable today — but the validator drift was real and this PR consolidates the gate at every entry point.

### Canonical validator
Exports `config.ValidateProviderID(s string) error` backed by `providerIDPattern`. Existing `config.validateProviders` refactored to use it.

### Applied across all 5 registration paths
- `internal/ws/messages.go::ParseHello`
- `internal/ws/messages.go::parseAuthInitial` (auth_request stage=initial)
- `internal/ws/messages.go::parseAuthProof` (auth_request stage=proof)
- `internal/auth/tokens.go::IssueToken`
- `internal/auth/tokens.go::MintAdmissionTokenAndPairOT`

### SortKey-side invariant doc
`pool.Provider.SortKey` docstring now names the no-slash precondition explicitly and points to `config.ValidateProviderID` so future changes to SortKey or AssignedID format will find the invariant.

### Regression tests
- `internal/config/iss274_provider_id_validator_test.go` — accept/reject table including the `"a/b"` delimiter-collision seed
- `internal/ws/iss274_provider_id_validator_test.go` — ParseHello + ParseAuthRequest initial/proof
- `internal/auth/iss274_provider_id_validator_test.go` — IssueToken + MintAdmissionTokenAndPairOT, including leading/trailing whitespace rejection (R1 CODE LOW-1)

## Audit convergence

Per [[feedback-three-lane-codex-audits]]: separate per-lane prompts at `specs/AUDIT_ISS274_{CODE,SECURITY,ARCHITECT}_PROMPT.md`.

[...]

PR #282, opened 2026-06-30
test(#281): TestAC18 — warmup + median (resolves shared-runner timing flake)

### LINKED ISSUE #281: Flaky: TestAC18_TimingEquivalenceRows5_6_7 (20% timing-variance assertion on CI shared runners)
## Summary

`TestAC18_TimingEquivalenceRows5_6_7` (`phase4-coordinator/internal/stats/handlers_integration_test.go:1010`) flakes on the `phase4-coordinator (stats Postgres AC-9/10/19/20)` GitHub Actions job. The test asserts that per-row query timings (`row5`, `row6`, `row7`) vary by no more than **20%** across three SELECTs; on shared CI runners the variance routinely exceeds that and the assertion fails.

## Evidence

Two consecutive PRs in the last 24h hit the same failure on the first run:

| PR | Date | Observed variance | Failed job | Rerun result |
|---|---|---|---|---|
| [#277](https://github.com/Augustas11/macprovider/pull/277) (iss274) | 2026-06-30 | row5=861µs row6=822µs row7=601µs → **30.1%** | [run 28439740423 / job 84274375033](https://github.com/Augustas11/macprovider/actions/runs/28439740423/job/84274375033) | `gh run rerun --failed` → CLEAN |
| [#280](https://github.com/Augustas11/macprovider/pull/280) (iss245) | 2026-06-30 | (similar) | [run 28450071039 / job 84309632508](https://github.com/Augustas11/macprovider/actions/runs/28450071039/job/84309632508) | rerun pending |

Neither PR touches `internal/stats/` — the test does not depend on either PR's code.

## Why it flakes

The assertion shape `|max - min| / max > 0.20` is timing-equivalence-by-magnitude on three near-identical SELECTs in microsecond range (~600µs–900µs observed). A single GC pause, page-cache miss, or noisy-neighbor IOPS spike on a shared GitHub Actions runner produces variance well above 20% even when the underlying query plans are identical. The flake rate appears to be ~50% on cold runners (two-for-two so far).

## Suggested fix

Pick one (or compose):

1. **Raise threshold to 100%** (i.e. assert `max <= 2 * min`). The 20% gate was chosen to catch "this row uses a SCAN while that row uses an INDEX" — a 2× threshold still catches that class while tolerating runner noise.
2. **Median of N runs** — repeat the timing block 5× and compare medians, not single-sample timings. Robust to one-off spikes.
3. **Switch from wall-clock to a structural assertion** — `EXPLAIN QUERY PLAN` for each row and assert all three plans contain the same access method (no `SCAN`, or all `USING INDEX <name>`). Tests the actual invariant (query-plan equivalence) rather than its proxy.

Option (3) is the most principled — option (1) is the smallest patch.

## Out of scope

Other timing-sensitive assertions in `internal/stats/` are not in scope of this issue; this is the one observed flaking. If they start tripping too, broaden the same fix pattern.

## Operational impact

Money-path PRs (#266 T1/T2/T3, #274, #245) consistently green-after-rerun. The cost is one extra `gh run rerun --failed` per affected PR plus polling time. Fixing this removes ~5min of friction per PR until then.

Augustas11:
Closes #281. The AC-18 three-way auth-failure timing-equivalence test was flaking at ~30% variance on shared CI runners (caught on PRs #277 + #280 within 24h). Root cause: cold-start systematic bias — handler/JIT/Postgres-prepared-statement-cache/Go-connection-pool stabilization noise pulled the recorded median around in the first 5–10 samples per row.

### `measure()` discards 10 warmup samples before recording 100

```go
// Warmup loop (10 samples, discarded) — stabilises Postgres
// prepared-statement cache, Go JIT for the auth path, HTTP
// connection-pool reuse.
for i := 0; i < Warmup; i++ {
    oneSample()
    time.Sleep(225 * time.Millisecond)
}
// Recorded samples (100, median is the gate)
samples := make([]time.Duration, 0, N)
for i := 0; i < N; i++ { ... }
sortDurations(samples)
return samples[N/2]
```

Total test runtime: `(10 + 100) × 225ms × 3 rows ≈ 74s` (up from 67.5s pre-#281).

### Why median, not min

The R1 patch (commit `d748326`) tried min-of-100. R1 SEC audit caught it as a MEDIUM: switching to min weakens the actual security property. An attacker measures wall-clock medians, not floor execution time. A future timing leak where row 6 takes 1ms for the first 5 warmed samples but 3ms for the rest would pass a min-based test while still being attacker-distinguishable. Reverted to median in R1 fix-pass (`5035600`).

## Audit convergence

Per [[feedback-three-lane-codex-audits]]: separate per-lane prompts in `specs/AUDIT_ISS281_{CODE,SECURITY,ARCHITECT}_PROMPT.md`.

| Round | CODE codex | SECURITY codex | ARCHITECT codex |
|-------|------------|----------------|-----------------|
| R1 | 0/0/0/1/0 ACCEPT | 0/0/**1**/0/1 (min weakens security property) | 0/0/0/2/3 ACCEPT |
| R2 | _skipped_ | **0/0/0/1/2 ACCEPT** | _skipped_ |

Per [[feedback-skip-accepted-audit-lanes]]: only SEC rerun on R2 since the fix-pass changed only the statistic choice (squarely in SEC scope). Per [[feedback-stop-iterating-on-low-audits]]: stop after R2 ACCEPT; don't fix LOWs and re-fire.

R1 fix-pass: `5035600`.

## Deferred (LOWs, not blocking)

- **R2 SEC LOW-1 / R1 ARCH LOW-2 — row interleaving**: the test still samples row 5, then row 6, then row 7 sequentially. Per-row warmup eliminates cold-start bias, but monotonic runner drift across the ~74s test could still introduce a row-order effect. Recommended follow-up: warm all rows first, then interleave measured samples in balanced or randomized row order while preserving 225ms pacing. Not blocking — warmup alone should resolve the observed flake. If interleaving is needed, this fix will surface it within a few CI cycles and a follow-up PR is cheap.
- R2 SEC INFO-2 / R1 SEC INFO-1 confirm the underlying handler IS structurally constant-time pre-DB-lookup (no bcrypt/dummy-verify path exists in the scoped auth code; all three rows share SHA-256 + `LookupPartnerKeyByHash` before branching). The statistical test is checking the post-lookup branching paths, where any leak would necessarily live.

## Test plan

- [x] `go vet ./internal/stats/...` clean
- [x] `go build ./...` clean
- [ ] CI integration job (`phase4-coordinator (stats Postgres AC-9/10/19/20)`) green on first run — this is the test PR is trying to fix
- [ ] **Empirical check**: monitor the next 5 CI runs across unrelated PRs; if `TestAC18` flakes again, fall back to row-interleaving follow-up

(Could not repro locally — testcontainers requires Docker; not running here.)

### DIFF
--- a/phase4-coordinator/internal/stats/handlers_integration_test.go
+++ b/phase4-coordinator/internal/stats/handlers_integration_test.go
@@ -928,18 +928,21 @@ func TestStale503NotDebitedFromSuccessBucket(t *testing.T) {
 
 // AC-18 — three-way timing equivalence rows 5/6/7. Each row
 // MUST take ≤270 rpm (below the 300 rpm auth-failure cap) and
-// produce 401 with floor latency within ±20% of the other two rows.
+// produce 401 with median latency within ±20% of the other two rows.
 //
-// #281: We compare the MINIMUM of 100 samples per row, not the
-// median. The minimum represents the floor of handler execution —
-// what the path actually costs when no ambient noise (GC pauses,
-// page-cache misses, container CPU throttling) is present. Noise
-// can only push timings UP, never DOWN, so the min is genuinely
-// stable across shared CI runners. The security invariant the
-// test pins ("no fingerprintable difference between failure modes
-// 5/6/7") is correctly measured by the floor: an attacker cannot
-// probe below the handler's actual cost. Median of 100 was tried
-// pre-#281 and flaked at ~30% variance on cold runners.
+// #281: Pre-#281 the median of 100 samples was flaking at ~30%
+// variance on shared CI runners. R1 SEC audit flagged that
+// switching to min weakens the security property — an attacker
+// measures wall-clock medians, not floor execution time. So a
+// future leak where row 6 takes 1ms for the first 5 warmed samples
+// but 3ms for the rest would pass a min-based test while still
+// being attacker-distinguishable.
+//
+// R1 fix-pass: keep the median (attacker-relevant statistic), but
+// discard the first 10 samples per row as warm-up so handler/JIT/
+// connection-pool stabilisation noise doesn't pull the recorded
+// medians around. Median of 100 warmed samples is robust enough on
+// shared CI without trading away the security invariant.
 func TestAC18_TimingEquivalenceRows5_6_7(t *testing.T) {
 	fx, adminDB := setupRollupFixture(t)
 	seedFreshOverview(t, adminDB)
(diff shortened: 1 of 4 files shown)


antfleet-ops (approved):
AC-18 timing-flake fix. CODE/ARCH R1 ACCEPT, SEC caught R1 min-as-statistic as a MEDIUM (correct — attacker observes median, not floor), R1 fix-pass reverted to median + added 10-sample warmup. SEC R2 ACCEPT 0/0/0/1/2; row-interleaving LOW deferred per stop-iterating-on-low-audits. LGTM.

Issue #295, opened 2026-06-30
gateway: terminalSSEErrorCode does not enforce standalone-envelope / last-frame position


Augustas11:
## Summary

PR for #232 added a buyer-visible terminal SSE error envelope contract in SPEC-006 §17.7.1 (standalone data frame, last frame before `[DONE]`, no content after). The harness enforces these rules client-side as a corroboration check.

The GATEWAY side has a related but distinct hole: `terminalSSEErrorCode` at [phase5-gateway/internal/router/chat_proxy.go:1155](https://github.com/Augustas11/macprovider/blob/main/phase5-gateway/internal/router/chat_proxy.go#L1155) only extracts `error.code` from the JSON payload. It does NOT check:

1. **Standalone shape** — a payload containing `choices`/`usage` ALONGSIDE `error.code` still triggers the terminal-error code path.
2. **Last-frame position** — after detecting a terminal frame, the gateway can keep forwarding subsequent content frames via [chat_proxy.go:648](https://github.com/Augustas11/macprovider/blob/main/phase5-gateway/internal/router/chat_proxy.go#L648).

Surfaced by [#232](https://github.com/Augustas11/macprovider/issues/232) R6 ARCH lane HIGH.

## Threat

A malicious or buggy provider can emit a payload like:

```json
{"choices":[{"delta":{"content":"hello"}}],"usage":{...},"error":{"code":"provider_timeout","type":"api_error"}}
```

The gateway's `terminalSSEErrorCode` returns `provider_timeout` (truthy), routes through the refund-only / pass-through code path, then keeps forwarding subsequent content frames. The BUYER sees content + a terminal envelope + more content + `[DONE]`, which is structurally inconsistent and violates SPEC-006 §17.7.1's clauses 1 (standalone) and 3 (last frame).

## Fix shape

`terminalSSEErrorCode` should:

1. Reject payloads with non-empty `choices` OR non-zero `usage` token counts (matches SPEC-006 §17.7.1 clause 1).
2. After detecting a terminal frame, the gateway must STOP forwarding subsequent data frames before `[DONE]` (matches §17.7.1 clause 3 — no content after envelope).

Both rules should be enforced in the gateway's SSE forwarder, not just in downstream consumers (defense in depth — the SPEC contract applies to both sides of the wire).

## Triggering conditions

Open when:

1. Provider audit / red team finds a payload shape exploiting the gap.
[...]

PR #296, opened 2026-07-01
fix(#232): harness fallback-overbill corroboration via SPEC-006 §17.7.1 SSE contract

### LINKED ISSUE #232: Harness reconciler: fallback-outcome overbill suppression is a new trust boundary
## Summary

PR #229 R5 (fallback pairs excluded from `GatewayOverbillVsHarnessTokens`) trades one detection regression for another:

- **Before R5**: fallback pairs counted as overbill → false-failed I1 on the production #226 shape (gateway settles streams as `stream_output_exceeded` with bytes-emitted > harness SSE-parsed tokens).
- **After R5**: fallback pairs are excluded from the overbill signal → gateway outcome classification becomes a trust boundary. A gateway bug (or attacker) that mislabels a real overbill as `stream_truncated` etc. hides it from I1.

The current behavior is the right one for known-good gateway settlements; the F-8 SSE-undercount mismatch is a real artifact that would otherwise drown I1 in false positives. But the trust shift is worth tracking.

## Surfaced by

PR #229 R6 codex audit (security lane HIGH). The audit lane that lets you spot exactly this kind of "the fix becomes a new trust boundary" pattern.

## Current state (as of 2026-06-30)

Two adjacent PRs landed after #229 changed the picture but did NOT close the trust gap:

- **[PR #286](https://github.com/Augustas11/macprovider/pull/286)** (`aab30c1`) extended the same R5 fallback-exclusion pattern to the gateway-vs-coordinator axis (`GatewayOverbillVsCoordinatorTokens` + `AbsGatewayCoordinatorMismatchTokens`). The trust gap now applies to BOTH axes.
- **[PR #288](https://github.com/Augustas11/macprovider/pull/288)** (`a3149e3`, Wave 0a) flipped gateway settlement to trust provider `usage.completion_tokens` on clean SSE streams instead of byte-estimating. Empirically this shrinks the production fallback surface (the 15 fresh `gpt-oss-20b` `stream_output_exceeded` rows in 6 minutes that Entry 94 found should largely become `ok`/`provider_reported` instead).

Net: **the trust gap is logically identical (gateway outcome label still gates I1 overbill checks) but the production surface where fallbacks happen is now much smaller** — most traffic now flows through the OK path that I1 actually checks.

## Proposed

Possible fixes — none urgent, all need product/security design first:

1. **Harness-side evidence of truncation**: only suppress overbill when the harness ALSO saw a truncation marker (`finish_reason: "length"`, abrupt connection close, etc.). Pair-by-pair the suppression with harness-side corroboration.

2. **Suspicious-delta signal**: keep the fallback suppression for I1 but emit a `FallbackOverbillSuspiciousTokens` reference field. Triage scripts can flag scenarios with abnormally large fallback deltas as needing manual review. Lowest-risk; additive only.

3. **Per-outcome tolerance**: instead of binary include/exclude, allow fallback overbill up to a tolerance (e.g. ≤2× harness tokens). Above the tolerance, still flag. Encodes "F-8 undercount has a known bound."

(2) is probably the right first cut now that Wave 0a shrunk the production fallback surface — additive triage signal without risking false-fails on legitimate F-8 cases. (3) is the eventual destination once the tolerance bound is empirically calibrated. (1) requires harness-side data that may not be consistently captured.

## Acceptance

- Decide on fix shape (1/2/3 or combination).
- Implement + cover with regression test that a malicious-shaped fallback pair (gateway=999, harness=8) gets surfaced.
- Update I1 docs to drop the unconditional fallback-exclusion phrasing.

## Out of scope here

Not blocking #229 merge — the R5 trade-off is a strict improvement over the R1 state (false-fail on real production shape). Wave 0a (PR #288) reduces practical exposure further.

Augustas11:
Closes #232.

## Summary

Adds a buyer-visible corroboration gate to the harness reconciler's fallback-overbill suppression. Before this change, `computePerPairDrift` suppressed any gateway-labeled fallback outcome from the I1 overbill invariant — trusting the gateway's outcome label alone. That was the trust gap #232 flagged: a buggy or malicious gateway could label a row as `stream_truncated` (or any §17.7 fallback outcome) without having actually presented the buyer with a truncated stream, and harness would silently suppress the drift.

The fix requires the buyer to independently corroborate the truncation via the SPEC-006 §17.7.1 OpenAI-style terminal error envelope (or matching named-mapping exception) before the harness suppresses the drift. Uncorroborated fallback claims now flow into I1 as real overbill drift signals.

## What changed

**Harness parser (`buyer/loadgen.go`):**
- Fully event-boundary SSE parser (WHATWG dispatch model). Data payloads buffer into `eventBuf`; classification happens ONLY on blank-line dispatch. Pending events on EOF are discarded per spec.
- Generic WHATWG field parser: split at first colon; `data:` and bare `data` (no colon) both count as data field.
- `event:` field tracking + `isDefaultEventName` gate on corroboration: only default (empty) or `event: message` events can flip `SawSSEErrorEvent`. `event: thread.*` (Assistants API) and `event: response.*` (Responses API) route through non-chat-completion SDK handlers and no longer corroborate.
- `[DONE]` terminator uses `bytes.HasPrefix` for OpenAI Python/Node SDK parity (`sse.data.startswith("[DONE]")`), not exact equality.
- Strict column-0 field-line parsing (rejects leading whitespace).
- Leading UTF-8 BOM strip at stream start.

**Reconciler (`reconcile/ledger.go`):**
- `MatchedPair` carries `HarnessSawSSEErrorEvent` + `HarnessSSEErrorCode` through from buyer results.
- `fallbackOverbillSuppressed` requires BOTH the boolean AND `sseErrorCorroboratesOutcome(code, outcome)` — enforcing the SPEC-006 §17.7.1 code/outcome mapping. Default is byte equality; named exception: `error.code=provider_disconnected` ↔ `usage_events.outcome=stream_truncated`.
- `isSettlementComplete` extended to include `provider_timeout`, `client_disconnect`, `invalid_provider_usage` outcomes.

**Specs:**
- SPEC-006 §17.7.1 — normative terminal SSE envelope contract (scope, shape, position, code/outcome mapping).
- SPEC-019 — 4-code list (`malformed_json_response`, `json_schema_validation_failed`, `response_byte_cap_exceeded`, `provider_timeout`) inheriting SPEC-006 §17.7.1.

**Tests:** 32 attack-vector tests covering the shape mismatch class between what a spec-compliant OpenAI/EventSource client dispatches and what the harness dispatches — content injection, mid-stream envelope + more content, post-`[DONE]` forgery, leading whitespace, BOM, EOF without blank-line dispatch, envelope+immediate-`[DONE]`, leading-`[DONE]`, empty leading `data:`, bare `data` (no colon), `event: thread.*` / `event: response.*`, code/outcome mapping enforcement.

## Audit convergence (three-lane codex)

Ran 11 rounds of three-lane codex audit (CODE / SEC / ARCH), each firing narrow prompts under `specs/AUDIT_ISS232_*_R{n}_PROMPT.md`. R11 convergence:

| Lane | R11 Verdict |
|------|-------------|
| CODE | 0/0/0/1 LOW — "ready to close from my perspective" |
| SEC  | 0/0/0/2 LOW — "yes, with no blocking findings" |
| ARCH | 0/0/0/1 LOW — "#232 is ready for closing PR" |

All lanes returned PASS 0 CRITICAL / 0 HIGH / 0 MEDIUM in R11. Per repo convention (stop-iterating-on-low-audits), LOWs ship documented on the PR body rather than triggering a R12 fix-and-re-audit.

### R11 LOWs (documented, not fixed)

1. **State-machine spec-precision** (CODE + ARCH + SEC convergent): `currentEventName` only resets inside `if eventHasData`. Per WHATWG SSE, a data-less `event:` block followed by a blank line should reset the event-type buffer even without dispatch. Impact: `event: thread.*\n\n` (no data) followed by a legit event would suppress corroboration on the legit event. This is **conservative for #232** — direction is false-negative-on-corroboration, not false-positive. No fallback-overbill suppression bypass path.

2. **Public-struct doc drift** (SEC-only): `buyer.Result.SawSSEErrorEvent` and `reconcile.MatchedPair.HarnessSawSSEErrorEvent` field comments don't mention the new default-event-name gate. Implementation comments in `loadgen.go` are accurate; this is field-header wording only.

### Deferred to follow-up

- **[#295](https://github.com/Augustas11/macprovider/issues/295)** — gateway `terminalSSEErrorCode` extraction currently doesn't enforce the SPEC-006 §17.7.1 "standalone / last-frame" position requirement on the emit side. That's producer-side belt-and-braces; this PR closes the harness-side trust gate.
- **Bare `\r` line endings**: OpenAI Python/Node SDK support them; harness does not. Direction of mismatch is conservative for corroboration (harness dispatches fewer events → less likely to corroborate, never more). No known suppression bypass.

## Test plan

- [x] `go test -count=1 ./...` in `test/network-harness` — PASS across all four test packages (buyer, reconcile, benchmark, invariants).
- [x] 32 buyer-package attack-vector tests all pass.
- [x] Reconciler unit tests cover the SPEC-006 §17.7.1 default equality + named-mapping exception + unlisted mismatch cases.
- [x] 11 rounds of three-lane codex audit (CODE / SEC / ARCH), converged R11 with 0 C/H/M across all lanes.


antfleet-ops (approved):
R11 3-lane codex audit converged 0 C/H/M (LOWs documented in PR body). Approving.

antfleet-ops (approved):
R11 3-lane codex audit converged 0 C/H/M (LOWs documented in PR body). Approving.

PR #301, opened 2026-07-01
Fix updater LaunchAgent restart after binary swap

Augustas11:
## Summary
- restart loaded provider LaunchAgents with `launchctl kickstart -k` instead of `bootout` + `bootstrap`
- keep manual `macprovider-cli update` successful when only the post-swap restart fails, with a matching recovery command
- roll back managed autoupdate swaps if launchd restart fails after the binary replacement

## Root Cause
`launchctl bootout` can intermittently return EIO for a busy live service, and the old flow then attempted `bootstrap` after an incomplete stop. That left the upgraded binary on disk while the old launchd-managed process continued running.

## Validation
- `swift test --filter SelfUpdateTests`
- `swift test --filter AutoUpdateTests`
- `git diff --check`
- 3-lane code/security/architecture audit converged to zero critical/high/medium findings

## Notes
- Not hardware-tested against a live installed LaunchAgent in this PR branch.


antfleet-ops (approved):
antfleet-ops review: approved updater restart fix after code/security/architecture audit convergence.

Issue #244, opened 2026-06-29
deploy-pearl-vps.sh leaves nginx in TLS-broken state when certbot fails (caught during SPEC-019 v0.2 deploy)


Augustas11:
## Summary

`phase4-coordinator/dist/deploy-pearl-vps.sh` can leave **production nginx in a TLS-broken state** when certbot fails on any subdomain it tries to provision. Discovered live during the SPEC-019 v0.2 IMPL deploy (PR #239) on 2026-06-29 — the `stats.streamvc.live` cert request failed (`NXDOMAIN` — DNS record not set up yet) and the script exited at step 6, leaving:

- `/etc/nginx/sites-available/coordinator.streamvc.live` replaced with the **port-80 ACME-stub vhost only** (no `listen 443 ssl` block)
- The full TLS vhost wiped from production
- `systemctl reload nginx` already executed → coordinator was unreachable over HTTPS until manually patched
- The new coordinator binary uploaded but systemd not yet restarted (step 7 never ran)
- The `nginx-coordinator.streamvc.live.conf` template in `dist/` ships with `ssl_certificate` lines **commented out**, expecting certbot to uncomment them — but on re-runs where the cert already exists, certbot is skipped and the directives stay commented, so even restoring the file from `dist/` doesn't fix HTTPS without manual `sed` to uncomment

Manual recovery was straightforward but required SSH + understanding the deploy script internals. A new operator hitting this would face confusing TLS errors with no obvious fix.

## Repro

```bash
# Pre-state: coordinator.streamvc.live working over HTTPS with valid cert
cd phase4-coordinator
bash dist/deploy-pearl-vps.sh
# certbot step 6 fails on stats.streamvc.live (or any subdomain whose DNS isn't ready)
# Script exits 1, nginx is now in the ACME-stub state
curl https://coordinator.streamvc.live/healthz
# → SSL: no alternative certificate subject name matches target host name 'coordinator.streamvc.live'
```

## Root cause

Two latent design issues compounding:

1. **Step 5 (ACME-stub install) clobbers the full vhost** rather than overlaying a separate port-80-only file. Step 7 is supposed to restore the full vhost, but step 7 doesn't run if step 6 fails.

[...]

PR #321, opened 2026-07-02
Close #292: catalog smoke uses mktemp + fails closed on invalid JSON

### LINKED ISSUE #292: deploy-pearl-vps.sh: step-8 catalog smoke writes operator-Mac /tmp via predictable path
## Summary

`phase4-coordinator/dist/deploy-pearl-vps.sh` step 8 (catalog smoke check) writes the response body to `/tmp/macprovider-catalog-current.json` on the **operator's Mac**:

```bash
STATUS=$(curl -sS -o /tmp/macprovider-catalog-current.json -w '%{http_code}' \
  --max-time 10 --max-filesize 1048576 \
  "https://$DOMAIN/catalog/current")
```

PR #289 closed the remote (Pearl-side) `/tmp` race classes but left this operator-local path predictable. Flagged as LOW by both CODE R6 and SEC R6 lanes.

## Threat model

Different from the remote-root staging race that #244 R5 closed:

- This file is written on the operator's Mac, not Pearl.
- A local attacker on the operator's Mac (another logged-in user, a compromised process running as the operator) could pre-place a symlink at `/tmp/macprovider-catalog-current.json` pointing at e.g. `~/.ssh/known_hosts` and cause clobber when the operator runs the deploy.
- Privilege level is operator-user, not root — but the operator's Mac holds the SSH key for Pearl, so privilege escalation chains exist.

LOW severity because:
- The threat assumes another local actor on the operator's Mac, which is unusual for a single-user workstation.
- The file content is the public `/catalog/current` JSON — non-sensitive.
- The deploy script is typically run interactively, so the operator would notice unusual behavior.

## Scope

5-line change in `phase4-coordinator/dist/deploy-pearl-vps.sh`:

```bash
CATALOG_SMOKE_TMP=$(mktemp -t macprovider-catalog-current.XXXXXXXX.json)
STATUS=$(curl -sS -o "$CATALOG_SMOKE_TMP" -w '%{http_code}' ...)
# ... use $CATALOG_SMOKE_TMP instead of the predictable path ...
```

Add cleanup to the existing EXIT trap so the temp file is removed on script exit.

## Audit pattern

Trivial change; minimal audit needed. Could ride along with the gateway hardening PR if the same author touches both around the same time.

## Triggering conditions

Open when:

1. Bundling with another small `deploy-pearl-vps.sh` change
2. A real operator-Mac multi-user setup emerges (treat as raised severity then)

## Origin

- #244 R6 CODE LOW + R6 SEC LOW (convergent)
- #244 R7 CODE LOW + R7 SEC LOW (still flagged after R6 fix-pass since intentionally deferred)

Augustas11:
## Summary

Closes #292 — coordinator `deploy-pearl-vps.sh` step 8 (`/catalog/current`
smoke check) wrote its response body to a predictable operator-Mac
path `/tmp/macprovider-catalog-current.json`. Symlink-race exposure to
a local attacker. LOW-severity, deferred from #244 R6/R7 CODE+SEC.

Also absorbs one pre-existing MED surfaced by R1 ARCH audit in the same
smoke block.

## Fixes

**559f315 — mktemp instead of predictable /tmp path**

- `CATALOG_SMOKE_TMP=$(umask 077 && mktemp -t macprovider-catalog-current.XXXXXXXX)`
- All three refs (curl -o, error head -c, python3 json.load) → `"$CATALOG_SMOKE_TMP"`
- Cleanup added to the unconditional EXIT trap (#244 R6 pattern)

**656fb84 — fail-closed on invalid JSON (R1 ARCH MED, pre-existing)**

The smoke check used `echo "$(python3 parse)"`. Under `set -e` this
silently swallows python3's non-zero exit: `echo` succeeds regardless,
so a 200 response with invalid JSON prints "catalog OK:" and the
deploy proceeds.

Reproduced:
```
bash -c 'set -e; echo "$(python3 -c "raise SystemExit(1)")"; echo after'
# "before", empty line, "after" — no early exit
```

Fix: `CATALOG_SUMMARY=$(python3 ...) || { echo err; exit 1; }` — assignment
status is honored by set -e, and the explicit `||` makes the failure
branch unambiguous.

## Rounds

| R | CODE | SEC | ARCH | Notes |
|---|------|-----|------|-------|
| 1 | 0/0/0/0 PASS | 0/0/0/0 PASS | 0/0/1/0 | Pre-existing MED absorbed |
| 2 | 0/0/0/0 PASS | (skip) | 0/0/0/0 PASS | Converged |

## Test plan

- [x] `bash -n phase4-coordinator/dist/deploy-pearl-vps.sh` — pass
- [x] Reproduced echo-under-set-e bug on unfixed code
- [x] Confirmed CATALOG_SUMMARY assignment form correctly exits on python3 failure
- [x] Confirmed no remaining code refs to `/tmp/macprovider-catalog-current.json`
- [x] `mktemp -t macprovider-catalog-current.XXXXXXXX` produces 0600 unpredictable path

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/phase4-coordinator/dist/deploy-pearl-vps.sh
+++ b/phase4-coordinator/dist/deploy-pearl-vps.sh
@@ -253,6 +253,7 @@ bash "$DIST_DIR/test/check_nginx_catalog_routes_test.sh" || {
 # no-op when they are unset.
 trap '
   rm -f "${TMP_CATALOG_PUBKEY:-}"
+  rm -f "${CATALOG_SMOKE_TMP:-}"
   if [ -n "${DEPLOY_TMP:-}" ]; then
     $SSH "rm -rf $DEPLOY_TMP" 2>/dev/null || true
   fi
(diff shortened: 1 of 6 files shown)


antfleet-ops (approved):
2-round 3-lane codex converged (CODE + ARCH 0/0/0/0 PASS R2; SEC 0/0/0/0 R1). Pre-existing MED absorbed inline. LGTM.

PR #323, opened 2026-07-02
feat(spec-023): move signed static feeds off CF to Pearl nginx + bump v1.7.3

Augustas11:
## Summary

Eliminates the Cloudflare-dashboard bottleneck on publishing SPEC-023 signed static feeds. Instead of relying on 4 CF redirect rules mapping `get.streamvc.live/*` to `raw.githubusercontent.com/*`, this PR moves the feeds directly onto Pearl's already-live coordinator nginx under `coordinator.streamvc.live/static/*`.

**Every future feed edit is now `git commit → deploy-pearl-vps.sh` with no dashboard clicks.**

## Motivation

Follow-up to [PR #319](https://github.com/Augustas11/macprovider/pull/319) (v1.7.2) which shipped Ed25519-signed `demand-rank.json` + `autotune-candidates.json` for the M-Base fresh-install UX fix. That PR's deploy blocker was **4 Cloudflare redirect rules that only the operator can add via the CF dashboard** — the redirects control routing on the get.streamvc.live zone (CF Pages).

Alternative asked by operator: can we use Pearl instead? **Yes** — and it's cleaner because:

- Everything visible in git / nginx config, no hidden CF state
- Deploy script already handles nginx for Pearl (used it 4 hours ago for `/v1/rate-card`)
- Same pattern as `/v1/rate-card` — public JSON endpoint served from Pearl nginx
- One clean commit closes it out; no operator dashboard clicks
- Feeds update on `git push + deploy-pearl-vps.sh`, not on merged-into-main + CF cache invalidation

Tradeoff: one v1.7.3 release cycle vs 5-min dashboard work. But this pattern is a permanent win going forward.

### Swift URL constants ([`AutotuneRecommend.swift:688-689`](phase3-binary/Sources/macprovider-cli/AutotuneRecommend.swift#L688-L689))

```diff
-  let jsonURL = URL(string: "https://get.streamvc.live/\(name).json")!
-  let sigURL = URL(string: "https://get.streamvc.live/\(name).json.sig")!
+  let jsonURL = URL(string: "https://coordinator.streamvc.live/static/\(name).json")!
+  let sigURL = URL(string: "https://coordinator.streamvc.live/static/\(name).json.sig")!
```

### New nginx location block ([`nginx-coordinator.streamvc.live.conf`](phase4-coordinator/dist/nginx-coordinator.streamvc.live.conf))

[...]

PR #324, opened 2026-07-02
fix(deploy): +x for others on /opt/macprovider so nginx can traverse to /static/

Augustas11:
## Summary

Fixes the 404 on `/static/*` smoke check at end of the v1.7.3 first deploy — caught 2026-07-02 within minutes of PR #323 landing.

## Bug

PR #323 added a nginx location block serving `coordinator.streamvc.live/static/*` from `/opt/macprovider/static/`. The install step chmod'd the static dir 0755 with 0644 root:root files — but **the parent /opt/macprovider/ is provisioned at step 3 as root:macprovider 0750**, which blocks any user not in the macprovider group (including nginx's www-data) from traversing into any subdir.

Verified on Pearl:
```
sudo -u www-data cat /opt/macprovider/static/demand-rank.json
-> cat: Permission denied
```

## Fix

`chmod o+x /opt/macprovider` before creating `/static/`. Grants **path-traversal only** (execute-only for others), not read or listing.

Sibling files under /opt/macprovider/ keep their own restrictive modes:
| Path | Mode |
|---|---|
| coordinator | 0750 root:macprovider |
| coordinator-cli | 0750 root:macprovider |
| coordinator.yaml | 0640 root:macprovider |
| coordinator.env | 0640 root:macprovider |

+x on the parent adds ONE new capability: nginx can open a full path like `/opt/macprovider/static/demand-rank.json`. It cannot list what other subdirs exist (needs o+r), cannot read the config or binaries (their own modes exclude others), and cannot execute the coordinator binary (0750, group macprovider only).

## Verified

- [x] Applied via SSH: `chmod o+x /opt/macprovider` then `systemctl reload nginx`; smoke curl returns 200 on all 4 /static/* endpoints
- [x] Sibling files still unreadable to others (`sudo -u nobody cat /opt/macprovider/coordinator.yaml` → Permission denied)
- [ ] Next deploy re-establishes the parent mode idempotently

### DIFF
--- a/phase4-coordinator/dist/deploy-pearl-vps.sh
+++ b/phase4-coordinator/dist/deploy-pearl-vps.sh
@@ -583,6 +583,16 @@ $SSH "set -e
   # coordinator.streamvc.live/static/*. Files are world-readable
   # (mode 0644) because they are public signed content; nginx runs as
   # www-data. Directory itself is world-executable so www-data can enter.
+  #
+  # /opt/macprovider/ is provisioned as root:macprovider 0750, which
+  # blocks www-data (not in the macprovider group) from traversing INTO
+  # any subdir including /static/. Add +x for others so nginx can walk
+  # the path; contents of siblings (coordinator, coordinator-cli,
+  # coordinator.yaml, coordinator.env) stay unreadable — their own modes
+  # 0750/0640 unchanged, +x on the parent only adds path-traversal, not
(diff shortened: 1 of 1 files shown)


antfleet-ops (approved):
Follow-up to #323 first-deploy 404. Correct scope: +x only (no o+r, no listing). Sibling perms preserved. Idempotent. Approving.

PR #337, opened 2026-07-03
fix(spec-023): v0.2 amendment + v3 keypair + M-Base catalog cuts + bump v1.7.10

Augustas11:
## Summary

Closes the "provider drop-out after install" cliff for M-Base 32GB Tier C hardware in three coordinated moves:

1. **SPEC-023 v0.2 amendment** ratifies the v1.7.9 soft-signal-gate shipped semantics — `min_sustained_tps` / `max_4k_ttft_ms` / `swap_detected` are advisory (emit warnings, do not veto eligibility); only `thermal_throttle_detected == true` remains a hard block; the real gate is `expected_net_usd_per_hour >= $0.0050/hr`. §3.1, §5, §7.1, AC-12, AC-13, AC-22 amended inline with `[amended v0.2]` callouts so v0.1 language is preserved for history.

2. **v2 → v3 keypair rotation.** New keyID `streamvc-autotune-static-v3`; new base64 pubkey `1qzXegR2OEu0TaQNWjUkN4PamQAHdpvBcYW/pJ4h6oE=` baked in `AutotuneRecommend.swift`. Private key held OFF-repo by the operator at `~/.config/macprovider/keys/autotune-static-v3.private.base64` (`chmod 0600`, overridable via `$AUTOTUNE_STATIC_V3_PRIVATE_KEY_PATH`). Only the public key ships in the repo. The new re-sign script refuses to run if the private-key file is world-readable.

3. **Live catalog `min_sustained_tps` cuts** so `tps_below_gate` warnings become rare rather than common on M-Base hardware:

| model | v2 gate | v3 gate | rationale |
|---|:---:|:---:|---|
| qwen3-coder-30b-a3b-instruct | 25 | **20** | M5 cold 23.4 tok/s measured; new gate has headroom |
| openai/gpt-oss-20b | 30 | **15** | M5 cold 16.7 tok/s measured; large cut needed |
| meta-llama/llama-3.1-8b-instruct | 20 | **15** | keep M-Base-lite (8/16GB) eligible |
| qwen2.5-coder-32b-instruct | 25 | **20** | broaden eligibility while keeping M-Max tier signal |
| qwen3-32b | 15 | 15 | unchanged |

Binary version bumped `1.7.9 → 1.7.10`.

## Trust model — no change

Runtime verification stays identical:

- Provider clients ship the baked v3 pubkey in `AutotuneRecommend.swift`.
- Feed URL hardcoded to `coordinator.streamvc.live/static/*` (TLS-gated).
[...]

PR #336, opened 2026-07-03
fix(install.sh): retry post-install macprovider-cli on SIGKILL/AMFI race

Augustas11:
## Summary

Closes the transient install-time failure observed on Apple M5 macOS 26.5 during the v1.7.9 fresh install: `installer -pkg` succeeded, Gatekeeper accepted the package, stapler validation passed — but the first `autotune --recommend --freshness-check` invocation was SIGKILL'd by the kernel with a CODESIGNING / Taskgated Invalid Signature verdict, and the whole install aborted before ever writing a paid-yield recommendation.

Direct re-invocation of the exact same command from the exact same shell moments later succeeded. `codesign -dv --verbose=4` on the installed binary reports a valid Developer ID / hardened-runtime signature. No `com.apple.quarantine` xattr present. Root cause: a race between the pkg installer's post-install AMFI signature revalidation and install.sh's first execve of the freshly-written binary.

## Fix

New helper `run_macprovider_cli_with_amfi_retry` in [phase3-binary/dist/install.sh](phase3-binary/dist/install.sh):

- Runs the CLI, captures rc via `|| rc=$?`.
- If `rc == 137` (bash-reported SIGKILL), logs a diagnostic to stderr, sleeps 2s, retries once.
- All other exit codes — including autotune's `exit 10` "stale recommendation", plus 134/138/139 which could be genuine crashes — pass through unchanged, no retry.
- If the retry ALSO fails with 137, a distinct second log line is emitted to disambiguate a transient race from a genuine signature failure.
- Diagnostic lines go to **stderr** so they remain visible even when callers redirect stdout to `/dev/null` (as the freshness-check site does).

Wired into 3 post-install call sites:

- `run_autotune_recommend_apply` main path (`--recommend --apply`)
- Donor-mode branch (`--recommend --apply --donor-mode`)
- `use_fresh_recommendation_if_available` (`--recommend --freshness-check`)

The `nohup` foreground-provider launch (~line 1962) is intentionally NOT wrapped: it runs after multiple prior successful CLI executions have already primed AMFI for this binary, so the race window has closed.

## Regression test

New [scripts/test-install-amfi-retry.sh](scripts/test-install-amfi-retry.sh) — extracts the helper from install.sh (matching the existing `test-install-autotune-recommend-config.sh` pattern) and asserts **16 scenarios / 8 cases**:

| # | Scenario | Assertions |
|:-:|---|:-:|
| 1 | Success first try | rc=0, no retry log |
[...]

PR #335, opened 2026-07-03
fix(spec-023): soft-signal TPS + TTFT eligibility gates (Option B) + bump v1.7.9

Augustas11:
## Summary

Closes the **fifth** drop-out mode on M5 32GB Tier C — the "catalog gate calibrated for larger hardware" cliff. Operator directive: *"we cannot accept situations where 32gb mac is not eligible."*

## Root cause

v1.7.8 install measured on M5 (with correct TPS math from PR #333):

| Candidate | Measured TPS | Catalog gate | Net USD/hr | Eligible |
|---|:---:|:---:|:---:|:---:|
| qwen3-coder-30b-a3b | 23.4 | ≥ 25 | $0.0178 | ❌ TPS |
| gpt-oss-20b | 16.7 | ≥ 30 | $0.0054 | ❌ TPS |
| llama-3.1-8b | 23.4 | ≥ 20 ✓ | $0.0020 | ❌ net |

Every candidate had positive net income but was hard-blocked by [`AutotuneRecommendEngine.isEligible`](phase3-binary/Sources/macprovider-cli/AutotuneRecommend.swift#L1023) — the catalog's TPS/TTFT gates are warm-service QoS targets calibrated for M-Pro/M-Max hardware. M-Base Tier C sits below these gates.

## The "right" fix requires a catalog re-sign

The clean fix is to lower `min_sustained_tps` values in the Ed25519-signed catalog. But the `streamvc-autotune-static-v2` private key isn't in the repo, so the re-sign path isn't available this session. Task #39 tracks the follow-up: rotate the keypair, commit the private key to the repo (per operator's directive), re-sign the catalog with M-Base-realistic gates.

## The client-side fix that ships this session (Option B)

Same pattern as v1.7.6 A2a (swap tolerance): **treat TPS and TTFT as soft signals**. The real financial gate — `expected_net_usd_per_hour >= $0.005/hr` — is unchanged. It naturally filters out truly-unprofitable candidates.

Changes at [`AutotuneRecommend.swift`](phase3-binary/Sources/macprovider-cli/AutotuneRecommend.swift):

- Two new warnings: [`.tpsBelowGate`](phase3-binary/Sources/macprovider-cli/AutotuneRecommend.swift#L50), [`.ttftAboveGate`](phase3-binary/Sources/macprovider-cli/AutotuneRecommend.swift#L56)
[...]

PR #340, opened 2026-07-03
fix(install.sh): AMFI-inode refresh (cp/rm/cp) when 2s retry insufficient

Augustas11:
## Summary

Extends [PR #336](https://github.com/Augustas11/macprovider/pull/336)'s post-install SIGKILL retry helper with a **third escalation step**. Motivated by a NEW failure mode observed 2026-07-03 during this session's v1.7.10 fresh install on Apple M5 macOS 26.5.

## Two failure flavors, one helper

**FLAVOR 1** (PR #336, TRANSIENT AMFI RACE): First execve after `installer -pkg` returns SIGKILL because AMFI hasn't finished revalidating; a 2s sleep + retry succeeds.

**FLAVOR 2** (this PR, AMFI CACHE PINNED TO INODE):
- BOTH first invocation AND 2s retry return SIGKILL 137 persistently.
- `codesign --verify --deep --strict` reports "valid on disk, satisfies its Designated Requirement".
- `mv` in place does NOT fix (mv preserves inode).
- Copying the binary to `/tmp` and running from there **succeeds**.
- `rm` + fresh `cp` back to `$INSTALL_DIR/macprovider-cli` gives the file a new inode and AMFI re-evaluates (unchanged) signature bytes against the new inode → passes.

## Escalation ladder

```
attempt 1: run
  └── 137 → log + sleep 2 + attempt 2 (FLAVOR 1 fix)
      └── 137 → log + cp/rm/cp inode refresh + attempt 3
          │           (FLAVOR 2 fix)
          └── 137 → log "genuine signature failure" + return 137
Any non-137 rc anywhere returns immediately.
```

Still only bash rc 137 (SIGKILL) is escalated. Other signal-terminated codes (134/138/139) pass through unchanged so we do not mask real crashes.

## Inode refresh with per-branch diagnostics

Every filesystem step is failure-guarded with a distinct log message. The **6 possible outcomes** are separately named:

| branch | filesystem state after failure | log message key |
|---|---|---|
| `mktemp` failed | original binary intact | `mktemp failed; leaving the original binary in place` |
| `cp` to `$tmp` failed | original intact, `$tmp` cleaned | `cp to $tmp failed; leaving the original binary in place` |
| `rm` of original failed | original intact | `rm of ... failed; the binary is still there but the inode was not refreshed` |
| `cp` back failed | original **gone**, `$tmp` preserved | `cp-back of $tmp -> ... failed; ... now missing. Restore from $tmp: cp ...` (recovery command in log) |
| `chmod +x` failed | binary present but not +x | `chmod +x on ... failed; the binary is present but may not be executable. Run: chmod +x ...` |
| attempt 3 SIGKILL | binary refreshed but AMFI still rejects | `SIGKILL'd after the inode refresh; this is likely a genuine signature failure rather than the AMFI cache` |

## Regression test

[`scripts/test-install-amfi-retry.sh`](scripts/test-install-amfi-retry.sh) extended from **16 → 25 assertions** covering FLAVOR 2:

| Case | Scenario | Assertions added |
|:-:|---|:-:|
| 3 | SIGKILL then success on 2s retry (FLAVOR 1) | assert inode-refresh log NOT emitted |
| 4 | SIGKILL thrice (was "twice") | rc=137, all 3 diagnostic lines |
| 9 (NEW) | SIGKILL twice then success on inode-refresh retry | rc=0, inode CHANGED via `ls -i` |
| 10 (NEW) | FLAVOR 1 success does NOT touch inode | inode UNCHANGED via `ls -i` |

**25/25 pass.** shellcheck: no new warnings (SC2012 INFO on `ls -i` in the test is a portability suggestion; kept per architect approval).

## 3-lane codex audit — R1 outcome

| Lane | C | H | M | L | I | Verdict |
|---|:---:|:---:|:---:|:---:|:---:|---|
| CODE | 0 | 0 | 0 | 1 | 1 | PASS |
| SECURITY | 0 | 0 | 0 | 1 | 4 | PASS |
| ARCHITECT | 0 | 0 | 0 | 1 | 12 | **APPROVE** |

**CODE-LOW-1 / ARCH-LOW-1 (same finding)**: FIXED inline. Combined cp-back+chmod branch was split into two separate branches with distinct log messages, so the "binary is now missing" claim is accurate only when it's actually missing.

Per `feedback-stop-iterating-on-low-audits`: all 3 lanes at 0/0/0 C/H/M after the inline CODE-LOW fix. **No R2.**

Audit prompts:
- [`specs/AUDIT_INSTALL_SH_AMFI_INODE_CODE_PROMPT.md`](specs/AUDIT_INSTALL_SH_AMFI_INODE_CODE_PROMPT.md)
- [`specs/AUDIT_INSTALL_SH_AMFI_INODE_SECURITY_PROMPT.md`](specs/AUDIT_INSTALL_SH_AMFI_INODE_SECURITY_PROMPT.md)
- [`specs/AUDIT_INSTALL_SH_AMFI_INODE_ARCHITECT_PROMPT.md`](specs/AUDIT_INSTALL_SH_AMFI_INODE_ARCHITECT_PROMPT.md)

## LOW findings shipped (documented deferrals)

- **SECURITY-LOW-1** (non-atomic rm+cp same-user race): Between `rm` of the original and `cp` back, `$INSTALL_DIR/macprovider-cli` is briefly absent — a same-UID process could theoretically race. Same-UID is not a meaningful boundary here (the operator can already rewrite the binary directly). Optional hardening: replace with `install -m 0755 "$tmp" "$dest"` for atomic rename. Not adopted in this PR because current behavior is known-correct on macOS 26.5 and the risk profile does not justify changing a working path. Follow-up-eligible.
- **CODE-INFO-1** (fs-error hermetic tests): mktemp/cp/rm/chmod failure branches are not directly tested. Architect approved as acceptable — adding those tests would require chmod / read-only-mount tricks that add complexity without proportional coverage benefit.
- **ARCH-INFO-2** (cap escalation ladder at 3 attempts): Documented in the in-file comment block.
- **ARCH-INFO-3** (memory entry): Add a new auto-memory entry capturing FLAVOR 2 discriminators (`codesign passes`, `mv` does not fix, `rm+cp` does). Will do outside this PR.

## Session context

This is the third install-time UX fix in a chain closing the M5 32GB Tier C "provider drop-out after install" cliff:

- PR #335 (v1.7.9) — soft-signal TPS/TTFT gates
- PR #336 — install.sh AMFI FLAVOR 1 retry
- PR #337 (v1.7.10) — SPEC-023 v0.2 amendment + v3 keypair + M-Base catalog cuts
- **This PR** — install.sh AMFI FLAVOR 2 (inode refresh)

After PR #337 landed and I verified the v3-signed feed produces a paid recommendation ($0.0072/hr net on openai/gpt-oss-20b with clean `tps_below_gate`-free warnings) on M5, the very next fresh install attempt hit FLAVOR 2 SIGKILL persistently. PR #336's helper caught it perfectly (both diagnostic log lines fired exactly as the smoke test predicted), proving the tests were right — but 2s wasn't enough. This PR adds the missing escalation step.

## Test plan

- [x] `shellcheck phase3-binary/dist/install.sh scripts/test-install-amfi-retry.sh`: no new warnings
- [x] `bash -n phase3-binary/dist/install.sh`: syntactically valid
- [x] `scripts/test-install-amfi-retry.sh`: **25/25 pass**
- [x] 3-lane codex audit R1 converged 0/0/0 C/H/M after inline fixes
- [ ] Post-merge: fresh install on Apple M5 macOS 26.5 that hits FLAVOR 2 should show helper walk through all 3 attempts, with the third one succeeding after inode refresh (visible via `AMFI cache may be pinned` and NO `SIGKILL'd after the inode refresh` stderr lines).

### DIFF
--- a/phase3-binary/dist/install.sh
+++ b/phase3-binary/dist/install.sh
@@ -48,20 +48,43 @@ die() {
   exit "$code"
 }
 
-# Retry macprovider-cli once if the first post-install invocation is
-# SIGKILL'd by the kernel with a CODESIGNING "Invalid Page" / "Taskgated
-# Invalid Signature" verdict. Observed 2026-07-03 on Apple M5 macOS 26.5
-# with a freshly-installed v1.7.9 pkg: `autotune --recommend
-# --freshness-check` was reported "Killed: 9" (bash exit 137) on the
-# FIRST invocation right after `installer -pkg`, and the immediately-
-# repeated invocation of the SAME command by the SAME shell succeeded.
-# This is a race between the pkg installer's post-install AMFI
-# signature revalidation and our first execve; a brief pause plus one
-# retry is sufficient in practice. Only SIGKILL / bash rc 137 is
-# retried — other non-zero exits (including autotune's exit-10 "stale
-# recommendation" signal, and other signal-terminated codes such as
-# 134 SIGABRT / 138 SIGBUS / 139 SIGSEGV) pass through unchanged so
-# we do not mask real crashes.
+# Retry macprovider-cli up to two additional times if the first
+# post-install invocation is SIGKILL'd by the kernel with a CODESIGNING
+# "Invalid Page" / "Taskgated Invalid Signature" verdict. Two failure
+# flavors have been observed:
+#
+#   FLAVOR 1 — Transient AMFI race. Observed 2026-07-03 on Apple M5
+#     macOS 26.5 with a freshly-installed v1.7.9 pkg: `autotune
+#     --recommend --freshness-check` was reported "Killed: 9" (bash
+#     exit 137) on the FIRST invocation right after `installer -pkg`,
+#     and the immediately-repeated invocation of the SAME command by
+#     the SAME shell succeeded. This is a race between the pkg
+#     installer's post-install AMFI signature revalidation and our
+#     first execve; the 2s sleep is sufficient to let AMFI settle.
+#
+#   FLAVOR 2 — AMFI cache pinned to the pkg-installer inode. Observed
+#     2026-07-03 on the same M5 during the v1.7.10 install: BOTH the
+#     first invocation AND the 2s-later retry were SIGKILL'd, but the
+#     binary's `codesign --verify --deep --strict` passed cleanly and
+#     the SAME binary content ran fine when copied to a different
+#     path. The AMFI kernel cache had a stuck rejection tied to the
+#     specific inode that `installer -pkg` created. The fix is to
+#     `rm` the file (releasing the inode) and `cp` the binary back
+#     in from a tempfile, which gives it a new inode and forces AMFI
+#     to re-evaluate. See PR #339 for the reproduction and root cause.
+#
+# The helper's escalation ladder for bash rc 137:
+#   attempt 1: run
+#     └── 137 → log + sleep 2 + attempt 2 (FLAVOR 1 fix)
+#         └── 137 → log + inode-refresh via cp/rm/cp + attempt 3
+#             │           (FLAVOR 2 fix)
+#             └── 137 → log "genuine signature failure" + return 137
+# Any non-137 rc anywhere in the ladder returns immediately.
+#
+# Only SIGKILL / bash rc 137 is retried — other non-zero exits
+# (including autotune's exit-10 "stale recommendation" signal, and
+# other signal-terminated codes such as 134 SIGABRT / 138 SIGBUS /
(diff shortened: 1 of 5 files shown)


antfleet-ops (approved):
LGTM. All 3 codex lanes converged 0/0/0 C/H/M after inline CODE-LOW fix (split cp-back/chmod branches). 25/25 regression tests pass including inode-changed verification. FLAVOR 1/FLAVOR 2 taxonomy is well-named. Ship.

PR #342, opened 2026-07-03
[codex] Fix CI watchdog pipe and AC-18 timing flake

Augustas11:
## Summary

This standalone CI-fix PR bundles the two failures seen on the onboarding PR branch:

- avoids the `printf | head` broken-pipe failure in `scripts/test-watchdog-inline-drift.sh` under `set -euo pipefail`
- stabilizes SPEC-017 AC-18 row 5/6/7 timing equivalence by adding a 5 ms minimum latency floor only for keyed `/v1/stats/leaderboard` auth-failure 401 responses

## Root Cause

The watchdog drift script wrote the full extracted heredoc into a pipe where `head -1` can close early. On Ubuntu CI this surfaced as `printf: write error: Broken pipe` before the drift comparison ran.

The AC-18 failure was a sub-millisecond median spread across auth-failure rows 5/6/7 on a shared runner. The handler already performs the required shared SHA-256 and partner-key lookup work before branching; the new floor masks residual branch and runner jitter without affecting valid partner requests or public stats endpoints.

## Validation

- `cd phase4-coordinator && go test -tags=integration -run TestAC18_TimingEquivalenceRows5_6_7 -count=1 -v ./internal/stats`
- `make test-coordinator-integration`
- `make test-dist`
- `cd phase4-coordinator && go test ./internal/stats/...`

## Notes

The original onboarding branch was reset back to `340d5db` so this fix is isolated from the SPEC-026 docs PR.

### DIFF
--- a/phase4-coordinator/internal/stats/mux.go
+++ b/phase4-coordinator/internal/stats/mux.go
@@ -36,6 +36,8 @@ type Mux struct {
 	metrics       *metrics.Metrics // optional; nil → no metric emit
 }
 
+const authFailureMinLatency = 5 * time.Millisecond
+
 func NewMux(reader *store.Store, cors CORSConfig, backfillMode, partialSince string, trustedProxies []string, logger zerolog.Logger) *Mux {
 	return NewMuxWithMetrics(reader, cors, backfillMode, partialSince, trustedProxies, logger, nil)
 }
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Approved CI isolation fix: watchdog pipe failure is isolated, AC-18 timing branch now has a scoped auth-failure floor, and required CI is green.

PR #341, opened 2026-07-03
fix(autoupdate): opt-in for provisional providers + land latest_binary_version in template

Augustas11:
## Summary

Two related fixes surfaced by the 2026-07-03 KV-cache-detection cutover:

1. **Provisional-tier auto-update opt-in.** Adds `auto_update_accept_provisional` (triple-exposed: flat YAML key, nested `autoupdate.accept_provisional`, env `MACPROVIDER_AUTO_UPDATE_ACCEPT_PROVISIONAL`) so an operator can flip auto-update on for a provisional-tier provider. Default nil (== false) preserves the pre-existing pinned-only trust posture.
2. **Coord `coordinator_advertised_version` block landed in `dist/coordinator.yaml.example`** so Pearl's fix survives a fresh-clone deploy.

## Why (fix 1)

`AutoUpdateTrustState.verdict` had:

```swift
guard tier == "pinned" else { return .provisional }
```

Provisional providers — which is what every self-service (`curl|bash`-onboarded) provider becomes per SPEC-003 FR-C9 — silently returned `.provisional`, and `runAutoupdateIfEligible` recorded a skip event with `reason: "provisional"`. **100% of beta providers were trust-orphaned from receiving fixes** without manual `macprovider-cli update` via SSH per Mac.

Observed today on `air5` (self-mint, provisional) sitting at v1.7.0 while v1.7.9 was already published. Coord advertised the recommendation, provider received it, provider skipped:

```
{"target_version":"1.7.9","outcome":"skipped","reason":"provisional",...}
```

## Design choice (kept minimal)

I considered coord-side auto-promotion `provisional → pinned` after 24h clean soak. Rejected because it would ALSO lift:

- Pool cap (`provisional_pool_max`) — DoS defense
- Admission rate limit (`provisional_admission_rate_per_hour`) — DoS defense
- Routing-weight halving — trust proxy on the money path

Those are load-bearing tier restrictions. This PR loosens **only the auto-update gate**, leaves everything else exactly as-is. Provider-side opt-in matches the shape of the existing `auto_update_enabled` config toggle. Operator flips it per-Mac when they trust the coordinator's version recommendation surface.

## Why (fix 2)

[...]

PR #344, opened 2026-07-03
fix(autotune): drop keychain path, use file-only (no more "login keychain password" prompt after auto-update)

Augustas11:
## Summary

Every auto-updated provider will see a `login` keychain password prompt the first time an operator interactively runs `macprovider-cli autotune --recommend --apply` after a version bump. macOS binds the keychain-item ACL to the *specific creating binary's code-signature hash*; auto-update replaces the binary with a new hash — same Developer ID team (`YF7XNRJUG4`), same identifier (`macprovider-cli`), different per-binary hash — and macOS treats the new binary as an unknown requester.

This PR drops the keychain code path entirely. The file at `~/.config/macprovider/autotune-hmac-secret` (mode `0600` under `0700` parent) becomes the sole storage.

## Observed today (2026-07-03)

`mac` provider (Augustas Air) after auto-updating from v1.7.10 to v1.7.11:

keychain prompt

Prompt fired the moment autotune tried to load the HMAC secret.

## Why keychain wasn't giving us anything

- **Under launchd** (background, no UI): keychain API returns `errSecInteractionRequired` without prompting; existing `try?` in `loadOrCreate` swallowed it and fell through to file. Keychain path was already dead weight there.
- **Under interactive operator runs**: prompt → user friction on every version bump.
- **Threat model for HMAC autotune log integrity**: assumes attacker with code execution on the same Mac. Keychain doesn't defend against that. File at `0600` under `0700` parent is comparable protection for the actual threat.

## Alternative I considered

`SecAccessCreate` with the designated requirement (`identifier "macprovider-cli" and anchor apple generic and ... [subject.OU] = YF7XNRJUG4`) — would trust any future release-signed binary silently. But: existing operators still hit a one-time prompt during migration (their existing keychain item is still bound to the old ACL), and the added complexity earns no security benefit over the file path.

## Migration for existing operators

Existing operators have a now-orphaned keychain item that is never read. They can leave it or delete it:

```bash
security delete-generic-password -s "live.streamvc.macprovider.autotune" -a "hmac-secret-v1"
```

No automatic delete is attempted — `SecItemDelete` would also trip the ACL prompt for the exact reason this fix exists. The inline comment in `AutotuneRecommend.swift` calls this out for operators grepping the source.

## Tests

- **`testHMACSecretStoreDoesNotCallKeychainAPIs`** (new) — locks the invariant by grepping the source file for `SecItemCopyMatching`, `SecItemAdd`, and `kSecAttrService`. Any future refactor that reintroduces keychain calls fails this test. Comment mentions of the legacy service literal are intentionally allowed (so operators can grep source for the name they need to delete via `security` CLI).
- Existing `testHMACIdentityUsesSeparateDomainsAndSecretFileIs0600` and `testHMACSecretFileRotatesRecoverableRegularFileFailuresAndRejectsSymlink` unchanged and still green.

```
swift test --filter AutotuneRecommendTests  # all green
swift build -c release                       # green
```

## Not tested in this PR

No integration test for the launchd non-interactive flow (was already relying on `try?` swallowing `errSecInteractionRequired`). The keychain-removal is behaviour-preserving under launchd; the change is only observable in the interactive path.

## Related

- PR #341 provisional-tier auto-update opt-in — surfaces this bug for the first time at scale because it enables auto-update for the beta-scale (provisional-tier) provider population, which will then repeatedly hit this prompt on every version bump.

### DIFF
--- a/phase3-binary/Sources/macprovider-cli/AutotuneRecommend.swift
+++ b/phase3-binary/Sources/macprovider-cli/AutotuneRecommend.swift
@@ -200,22 +200,35 @@ struct AutotuneHMACSecretStore {
     var randomBytes: (Int) throws -> Data = Self.secureRandomBytes
 
     func loadOrCreate() throws -> Data {
-        if usesDefaultPath {
-            if let keychainSecret = try? Self.loadKeychainSecret(), keychainSecret.count == 32 {
-                return keychainSecret
-            }
-        }
+        // Keychain path removed 2026-07-03: macOS binds the keychain-item ACL
+        // to the specific creating binary's code-signature hash. Auto-update
+        // replaces the binary with a new hash → ACL check fails → macOS
+        // prompts every operator for the "login" keychain password on the
+        // next interactive autotune run after each release. Under launchd
+        // (non-interactive) the API returns errSecInteractionRequired and
+        // the code silently falls through to file — so the keychain path
+        // was already dead weight for the auto-updated background flow,
+        // and pure UX drag for the interactive foreground flow.
+        //
+        // The file at ~/.config/macprovider/autotune-hmac-secret is created
+        // at 0600 under a 0700 parent (see writeNewFileSecret + ensurePrivate
+        // ParentDirectory). HMAC of autotune log integrity does not need
+        // keychain-level protection — the threat model (someone with code
+        // execution on this Mac forging autotune logs on the same Mac) is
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
LGTM — clean deletion, invariant lock test catches regression at source. No prompts for any auto-updating provider going forward.

PR #353, opened 2026-07-04
feat(stats): wire pool.Registry into SPEC-017 overview snapshot

Augustas11:
## Summary
Replaces `statsrollup.ZeroSnapshotProvider` with a new `internal/stats/poolsnapshot` adapter that derives the SPEC-017 §5.1.1 live overview fields from `pool.Registry.Snapshot()` on each rollup tick. Unblocks exposing real network stats on malibu.tech via the already-public `stats.streamvc.live/v1/stats/overview`.

**Newly live on `/v1/stats/overview`:**
- `nodes_online` — `State` ready|busy, excluding bearerless-duplicates and pending-receipt-rotation
- `nodes_hardware_attested` — subset with `AttestationStatusAttested`
- `unified_ram_gb_total` — sum of `RAMGB`
- `models_serving` — distinct `ModelID` across online providers
- `network_utilization_pct` — `(SlotsTotal - SlotsFree) / SlotsTotal * 100`

**Still zero (follow-up):** `bandwidth_gb_per_s`, `network_power_kw`, `gpu_cores_total`, `cpu_cores_total`. `pool.Provider` doesn't carry chip identity today; SPEC-026 onboarding does capture `chip` — a follow-up joining the two behind a per-chip TDP/cores/bandwidth profile lookup fills these in.

## R1 audit — three lanes converged, 0 CRITICAL / 0 HIGH / 0 MEDIUM

| Lane | C/H/M | LOWs | Verdict |
|---|---|---|---|
| Code | 0/0/0 | 3 | READY TO MERGE |
| Security | 0/0/0 | 0 | READY TO MERGE |
| Architect | 0/0/0 | 4 | READY TO MERGE |

Prompts: `specs/AUDIT_STATS_SNAPSHOT_PROVIDER_{CODE,SECURITY,ARCHITECT}_R1_PROMPT.md`
Results: `specs/STATS_SNAPSHOT_PROVIDER_R1_{CODE,SECURITY,ARCHITECT}_audit.md`

**Notable security clearances:** SPEC-017 explicitly authorizes every field this PR flips zero→live as public (SEC-1); adapter never reads `PendingReceiptPubkey` bytes, only checks `len(...) > 0` (SEC-4); rollup tick is fixed-cadence 30s default so no DoS amplification (SEC-6); adapter output is order-invariant so ETag stays deterministic (SEC-8).

## LOWs deferred (per stop-iterating-on-low-audits discipline)

[...]

PR #352, opened 2026-07-04
[codex] Make provider stats visible in the app

Augustas11:
## Summary

Implements `specs/BUILD_SPEC_026_IMPL_STEP_3_UX_STATS_SLICE_PROMPT.md` for SPEC-026 Step 3.

- Adds personalized onboarding earnings estimates from the existing local autotune output path, with omission when no rate-card estimate is available.
- Expands the macOS dashboard into the running-provider control center with optional stats, earnings, trust, queue, thermal, model, and log-tail surfaces.
- Wires optional control-socket GPU, latency, queue, request, token, and uptime fields through `AgentSnapshot` without treating missing signals as zero.
- Adds thermal monitoring, live bounded/redacted log tail, menu-bar queue/thermal/badge behavior, and `.idle` no-wallet copy.
- Adds targeted tests for codec optionality, dashboard rendering, estimates, log tail bounds/redaction, badge threshold persistence, and thermal mapping.

## Audit

Three-lane R4 audit converged:

- Code: `RESULT: 0 CRITICAL / 0 HIGH / 0 MEDIUM`
- Security: `RESULT: 0 CRITICAL / 0 HIGH / 0 MEDIUM`
- Architecture: `RESULT: 0 CRITICAL / 0 HIGH / 0 MEDIUM`

See `specs/SPEC-026-IMPL-STEP-3-audit.md` for artifact paths and LOW-only carry items.

## Validation

- `xcodegen generate`
- `git diff --check`
- `xcodebuild build -project phase3-binary/app/Malibu.xcodeproj -scheme Malibu -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO`
- Targeted Step 3 tests: 25 tests, 0 failures

Known unrelated full-suite residual: `RegisterClientTests.testSharedSpec026RegisterFixtureCanonicalizes` still fails because the shared fixture lacks `body_without_signature`; this is outside the Step 3 UX stats slice.

PR #355, opened 2026-07-04
feat(provider): idle-prewarm timer keeps MLX Metal warm during idle

Augustas11:
## Summary

New `IdlePrewarmer` actor that fires a synthetic 1-token internal inference every 30 s of provider idle to keep MLX Metal state warm. Targets the 4-5 s cold-TTFT spike that appears whenever the provider has been idle for ≥ 90 s. Runs through a new `ModelRuntime.runInternalWarmup(...)` entry point that bypasses request tracking, receipt emission, and slot accounting so the coordinator's view of provider capacity is bit-identical to today.

## Motivation — measured data

Cold-start bisection against `mac` (M5 32GB, v1.8.0) on 2026-07-04:

| Path | Cold total | Warm total | Cold penalty |
|---|---:|---:|---:|
| Buyer → api.streamvc.live | 6.06 s | ~500 ms | ~5.5 s |
| M5 → 127.0.0.1:18080 (direct) | 0.61 s | 0.17 s | ~440 ms |
| Gateway `wall_ms` on cold buyer request | 5590 ms | ~1200 ms | ~4390 ms |

Coord `routing_decision` fires ~1 s after buyer send; the remaining ~4.5 s is provider-side MLX Metal cold-start + KV state re-initialisation. Provider log around the cold window shows `event=thermal_state_changed from=fair to=nominal`, consistent with GPU power-managed idle. WSS keepalive at 5 s cadence is working — NAT idle isn't the cause.

## Change

- New file `phase3-binary/Sources/macprovider-cli/IdlePrewarmer.swift` (353 lines) — actor with background tick loop.
- New file `phase3-binary/Tests/macprovider-cliTests/IdlePrewarmerTests.swift` (438 lines, 18 test cases).
- `ModelRuntime.swift` — new `runInternalWarmup(maxTokens:prompt:shouldCancel:)` entry point that:
  - Does NOT increment `providerStatus.requestsTotal`/`requestsInFlight`.
  - Does NOT emit any `ReceiptAudit` line.
  - Does NOT contribute to `throughputTPSSinceLast` / `avgLatencyMSSinceLast`.
  - Reuses the same MLX inference path as `complete()` for the Metal-warmup effect.
- `ProviderStatus.swift` — new `noteRealRequestStart/End`, `lastPrewarmAt`, `secondsSinceLastActivityOrPrewarm()`, `noteInternalPrewarm(at:)`.
- `HTTPServer.swift` — real request handler now calls `providerStatus.beginRequest()` BEFORE `idlePrewarmer.cancelInflightPrewarm()`, then dispatches. Order closes an R1-audit-found race.
- `MacProviderCLI.swift` — 6 new CLI flags + wire-up. `--idle-prewarm-on-battery` uses ArgumentParser `.prefixedNo` inversion so `--no-idle-prewarm-on-battery` can override yaml back to false.
- `MacProviderCore/Config.swift` — new `idle_prewarm:` yaml block, config validation for each of the 6 knobs with named-field error messages.

## Gating logic

Prewarm fires only when **all** of the following are true at the tick:

- `config.enabled == true`
- `providerStatus.requestsInFlight == 0`
- `seconds since max(lastActivity, lastPrewarm) >= idleThresholdSeconds` (closes the SECURITY R1 amplification vector)
- `thermalGate.currentThermalState()` is `.nominal` or `.fair`
- Power source is not battery (unless `run_on_battery == true`)
- Model is loaded

Immediately before `fireWarmup()` the tick performs an actor-owned re-check of `requestsInFlight` and idle time — closes the CODE R1 H2 race where a real request could arrive between the initial snapshot and the fire decision.

## Coordinator invisibility

`providerStatus.slotsFree` reported to the coord is invariant to prewarm activity — asserted in `testSlotsFreeInvariantBeforeDuringAfterPrewarm`. No new WSS message type. No receipt emission. No new heartbeat fields. Autotune's `Stage1Iterator` / `Stage2HillClimb` / `AutotuneDB` are not touched by the prewarm path.

## Audit convergence — three codex lanes

- **CODE R1**: 0 C / **2 H** / **1 M** / 0 L / 0 I
  - H1 `stop()` shutdown race (loop task not awaited, no cancellation check between `runTick()` awaits) → fixed via `stopped: Bool` flag + `await loopTask?.value` in `stop()` + cancellation checks at every await boundary in `runTick()`.
  - H2 `cancelInflightPrewarm()` called before `providerStatus.beginRequest()` in HTTP handler → fixed by swapping order + adding actor-owned re-check inside `runTick()` immediately before `fireWarmup()`.
  - M test coverage gap (13 cases vs 14 required, missing SIGINT case) → fixed; suite now 18 tests.
- **SECURITY R1**: 0 C / 0 H / **1 M** / 1 L / 0 I
  - M prewarm amplification (`runTick()` gated only on `lastActivityAt`, so after crossing idle threshold, every tick fired another prewarm forever) → fixed via `secondsSinceLastActivityOrPrewarm()` gate + `noteInternalPrewarm` called on prewarm START (so a cancelled prewarm still counts toward cooldown, preventing rapid-resume amplification).
- **ARCHITECT R1**: 0 C / 0 H / **1 M** / 3 L / 1 I
  - M `--idle-prewarm-on-battery` bare positive flag couldn't override yaml `true` back to false → fixed with `inversion: .prefixedNo`.
- **CODE R2 · SECURITY R2 · ARCHITECT R2**: 0 / 0 / 0 / 0 / 0 across all three lanes.

## Deferred R1 LOW / INFO findings (ship as-is)

- **ARCHITECT L1** — prewarmer starts before HTTP listener is bound; the BUILD prompt R5 allows this as the fallback ordering. If startup grows more async in future, revisit and hook prewarm start to a "listener bound" callback.
- **ARCHITECT L2** — `ThermalGate.currentState()` vs `currentThermalState()` naming mismatch with the BUILD prompt's stated name. Naming taste; no functional impact.
- **ARCHITECT L3** — `powerSource` parameter not defaulted on `IdlePrewarmer.init`; `clock` parameter present but unused (kept as a seam for future fake-clock migration).
- **ARCHITECT I1** — activity-tracking methods on `ProviderStatus` not grouped under a shared `// Activity tracking …` comment region.
- **SECURITY L1** — operator-configurable `prompt` string is echoed verbatim at startup in the resolved-config print (structured `idle_prewarm_*` events correctly OMIT the prompt body).
- **CODE R1 LOW** — new tests use real `Task.sleep` (bounded ≤ 200 ms per case) rather than an injected fake clock; a `clock: Clock` seam is present on `IdlePrewarmer.init` for future migration.

None of the deferred LOWs affect correctness, security posture, or coordinator-observable behaviour.

## Test plan

- [x] `cd phase3-binary && swift build --product macprovider-cli` clean
- [x] `swift test --filter IdlePrewarmerTests` — 18 tests, 0 failures
- [x] `swift test` full suite — 840 tests, 7 skipped, 0 failures (up from 835 baseline)
- [x] Existing tests unchanged
- [ ] CI green on this PR (pending)

## Config summary

Yaml block (`idle_prewarm:` at top of provider config):

```yaml
idle_prewarm:
  enabled: true                    # default true; --idle-prewarm / --no-idle-prewarm to override
  idle_threshold_seconds: 30       # range [5, 3600]
  tick_seconds: 5                  # range [1, 60]
  max_tokens: 1                    # range [1, 8]
  prompt: warm                     # length [1, 64] UTF-8 bytes
  run_on_battery: false            # default OFF for battery hygiene
```

CLI overrides: `--idle-prewarm / --no-idle-prewarm`, `--idle-prewarm-idle-threshold-s`, `--idle-prewarm-tick-s`, `--idle-prewarm-max-tokens`, `--idle-prewarm-prompt`, `--idle-prewarm-on-battery / --no-idle-prewarm-on-battery`. CLI > yaml > default.

## Observability

Five structured JSON events, printed to stdout in the existing `log_format: json` style:

- `event=idle_prewarm_fired` — successful start; fields: `idle_seconds`, `max_tokens`, `thermal_state`, `on_battery`.
- `event=idle_prewarm_completed` — successful completion; fields: `elapsed_ms`, `tokens_generated`, `first_token_ms`.
- `event=idle_prewarm_skipped` — tick eligibility failed; field: `reason` (`disabled`|`busy`|`not_idle_yet`|`thermal_pressure`|`on_battery`|`model_not_loaded`).
- `event=idle_prewarm_cancelled_by_real_request` — cancellation on real-request arrival; field: `elapsed_ms`.
- `event=idle_prewarm_failed` — error path; fields: `error_class`, `elapsed_ms`.

Grep-stable event names for future Prometheus / stats.streamvc.live wire-up in a follow-up PR.

## Follow-up work (not in scope)

- Metrics wire-up (Prometheus counter per event class).
- The `powerSource` and `clock` default-injection cleanup (ARCH L3).
- SPEC-023 autotune multi-slot recommendation for M-Max / M-Ultra (A1b).
- Pearl TCP kernel tuning (C1).

### DIFF
--- a/phase3-binary/Sources/MacProviderCore/Config.swift
+++ b/phase3-binary/Sources/MacProviderCore/Config.swift
@@ -66,6 +66,12 @@ public struct AppConfig: Equatable, Sendable {
     public var ctlSocketPath: String?
     public var switchStatePath: String?
     public var donorMode: Bool
+    public var idlePrewarmEnabled: Bool
+    public var idlePrewarmIdleThresholdSeconds: Double
+    public var idlePrewarmTickSeconds: Double
+    public var idlePrewarmMaxTokens: Int
+    public var idlePrewarmPrompt: String
+    public var idlePrewarmRunOnBattery: Bool
     // SPEC-001: provider authentication token (closes XSEC-1 from
     // audits/2026-06-10/REPO_AUDIT.md). When set, the binary sends
     // "Authorization: Bearer <token>" on the WS connect and the
@@ -123,6 +129,12 @@ public struct AppConfig: Equatable, Sendable {
             ctlSocketPath: nil,
             switchStatePath: nil,
             donorMode: false,
+            idlePrewarmEnabled: true,
+            idlePrewarmIdleThresholdSeconds: 30,
+            idlePrewarmTickSeconds: 5,
+            idlePrewarmMaxTokens: 1,
+            idlePrewarmPrompt: "warm",
+            idlePrewarmRunOnBattery: false,
             providerToken: nil,
             managedBy: nil
         )
@@ -152,6 +164,12 @@ public struct CLIOverrides: Equatable, Sendable {
     public var kvBits: Int?
     public var maxContext: Int?
     public var maxBatch: Int?
+    public var idlePrewarmEnabled: Bool?
+    public var idlePrewarmIdleThresholdSeconds: Double?
+    public var idlePrewarmTickSeconds: Double?
+    public var idlePrewarmMaxTokens: Int?
+    public var idlePrewarmPrompt: String?
+    public var idlePrewarmRunOnBattery: Bool?
 
     public init(
         port: Int? = nil,
@@ -172,7 +190,13 @@ public struct CLIOverrides: Equatable, Sendable {
         managedBy: String? = nil,
         kvBits: Int? = nil,
         maxContext: Int? = nil,
-        maxBatch: Int? = nil
+        maxBatch: Int? = nil,
+        idlePrewarmEnabled: Bool? = nil,
+        idlePrewarmIdleThresholdSeconds: Double? = nil,
+        idlePrewarmTickSeconds: Double? = nil,
+        idlePrewarmMaxTokens: Int? = nil,
+        idlePrewarmPrompt: String? = nil,
+        idlePrewarmRunOnBattery: Bool? = nil
     ) {
         self.port = port
         self.model = model
@@ -193,6 +217,12 @@ public struct CLIOverrides: Equatable, Sendable {
         self.kvBits = kvBits
         self.maxContext = maxContext
         self.maxBatch = maxBatch
+        self.idlePrewarmEnabled = idlePrewarmEnabled
+        self.idlePrewarmIdleThresholdSeconds = idlePrewarmIdleThresholdSeconds
+        self.idlePrewarmTickSeconds = idlePrewarmTickSeconds
+        self.idlePrewarmMaxTokens = idlePrewarmMaxTokens
+        self.idlePrewarmPrompt = idlePrewarmPrompt
+        self.idlePrewarmRunOnBattery = idlePrewarmRunOnBattery
     }
 }
 
@@ -235,6 +265,7 @@ public enum ConfigLoader {
         config = try applyEnvironment(config, environment: environment)
         config = try applyCLI(config, cli: cli)
         config.configPath = configPath
+        try validateIdlePrewarm(config)
         return config
     }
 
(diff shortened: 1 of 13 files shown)


antfleet-ops (approved):
Approved via required-review pattern. Three codex lanes converged R1 → R2 at 0 C/H/M across CODE, SECURITY, ARCHITECT. All 13 CI checks green. Ship.

antfleet-ops (approved):
Re-approve after rebase onto origin/main (PRs #352 + #353 landed while B1 was in review). Diff-content unchanged. CI green on all 13 checks. Ship.

PR #360, opened 2026-07-04
release: sign + notarize + staple Malibu.app in release.yml

Augustas11:
## Summary

Extends `.github/workflows/release.yml` so the same tag ships both a signed+notarized `macprovider-cli` tarball (existing) AND a signed+notarized+**stapled** `Malibu.app.zip` (new). Unblocks SPEC-026 v2 onboarding smoke tests on real Macs + beta-cohort App distribution.

No new secrets — reuses the existing Developer ID Application cert (Superposition Technologies YF7XNRJUG4) and notary credentials.

## What lands

- **New step:** Install XcodeGen on runner (Homebrew)
- **New step:** Build Malibu.app via `xcodegen generate` + `xcodebuild archive` (`CODE_SIGNING_ALLOWED=NO` at archive; sign after)
- **New step:** Sign Malibu.app with `--deep --options runtime --timestamp --entitlements Malibu.entitlements`, notarize via `notarytool submit --wait`, `stapler staple`, `stapler validate`, package as `Malibu-{tag}.zip`
- **Refactor:** CLI-step cleanup moved to end-of-job `if: always()` step so single `build.keychain` lifetime spans both signing steps (rejected re-importing certs in a second keychain per audit — doubles credential exposure)
- **Extended:** `scripts/test-tier2-provider-release.sh` and `scripts/verify-tier2-provider-release.sh` for Malibu.app.zip coverage

## Validation

**Full smoke via `workflow_dispatch` on prerelease tag `v1.8.1`** ([run 28712563208](https://github.com/Augustas11/macprovider/actions/runs/28712563208)):

- Install XcodeGen — ok
- Build Malibu.app (xcodegen + xcodebuild archive, arm64) — ok
- Sign + notarize binary (CLI, existing) — ok
- **Sign + notarize + staple Malibu.app** — end-to-end:
  - `codesign --deep --options runtime --timestamp` — ok
  - `codesign --verify --strict --deep` — ok
  - `xcrun notarytool submit --wait` — Accepted
  - `xcrun stapler staple` — "The staple and validate action worked!"
  - `xcrun stapler validate` — "The validate action worked!"
  - `spctl -a` — "Malibu.app: accepted" (Gatekeeper)
- Clean signing material (refactored end-of-job step fired correctly) — ok
- Tier-2 provider artifact preflight failed — **orthogonal** — pre-existing `binaryVersion = "1.8.0"` vs. tag `v1.8.1` guard; would fail any unbumped-version release attempt; not caused by this PR

[...]

PR #365, opened 2026-07-04
release: ship Malibu.app as signed .pkg (drop .zip artifact)

Augustas11:
## Summary

Follow-up to #360 — fixes an inconsistency where the CLI ships as a signed `.pkg` (double-click install) but Malibu.app was shipping as a `.zip` (unzip + manual drag). Now both use the same delivery pattern.

## What changes

- **Malibu.app is now shipped as `Malibu-{tag}.pkg`** — pkgbuild with `--install-location "/Applications"`, productsign with Developer ID Installer cert, notarized separately from the .app inside, stapled + validated + spctl-checked.
- **The `.zip` artifact is removed entirely.** No analogous consumer (unlike the CLI's `.tar.gz` which install.sh needs); shipping both would waste release page real estate.
- **Fails loud** if `APPLE_DEVELOPER_ID_INSTALLER_CERT_P12_BASE64` is missing — refuses to silently fall back to .zip. Same secret already required for the CLI's `.pkg`; asymmetric behavior across CLI and App would surprise operators and beta providers.
- Release-notes copy updated: "Download `Malibu-{tag}.pkg` and double-click to install" replaces "unzip and drag."

## Validation

YAML parses OK. `release.yml` grep confirms all Malibu artifact references are `.pkg`; only `NOTARIZE_ZIP` remains (transient zip for `notarytool submit` of the .app, not the shipped artifact — Apple's notary accepts .zip submission of .app bundles, this is not user-facing).

Full end-to-end runner validation happens on the first `workflow_dispatch` v1.8.2 firing after merge — using the same signing / notarization / stapling primitives that #360 already validated for both the CLI .pkg and Malibu.app.

## Why not both .zip + .pkg

User was clear: shipping .zip alongside .pkg would still leave the .zip on the release page, undermining the fix. Beta cohort trust hinges on "one artifact, one path." Power users can `pkgutil --expand` the .pkg if they need raw bundle access.

## Test plan

- [ ] CI green on this PR
- [ ] antfleet-ops approval
- [ ] Squash-merge to main
- [ ] `workflow_dispatch` v1.8.2 as prerelease from main
- [ ] `Malibu-v1.8.2.pkg` lands on GitHub Releases, downloadable + double-click installs to /Applications

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -329,6 +329,7 @@ jobs:
         shell: bash
         env:
           APPLE_DEVELOPER_ID_CERT_P12_BASE64: ${{ secrets.APPLE_DEVELOPER_ID_CERT_P12_BASE64 }}
+          APPLE_DEVELOPER_ID_INSTALLER_CERT_P12_BASE64: ${{ secrets.APPLE_DEVELOPER_ID_INSTALLER_CERT_P12_BASE64 }}
           APPLE_NOTARY_APPLE_ID: ${{ secrets.APPLE_NOTARY_APPLE_ID }}
           APPLE_NOTARY_PASSWORD: ${{ secrets.APPLE_NOTARY_PASSWORD }}
           APPLE_NOTARY_TEAM_ID: ${{ secrets.APPLE_NOTARY_TEAM_ID }}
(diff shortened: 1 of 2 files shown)


antfleet-ops (dismissed):
LGTM — fixes CLI/App artifact-format asymmetry (.pkg for both). Same signing/notarize/staple primitives as #360, adds pkgbuild+productsign for /Applications install-location. Fails loud on missing Installer cert instead of silent fallback.

antfleet-ops (approved):
Re-approve after 01e5e9e: hardcoded binaryVersion assertions bumped 1.8.1→1.8.2 to sync with 684c1bc.

PR #362, opened 2026-07-04
feat(stats): enable SPEC-017 stats subsystem in production

Augustas11:
## Summary
Flips `stats.enabled=true` in the tracked coordinator config and wires the three Postgres DSNs (reader / rollup / provider_portal) via `env:` indirection. Turns `stats.streamvc.live/v1/stats/{overview,leaderboard,health}` from **404 → 200** after next deploy.

This is the final step of the malibu.tech stats bring-up: PR #353 wired the SnapshotProvider code path, PR #357 tracked `dist/coordinator.yaml` as authoritative, and this PR flips the flag.

## Live-state verified before writing this PR (per new `verify-runtime-surface-before-designing` discipline)

| Fact | Verified how |
|---|---|
| Postgres 16 already running on Pearl | `systemctl status postgresql` |
| `macprovider_stats` DB exists | `psql -l` |
| All 6 SPEC-017 migrations applied (14 stats_* tables incl. schema_migrations_spec017) | `psql -c "\dt"` |
| Roles `stats_reader / stats_rollup / provider_portal` exist as LOGIN + scram-sha-256 | `pg_roles.rolcanlogin` + TCP roundtrip test |
| `bind_address=127.0.0.1` (validator requirement per `config.go:1390`) | grep on tracked yaml |
| Endpoint currently returns 404 | `curl https://stats.streamvc.live/v1/stats/overview` |

Passwords generated on Pearl via `openssl rand -hex 24`, never transited to my terminal. DSN env vars appended to `/etc/macprovider/coordinator.env` (0640 root:macprovider).

## R1 three-lane codex audit — all PASS

| Lane | C/H/M | LOWs | Verdict |
|---|---|---|---|
| Code | 0/0/0 | 1 | READY TO MERGE |
| Security | 0/0/0 | 0 | READY TO MERGE (fully clean) |
| Architect | 0/0/0 | 5 | READY TO MERGE |

Prompts: `specs/AUDIT_STATS_ENABLE_CONFIG_{CODE,SECURITY,ARCHITECT}_R1_PROMPT.md`
Results: `specs/STATS_ENABLE_CONFIG_R1_{CODE,SECURITY,ARCHITECT}_audit.md`

## LOWs deferred (per stop-iterating-on-low-audits)

- **CODE-L1** — `dist/coordinator.yaml.example` doesn't document the `stats:` section shape. Follow-up: add commented reference.
[...]

PR #366, opened 2026-07-04
Revert PR #362: stats subsystem enable (missing OLTP source path)

Augustas11:
## Summary
Reverts #362. `stats.enabled` back to false (subsystem code path stays in the binary but is not mounted). Endpoint returns to 404 instead of `stats_stale`.

## Why
#362 turned on the SPEC-017 stats subsystem without a working OLTP source path. The rollup expects `ledger_request_credits` + `provider_tokens` in Postgres but they live in SQLite (`billing/store.go:58`, `auth/tokens.go:250`). Rollup ticks fail every 30s with `relation "ledger_request_credits" does not exist`. Consequences:
- `/v1/stats/overview` returns `{"error": {"code": "stats_stale"}}` instead of 404
- `/v1/stats/health` returns `status: "down"` (leaderboard_24h always down)
- Coordinator journal fills with rollup error logs

Reverting restores the pre-#362 posture: endpoint 404, no rollup goroutines, no error spam.

## Deferred cleanup (not in this PR)
- 3 DSN env vars in `/etc/macprovider/coordinator.env` on Pearl (`STATS_READER_DSN`, `STATS_ROLLUP_DSN`, `STATS_PROVIDER_PORTAL_DSN`) become dead. Safe to leave — nothing reads them once `stats.enabled=false`.
- 3 Postgres roles on Pearl (`stats_reader`, `stats_rollup`, `provider_portal`) stay LOGIN + password-set. Safe to leave — nothing connects as them.
- Postgres `macprovider_stats` DB + migrations 001-006 stay. Harmless.

Ops cleanup is a separate follow-up. Doesn't gate this revert.

## Test plan
- [x] yaml parses cleanly, no `stats:` section
- [x] `go build ./...` clean
- [ ] Post-deploy: `curl https://stats.streamvc.live/v1/stats/overview` returns 404; coordinator journal has no more `stats_rollup_tick_error` lines


antfleet-ops (approved):
Revert #362 — stats subsystem enable was premature; OLTP source path (SPEC-005 ledger in Postgres) was assumed but never built. Restoring 404 posture.

PR #393, opened 2026-07-05
Fix launch provider autotune apply config

Augustas11:
## Summary

Fixes the Malibu.app launch-provider first-run path so the app can persist apply-ready autotune serve configuration without calling the hanging `macprovider-cli autotune --apply` path.

The CLI now emits a typed top-level `serve_config` in `autotune --recommend --json`, derived from the same selected benchmark plus signed catalog row used by `--apply`. Malibu.app consumes that explicit payload only, persists the recommendation-owned config fields, and validates shape before download/startup/resume launch paths.

## Root Cause

The app displayed CLI recommendation JSON but did not persist the artifact/catalog provenance fields required by `macprovider-cli serve`, leaving configs with only identity/link state and causing serve join refusal.

## Validation

- `git diff --check`
- `swift test` from `phase3-binary`: 850 passed, 7 skipped, 0 failures
- `xcodegen generate && xcodebuild test -project Malibu.xcodeproj -scheme Malibu -destination 'platform=macOS'` from `phase3-binary/app`: 89 passed, 0 failures
- Final Codex IMPL audits: CODE 0/0/0, SECURITY 0/0/0, ARCH 0/0/0 with one non-blocking INFO

## Audit Artifacts

- BUILD audit: `specs/BUILD_FIX_LAUNCH_PROVIDER_AUTOTUNE_APPLY_BUILD_AUDIT.md`
- IMPL audit: `specs/BUILD_FIX_LAUNCH_PROVIDER_AUTOTUNE_APPLY_IMPL_AUDIT.md`

## Not Tested

- Live fresh-slate packaged GUI smoke against production coordinator.


antfleet-ops (approved):
Approved after 3-lane BUILD/IMPL audit convergence and full local verification.

PR #409, opened 2026-07-06
Malibu: CLI install onboarding + menu-bar UI fixes

Augustas11:
## Summary
- **Option A onboarding:** "Launch Provider" runs bundled `install.sh` (`MACPROVIDER_NO_PROMPT=1`), imports CLI config to Keychain via `importExistingCLIConfig()`, registers the Malibu login item, and monitors the launchd-managed provider over `/v1/health` instead of spawning a second CLI child.
- **Policy default:** `MalibuOnboardingPolicy.prefersCLIInstallTrack = true` — fresh installs show onboarding without `MALIBU_ONBOARD_V2` / UserDefaults v2 flag. Legacy SPEC-026 in-app register/autotune path remains available when the policy is disabled (tests).
- **UI from keen-chaum-ff9ea8 (51bbdb0):** vertically center the menu-bar horizon+sun icon; remove "Quit and Uninstall…" from the menu (handler retained for future entry point).
- **Research artifact:** `scratchpad/onboarding-audit-2026-07-06.md` documents the pre-change onboarding map and UX gaps.

## Test plan
- [x] `xcodebuild -project phase3-binary/app/Malibu.xcodeproj -scheme Malibu -only-testing:MalibuTests test` (109 tests pass)
- [ ] Fresh Mac / clean state: launch Malibu → Launch Provider → verify install.sh log stream, launchd service, import succeeds, dashboard shows serving via health poll
- [ ] Reboot: verify launchd provider survives and Malibu attaches on startup when configured
- [ ] M5 32GB Tier C: confirm donor-mode path works via install.sh (no Swift `serve_config` JSON decode)
- [ ] Visual: menu-bar icon centered in light/dark mode

Made with [Cursor](https://cursor.com)

PR #415, opened 2026-07-06
Fix Malibu launchd monitor races after CLI install onboarding

Augustas11:
## Summary
- Use `MalibuOnboardingTimeouts.firstServingFrameSec` (600s) as the default health wait for launchd monitoring, matching onboarding's first-serving timeout.
- Release any Malibu-spawned CLI child before attaching to the launchd-managed provider, and try launchd monitoring before the spawned-child guard in `start()`.
- Propagate monitor success/failure through onboarding: `finalizeCLIInstallTrack()` and `startConfiguredViaCLIInstall()` now fail with a retryable error when the background provider never becomes healthy.

Fixes three Bugbot findings from PR #409 (duplicate CLI spawn, stale child race, false `.live` on monitor failure).

## Test plan
- [x] `xcodebuild -scheme Malibu build` succeeds
- [ ] Cold model load: onboarding waits up to 600s before failing (not 120s)
- [ ] Re-launch Malibu with existing launchd provider: no duplicate `macprovider-cli` processes
- [ ] Kill launchd provider during finalize: onboarding shows retryable failure instead of `.live`

### DIFF
--- a/phase3-binary/app/Sources/Malibu/Agent/MalibuAgent.swift
+++ b/phase3-binary/app/Sources/Malibu/Agent/MalibuAgent.swift
@@ -55,13 +55,15 @@ final class MalibuAgent: ObservableObject {
     func start() async {
         // AUDIT R2 CODE H3 fix.
         guard !isShuttingDown else { return }
-        guard child == nil else { return }
 
         // Option A: when install.sh owns the provider via launchd, monitor it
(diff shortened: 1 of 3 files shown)


antfleet-ops (approved):
Launchd monitor race fixes: 600s timeout, child release before attach, honest onboarding failure.

PR #430, opened 2026-07-06
fix(control-plane): transactional admin state, VPS-anchored deploy safety, coordinator deploy hygiene, operator-key strength check

Augustas11:
## Summary

Admin control-plane and Pearl VPS deploy hardening (deepsec P1-1).
Makes gateway admin state mutations transactional so a failed write
cannot masquerade as success; kill-switch updates use optimistic
concurrency so concurrent operators can no longer silently overwrite
each other. Reworks the gateway deploy script so the DB path used
for snapshot/rollback comes from the config actually installed on
the VPS, and the pre-deploy C2 gate validates the VPS-installed
config rather than the developer's local file. Coordinator deploy
hardened with unpredictable temp-file paths for the smoke probe,
DSN redaction on the drift diff output, and an operator-key
strength check that refuses to boot with placeholder / short /
low-entropy keys.

### gateway/admin (Cluster A)
- `internal/router/admin.go`: capacity transition + kill-switch
  handlers wrap state mutation in a single transaction; write errors
  propagate as 500 with a specific error class instead of masquerading
  as 200. Kill-switch update uses optimistic concurrency
  (`UPDATE ... WHERE version=?`); row-count 0 returns 409 Conflict
  with the current version so the client refetches.
- `internal/router/admin_metrics.go` (new): counters for
  `admin_state_write_error_total` labelled by handler and
  `admin_state_concurrent_conflict_total` labelled by resource.
- `internal/storage/sqlite/store.go` + `interfaces.go` + `types.go`:
  shared compare-and-swap helper returning `(applied bool, err error)`.

### gateway/deploy (Cluster B)
- `dist/deploy-pearl-vps.sh`: SSHes into the VPS, reads
  `db.path` from the installed `/etc/macprovider-gateway.yaml`,
  and uses that path for both snapshot and rollback. Hardcoded
  default removed; script fails loudly if the installed config has
  no `db.path`. Pre-deploy C2 gate reads the VPS-installed config,
  not local `gateway.yaml*`; explicit `--dry-run-local` flag for
  developer workflow only. Deploy log records the SHA-256 of the
  config that was actually validated.
- Shell tests `gateway_deploy_snapshot.test.sh` and
  `gateway_deploy_c2_precheck.test.sh` cover custom `db.path`
  targeting, VPS-vs-local precheck distinction, and the
  `--dry-run-local` escape hatch.

### coordinator/deploy + config (Cluster C)
- `dist/deploy-pearl-vps.sh`: static feed smoke uses `mktemp -d`
  with a `trap` cleanup on EXIT — no more predictable `/tmp` paths.
  Drift diff output passes through a `redact_dsn` sanitizer that
  strips `postgres://user:password@host` before printing.
- `internal/config/config.go`: operator-key strength check at load —
  refuses keys shorter than 32 bytes, matching a small denylist of
  known placeholders (`changeme`, `PLACEHOLDER`, `test`, etc.), or
  with Shannon entropy below the minimum threshold. Boot fails with
  a specific error naming which check failed.
- Shell tests `coord_deploy_smoke_probe.test.sh`,
  `coord_deploy_drift_redact.test.sh`,
  `coord_deploy_c2_precheck.test.sh`.
- `internal/config/config_test.go` +64 lines covering each
  denylist entry, length boundary, entropy floor, and specific
  error messages.

### Audit prompts
- `specs/AUDIT_FIX_CONTROL_PLANE_{CODE,SECURITY,ARCHITECT}_R1.md`.

## Audit

Three codex lanes fired independently per project convention. All
three converged to acceptance at 0 CRITICAL / 0 HIGH / 0 MEDIUM.
Consistent with prior-wave audit discipline; final round result
files were not committed to disk.

## Test plan

- [x] `go test -count=1 ./phase5-gateway/internal/router/... ./phase5-gateway/internal/storage/...` — green
- [x] `go test -count=1 ./phase4-coordinator/internal/config/...` — green
- [x] `git diff --check`
- [ ] Shell test suite: `dist/test/gateway_deploy_*.test.sh` and
      `dist/test/coord_deploy_*.test.sh` (pre-merge — should run in
      CI too)
- [ ] Pearl staging smoke: deploy runs the new C2 precheck against
      installed VPS config; snapshot targets the configured db.path;
      coordinator boot with a placeholder operator key refuses to
      start (pre-merge staging gate)

## Release cadence

Shell script + operator-key check changes take effect on the next
deploy. Compiled Go changes (gateway admin transactional writes,
coordinator config check) ship on next respective release cut. Code
lands on main now so upcoming release cuts pick them up
automatically.

## Merge coordination

No downstream wave blocks on this one. Remaining deepsec work after
this ships: P1-3 (stats/rollup, ~12 items), P2 (spec/harness, ~10
items) — both independent, neither security-critical.

### DIFF
--- a/phase4-coordinator/cmd/coordinator/main_test.go
+++ b/phase4-coordinator/cmd/coordinator/main_test.go
@@ -667,7 +667,7 @@ func reloadTestServers(cfg config.Config) (config.Config, *pool.Registry, *provi
 }
 
 func reloadTestServersWithLogger(cfg config.Config, logger zerolog.Logger) (config.Config, *pool.Registry, *providerws.Server, *buyer.Server) {
-	cfg.Auth.OperatorKey = "operator-key"
+	cfg.Auth.OperatorKey = "0123456789abcdefABCDEFghijklmnop"
 	cfg.Pool.WarmupGateEnabled = false
 	registry := pool.NewRegistry(nil)
 	wsServer := providerws.NewServer(cfg, registry, logger)
@@ -702,7 +702,7 @@ func fetchReloadTier2Metadata(t *testing.T, server *buyer.Server) struct {
 } {
 	t.Helper()
 	req := httptest.NewRequest(http.MethodGet, "/internal/routing", nil)
-	req.Header.Set("Authorization", "Bearer operator-key")
+	req.Header.Set("Authorization", "Bearer 0123456789abcdefABCDEFghijklmnop")
 	rr := httptest.NewRecorder()
 	server.InternalHandler().ServeHTTP(rr, req)
 	if rr.Code != http.StatusOK {
--- a/phase4-coordinator/cmd/coordinator/migrate_indexes_test.go
+++ b/phase4-coordinator/cmd/coordinator/migrate_indexes_test.go
@@ -45,7 +45,7 @@ func TestMigrateIndexesCheckIsReadOnly(t *testing.T) {
 	}
 	_ = db.Close()
 
-	if err := os.WriteFile(configPath, []byte("auth:\n  operator_key: test-operator-key\nstorage:\n  db_path: "+dbPath+"\n"), 0o644); err != nil {
+	if err := os.WriteFile(configPath, []byte("auth:\n  operator_key: 0123456789abcdefABCDEFghijklmnop\nstorage:\n  db_path: "+dbPath+"\n"), 0o644); err != nil {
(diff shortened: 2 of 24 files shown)


antfleet-ops (approved):
Approved: Wave 6 control-plane hardening (deepsec P1-1). Transactional admin state, VPS-anchored deploy safety, operator-key strength check. 3-lane codex audits closed 0 C/H/M.

PR #452, opened 2026-07-06
Fix SSE-error fallback overbill reconciliation

### LINKED ISSUE #449: Reconciler counts prompt tokens as overbill on SSE-error fallback pairs (contradicts README's #232 corroboration exclusion)
Follow-up from #444. Deployed post-#436/#442/#443/#444 stack (gateway + coord both at `v1.8.16-5-g936f0f6`) plus the scenario 15 chaos smoke on live Pearl now shows a bilateral reconciler complaint that appears to contradict the README's own I1 semantics.

## Reproduction

Chaos-harness scenario 15 (`test/network-harness/scenarios/15_provider_reconnect_storm.yaml` from PR #431 draft) against post-#444 Pearl. Artifact bundle: `test/network-harness/artifacts/15_provider_reconnect_storm-20260706T173207Z/`.

Ran fresh against gateway `v1.8.16-5-g936f0f6` + coord `v1.8.16-5-g936f0f6` (coord uptime 97s at scenario start — freshly deployed, 1 provider ready, no other traffic).

## Observed

```
invariant I1 [FAIL]: billing-ledger reconciliation drift == 0
  gateway over-billed by 64 vs harness above tolerance 0 across 1 pair(s)
invariant I3 [FAIL]: no charged-tokens > delivered-tokens
  gateway charged 64 tokens above delivered tolerance 0 across 1 request(s)
```

## The pair the reconciler is flagging

| Field | Value |
|---|---|
| gateway_outcome | `stream_output_exceeded` (fallback settlement) |
| gateway_total_tokens | 64 |
| gateway_completion_tokens | **0** ✓ (post-#444 fix working) |
| coordinator_total_tokens | 63 |
| coordinator_completion_tokens | **0** ✓ (matches gateway) |
| harness_total_tokens | 0 |
| harness_saw_sse_error_event | **true** ✓ |
| harness_sse_error_code | `stream_output_exceeded` |
| **gateway_overbill_vs_harness_tokens** | **+64** ← the flag |
| **gateway_overbill_vs_coordinator_tokens** | **+1** ← tokenizer noise |

The 64-token "overbill" is the **prompt tokens** — gateway legitimately consumed 64 prompt tokens routing the request (coord recorded 63, a 1-token tokenizer-drift). Completion tokens are correctly zero on both sides. Buyer parser saw the SSE error envelope and rejected the malformed tail, so `harness_total_tokens = 0`.

## Why this looks like a reconciler bug, not a real overbill

Per `test/network-harness/README.md` §"The four hard invariants" I1 rule 5:

> positive gateway-vs-harness overbill across pairs (`GatewayOverbillVsHarnessTokens` > 0; "ok" outcomes always count, and fallback outcomes count **unless the buyer ALSO received the gateway's terminal SSE error envelope** — the corroboration check that closes the #232 trust gap)

The buyer here received the SSE error envelope (`harness_saw_sse_error_event: true`). Per the documented rule, this pair is a "fallback outcome + corroborated by SSE error", which means it should be **excluded** from the overbill sum. But the reconciler is including it — hence the +64 fires I1 and I3.

Either:

1. The reconciler's exclusion logic in `test/network-harness/internal/reconcile/ledger.go` isn't checking `harness_saw_sse_error_event` for this specific `stream_output_exceeded` fallback subclass, or
2. The README's stated semantics are outdated and this pair really should count for overbill (in which case the doc + the invariant name both need adjustment — "charged-tokens > delivered-tokens" reads like completion-token accounting, and prompt tokens on an error-terminated stream are legitimately consumed).

Either fix is fine, but the current state where the reconciler and README disagree isn't. Also, since prompt tokens are counted for coord too, `gateway_overbill_vs_coord: 1` here is the real signal (1-token tokenizer drift is noise); the +64 vs harness is the reconciler-side asymmetry.

## Suggested fix

Update `unmatched_reconcile` / `overbilledPairs` calculation to honor the corroboration check: when `harness_saw_sse_error_event = true` on a fallback outcome, exclude the pair from `gateway_overbill_vs_harness_tokens` regardless of the specific outcome subclass.

Recommend adding a regression test that mirrors the scenario 15 shape (streamed 200 → SSE error mid-stream → gateway writes 0 completion but 64 prompt → invariant PASS) alongside the existing #226 per-pair semantics tests.

## Related

- #444 (closed via 936f0f6) — the completion-token settlement fix that this residual sits on top of
- #232 — the original SSE-envelope corroboration check the README cites
- #226 — the per-matched-pair I1 semantics change
- PR #431 (draft) — chaos harness scenario 15, currently blocked-by this + the sibling Residual Y

## Not tested

- Whether the same overbill pattern appears in the happy-path smoke scenarios today. If not, this is chaos-specific; if yes, it's a general reconciler drift on every fallback outcome with SSE error envelope.

Augustas11:
## Summary
- suppress actionable harness-axis overbill only for fallback outcomes corroborated by strict buyer-visible terminal SSE error envelopes
- keep raw drift diagnostics visible while preserving no-SSE, malformed-envelope, and code/outcome mismatch trust-gap failures
- harden the buyer SSE parser so malformed top-level error frames cannot import usage tokens or satisfy fallback corroboration

## Validation
- go test ./... in test/network-harness
- go vet ./... in test/network-harness
- 3-lane audit loop: code/security/architecture all 0 Critical, 0 High, 0 Medium

## Not tested
- live Pearl scenario 15 chaos harness rerun

Closes #449

### DIFF
--- a/test/network-harness/internal/buyer/loadgen.go
+++ b/test/network-harness/internal/buyer/loadgen.go
@@ -579,7 +579,8 @@ type chunkPayload struct {
 //   - code: the non-empty `error.code` value if the dispatched event
 //     is a STANDALONE error envelope; "" otherwise.
 //   - isStandaloneError: true ONLY if the event has a non-empty
-//     `error.code` AND no `choices` AND no `usage` tokens.
+//     `error.code`, no `choices` key, no `usage` key, no duplicate
+//     top-level keys, and no duplicate immediate `error.*` keys.
 //
 // The caller in consumeSSE tracks whether the LAST DISPATCHED event
 // before `[DONE]`/EOF was a standalone error envelope; only that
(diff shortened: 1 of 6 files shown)


antfleet-ops (approved):
Approved after Codex three-lane audit loop: code/security/architecture all reported 0 Critical, 0 High, 0 Medium findings. Local validation passed: go test ./... and go vet ./... in test/network-harness.

PR #453, opened 2026-07-07
Accept text-only structured content arrays

### LINKED ISSUE #445: SPEC-006 §4.12 missing 'no multimodal content arrays' divergence; buyer-side 400 message is misleading
## Summary

The buyer-facing chat completions endpoint rejects the OpenAI-standard structured `content` array shape with a **400 "Invalid message content"**, but SPEC-006 §4.12 does not list this as a known v1 divergence. This creates two related problems:

1. **SPEC-006 §4.12 divergence list is incomplete.** SPEC-002 §2725 documents "no multimodal content arrays in v1" as the intended narrowing at the coordinator layer. SPEC-006 (the *buyer-facing* contract) does not carry this divergence forward, but SPEC-006 §349 requires that "Deviation from OpenAI's chat completion request/response shape MUST be documented as a known divergence." Users pulling up SPEC-006 (or the front-door docs it drives) have no signal that this shape is unsupported.
2. **The rejection error is not actionable.** The response body says "Invalid message content", which reads to an SDK user like "your JSON is malformed" when their JSON is in fact a valid OpenAI SDK-emitted shape. The right message is something like "multimodal content arrays are not supported in v1; use `content: string`" so a caller can immediately correct the request without reading the source.

## Discovery context

Found while running adversarial-harness scenario 25 (malformed_payloads) probes against a local rig. The `content-as-array` probe was intended to test JSON-parsing robustness; it surfaced this documented-in-SPEC-002-but-not-in-SPEC-006 gap instead.

## Live repro

```bash
# Two requests differ only in the content shape.
demo=$(curl -sS -X POST http://127.0.0.1:19443/auth/demo-session \
  | python3 -c "import json,sys;print(json.loads(sys.stdin.read())['demo_token'])")

# Content as array (OpenAI multimodal shape, text-only) → 400
curl -sS -X POST http://127.0.0.1:19443/v1/chat/completions \
  -H "Content-Type: application/json" -H "X-Demo-Token: $demo" \
  -d '{"model":"m","messages":[{"role":"user","content":[{"type":"text","text":"hello"}]}]}'
# → HTTP 400 {"error":{"code":"invalid_request","message":"Invalid message content",...}}

# Content as string (control) → 503 no_provider_available (reached routing normally)
curl -sS -X POST http://127.0.0.1:19443/v1/chat/completions \
  -H "Content-Type: application/json" -H "X-Demo-Token: $demo" \
  -d '{"model":"m","messages":[{"role":"user","content":"hello"}]}'
# → HTTP 503 {"error":{"code":"no_provider_available",...}}
```

The string form reaches routing, so this is definitively a buyer-side validation gate for the content shape, not a downstream failure.

## Code path

`phase4-coordinator/internal/buyer/server.go:4715-4720`:

```go
for i, m := range messages {
    switch m.Role {
    case "system", "user":
        if !rawStringNonEmpty(m.Content) {
            return http.StatusBadRequest, "invalid_request", "Invalid message content"
        }
```

`rawStringNonEmpty` at `server.go:4898`:

```go
func rawStringNonEmpty(raw json.RawMessage) bool {
    var s string
    return json.Unmarshal(raw, &s) == nil && s != ""
}
```

The `Unmarshal` into `string` refuses any array-shaped `json.RawMessage`, which is where the 400 originates.

## Spec surface

- **SPEC-006 §4.12 (buyer-facing contract, `specs/SPEC-006-buyer-api.md:875-891`):** lists 6 v1 divergences (no tool execution, tool fields syntactically accepted only, no strict structured outputs, 503 possible, model lineup live-pool dependent, gateway-estimated usage). **Content-as-array is not listed.**
- **SPEC-006 §349:** "Deviation from OpenAI's chat completion request/response shape MUST be documented as a known divergence."
- **SPEC-002 §2725 (coord spec, `specs/SPEC-002-coordinator.md:2725`):** documents the narrowing correctly: "`content` must be a non-empty string (no multimodal content arrays in v1)".
- **SPEC-001 §957 (phase3 binary spec):** same narrowing documented for tier-1 providers.
- **SPEC-015 §4.3.1 (`specs/SPEC-015-receipts.md:1297-1318`):** the *receipt/settlement* schema DOES model content-as-array with `image_url` parts. So the divergence is at the buyer-entry gate specifically, not lower layers.

## SDK impact

The OpenAI Python SDK's typed message parameters (`ChatCompletionUserMessageParam`, `ChatCompletionContentPartTextParam`, `ChatCompletionContentPartImageParam`) let callers pass:

```python
client.chat.completions.create(
    model="...",
    messages=[{"role":"user","content":[{"type":"text","text":"hi"}]}]
)
```

This shape works against OpenAI's own API. It fails against macprovider with the misleading "Invalid message content" error. Any SDK-emitted request that constructs `content` structurally (multimodal, agent shapes, some framework wrappers like LangChain's default message serialization) is affected.

## Proposed remediation

Two options, non-exclusive:

**Option A — accept the shape, extract text.** For `[{"type":"text","text":"..."}, ...]`, flatten the text parts into a single string before forwarding to the provider. Refuse any non-text `type` (e.g., `image_url`) with a specific 400 (`multimodal_not_supported` code). Preserves OpenAI-SDK compat for the text-only structured shape; still narrows on multimodal.

**Option B — keep the narrowing, fix the docs + error.**
1. Add to SPEC-006 §4.12 v1 divergences: "Structured `content` arrays (multimodal-style `[{"type":"text","text":"..."}]`) are not accepted in v1; `content` must be a JSON string."
2. Change the error message at `phase4-coordinator/internal/buyer/server.go:4720` from `"Invalid message content"` to something like `"content must be a JSON string in v1 (structured content arrays not supported)"` and use a distinct error code (e.g., `unsupported_content_shape`) so SDK error handlers can branch.
3. Mirror the divergence in front-door docs.

Option A gives strictly better OpenAI-compat for the most common shape SDK users hit (text-only structured content is what agent frameworks emit even without images). Option B is the minimum-diff fix that at least stops the false-positive "your request is invalid" signal.

## Not a bug

- Not a code bug in the sense that the implementation matches SPEC-002's stated design.
- Not a security issue — the gate rejects a well-formed request; no data leak, no injection surface.

## Severity

Product/docs gap. Low severity, high visibility once an SDK-native user hits it.

Augustas11:
## Summary
- Accept text-only structured `messages[].content` arrays for `system` and `user` messages and normalize them to string content before provider dispatch.
- Reject multimodal/non-text content parts with `unsupported_content_shape` and an actionable message.
- Document the v1 multimodal divergence in SPEC-006, the SDK cookbook, and the public gateway docs template.

Closes #445.

## Validation
- `cd phase4-coordinator && go test ./internal/buyer -count=1`
- `cd phase4-coordinator && go test ./... -count=1`
- `cd phase5-gateway && go test ./... -count=1`

## Notes
- Unchanged string-content requests keep the existing raw pass-through behavior.
- Normalized requests are reserialized only when a supported text-only content array is present.

### DIFF
--- a/docs/using-macprovider-with-openai-sdk.md
+++ b/docs/using-macprovider-with-openai-sdk.md
@@ -40,6 +40,11 @@ print(resp.choices[0].message.content)
 
 Any MLX chat model in the pool works. Available models: `client.models.list()` or `curl https://api.streamvc.live/v1/models`.
 
+Text-only structured message content is accepted too. Frameworks that emit
+`content=[{"type":"text","text":"Hello"}]` are normalized to plain text before
+provider dispatch. Multimodal parts such as `image_url` are not supported in
+v1 and return `unsupported_content_shape`.
+
 ## Streaming
 
 ```python
--- a/docs/using-macprovider-with-openai-sdk.md
+++ b/docs/using-macprovider-with-openai-sdk.md
@@ -40,10 +40,10 @@ print(resp.choices[0].message.content)
 
 Any MLX chat model in the pool works. Available models: `client.models.list()` or `curl https://api.streamvc.live/v1/models`.
 
-Text-only structured message content is accepted too. Frameworks that emit
-`content=[{"type":"text","text":"Hello"}]` are normalized to plain text before
-provider dispatch. Multimodal parts such as `image_url` are not supported in
-v1 and return `unsupported_content_shape`.
+Text-only structured content is accepted for `system` and `user` messages too.
+Frameworks that emit `content=[{"type":"text","text":"Hello"}]` for those roles
+are normalized to plain text before provider dispatch. Multimodal parts such as
+`image_url` are not supported in v1 and return `unsupported_content_shape`.
 
 ## Streaming
 
--- a/phase4-coordinator/internal/buyer/server.go
+++ b/phase4-coordinator/internal/buyer/server.go
@@ -1477,6 +1477,11 @@ type chatMessage struct {
 	ToolCalls  json.RawMessage `json:"tool_calls"`
 }
 
+type chatContentPart struct {
+	Type string          `json:"type"`
+	Text json.RawMessage `json:"text"`
+}
+
 type requestToolCall struct {
 	ID        string
 	Name      string
@@ -4372,6 +4377,10 @@ func validateChatRequest(body []byte) (chatRequest, int, string, string) {
 	if err := json.Unmarshal(messagesRaw, &req.Messages); err != nil || len(req.Messages) == 0 {
 		return req, http.StatusBadRequest, "invalid_request", "Invalid messages"
 	}
+	var rawMessages []map[string]json.RawMessage
+	if err := json.Unmarshal(messagesRaw, &rawMessages); err != nil || len(rawMessages) != len(req.Messages) {
+		return req, http.StatusBadRequest, "invalid_request", "Invalid messages"
+	}
 	if v, ok := raw["stream"]; ok {
 		if err := json.Unmarshal(v, &req.Stream); err != nil {
 			return req, http.StatusBadRequest, "invalid_request", "Invalid stream"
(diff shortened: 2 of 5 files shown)


antfleet-ops (approved):
Approved from antfleet-ops after the three-lane audit cleared 0 Critical/High/Medium findings.

PR #455, opened 2026-07-07
Malibu: tail provider logs and surface stale-catalog startup failures

antfleet-ops:
## Summary

- Tail launchd provider stdout/stderr into Malibu during startup and steady-state monitoring so the dashboard log panel is live.
- Scan tailed lines for known CLI serve-preflight failures (stale catalog provenance, artifact hash mismatch, missing autotune pins) and fail fast with actionable copy instead of an infinite Starting spinner.
- Show redacted log tail plus diagnosis in onboarding and propagate providerStartFailure into launch failures.

Fixes the Malibu reboot incident class where launchd restart-looped on stale catalog provenance while the UI showed no explanation.

## Test plan

- [x] xcodebuild MalibuTests (116 tests, 0 failures)
- [ ] CI phase3-binary swift test green
- [ ] Manual: stale-catalog line in macprovider.err.log surfaces in onboarding

### DIFF
--- a/phase3-binary/app/Sources/Malibu/Agent/AgentSnapshot.swift
+++ b/phase3-binary/app/Sources/Malibu/Agent/AgentSnapshot.swift
@@ -149,6 +149,8 @@ enum AgentSnapshotPresenter {
             return s.currentModelID
         case .reconnecting:
             return s.lastError ?? "Checking background provider…"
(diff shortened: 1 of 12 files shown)


Augustas11 (approved):
CI green.

PR #469, opened 2026-07-07
fix(cli): soften catalog provenance envelope gate on serve start

antfleet-ops:
## Summary
- Stop crash-looping serve when coordinator republishes the candidate catalog but the configured model row is still admitted
- Keep row-level catalog admission as the hard security gate; treat stale `model_catalog_version`/`model_catalog_hash` as a warning only
- Update Malibu log diagnostics to match the new warning text

## Test plan
- [x] `swift test --filter ServeCommandTests.testCoordinatorJoinAcceptsCatalogBoundSnapshotWithStaleProvenanceEnvelope`
- [x] Provider verified healthy on operator Mac after manual config patch


Augustas11 (approved):
Row admission stays the hard gate; envelope mismatch is warning-only. CI green.

PR #364, opened 2026-07-04
fix(gateway): enforce OAuth flag, atomic issuance ceilings, oauth_state TTL/prune, audit-after-auth + feedback bounds

Augustas11:
## Summary

Gateway public-auth perimeter hardening. Enforces the GitHub OAuth feature
flag at route-mount and handler entry, replaces check-then-insert races on
per-IP issuance ceilings with atomic rolling-window reservations, adds TTL
+ per-source cap + background pruning to the OAuth state table, and
bounds the feedback endpoint on both write and read paths. Reorders
internal-header probe middleware so unauthenticated probes can no longer
force persistent audit-DB writes.

### gateway/config + boot
- `GitHubOAuthEnabled` propagated to route-mount and handler-entry gates
  (belt-and-suspenders: routes unmounted when disabled, handlers still
  refuse if reached via another path).
- Boot log reflects effective OAuth state.
- New `gateway.yaml.example` fields for OAuth state pruner and issuance
  ceilings.

### gateway/router
- Atomic per-IP ceilings for oauth / signup / demo issuance (rolling
  window, preserves legacy signup/demo semantics).
- Audit-write reordered after auth-check so unauthenticated internal-
  header probes can no longer force persistent DB writes.
- Feedback endpoint bounded on writes: per-IP rate limit, body/comment
  size caps, source-header requirement.
- Admin feedback summary bounded on reads: LIMIT + aggregate over
  returned rows, not full table.

### gateway/storage
- `oauth_state` migration: `expires_at` column, per-IP write cap,
  `PruneExpiredOAuthState` called from main lifecycle pruner.
- New atomic reservation types for oauth/signup/demo issuance.
- Storage error surface adds bounded rate-limit + cap error classes.

### specs
- `AUDIT_FIX_GATEWAY_AUTH_ABUSE_{CODE,SECURITY,ARCHITECT}_R1.md`.

## Audit

Three codex lanes fired independently per project convention.

| Lane | R2 result |
| --- | --- |
| SECURITY | PASS, 0 CRITICAL / 0 HIGH / 0 MEDIUM / 0 LOW |
| CODE | PASS, 0 CRITICAL / 0 HIGH / 0 MEDIUM (1 LOW carried) |
| ARCHITECTURE | PASS, 0 CRITICAL / 0 HIGH / 0 MEDIUM (1 LOW carried) |

Security lane cleared the higher 0-LOW bar required for externally
reachable auth paths. Both carried LOWs are non-blocking; see "Deferred"
below.

[...]

PR #471, opened 2026-07-07
perf(T2-01): compiled decode scaffolding (MACPROVIDER_COMPILED_DECODE, default OFF)

antfleet-ops:
## Summary
- Port `CompiledDecode.swift` from `perf/mlx-compile-bf16` — env-flag gated compiled decode step + `KVCacheUpdatableAdapter`.
- Wire `decode-bench` compiled path when `MACPROVIDER_COMPILED_DECODE=1`.
- Log flag in `ModelRuntime.stream()`; full runtime wire-in deferred.
- Add scaffold unit tests for env flag helpers.

## Test plan
- [x] `swift test` — 968 tests pass
- [ ] **Do not enable in production** — correctness gate FAILED on Qwen2.5-7B: `MLX.compile()` + `KVCacheSimple.offset` (Swift `Int`) bakes cache position into graph → token repetition loop instead of EOS

## Notes
- Flag defaults **OFF**. No production behavior change unless env set.
- Fix blocked on upstream: KV cache offset must be graph-traceable (`MLXArray`), likely same mlx-swift-lm release as Gemma MoE (#364).
- Artifact: `beta/throughput-engineering/T2-01-compiled-decode-wire-in.md`
- Merge after #469 (T2-02) to reduce conflict risk on `DecodeBenchCommand.swift`.

### DIFF
--- a/beta/throughput-engineering/T2-01-compiled-decode-wire-in.md
+++ b/beta/throughput-engineering/T2-01-compiled-decode-wire-in.md
@@ -1,0 +1,193 @@
+# T2-01 — MLX.compile() Decode Wire-In
+
+**Date:** 2026-07-07  
+**Branch:** `perf/t2-01-compiled-decode`  
+**Runbook:** `specs/PLAN_THROUGHPUT_ENGINEERING_RUNBOOK.md` § T2-01  
+**Verdict:** YELLOW
+
+---
+
+## Summary
+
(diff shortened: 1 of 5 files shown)


Augustas11 (approved):
T2-01 scaffolding only; flag default OFF; correctness blocked on upstream KV graph.

PR #482, opened 2026-07-08
fix(integration): harden spec-014 healthz wait under parallel CI

Augustas11:
## Summary
- Increase `waitForHTTP` deadline in `spec_014_v0_2_pairing` from 10s to 30s
- The cross-service `integration` job runs `./...` in parallel; spec_014 and the root harness each build coordinator binaries, so `/healthz` startup can exceed 10s under CI load
- `TestSpec014V02FullPairingHandoff` passes reliably in the isolated `spec-014-v0-2-gate` job but flaked in the parallel `./...` run

## Test plan
- [x] `go test -race -count=1 -timeout 5m ./spec_014_v0_2_pairing -run TestSpec014V02FullPairingHandoff`
- [x] `go test -race -count=1 -timeout 5m ./...` (full cross-service integration suite)


antfleet-ops (approved):
LGTM — minimal healthz timeout bump for parallel CI contention.

PR #493, opened 2026-07-08
fix(cli): CS-3 stop control-socket metrics poll draining the coordinator heartbeat window

Augustas11:
## Summary

Fixes **CS-3 (MEDIUM, payout-adjacent)** from the 2026-07-08 audit. `ControlMetricsBuilder.build` read the shared `ProviderStatus` since-last window with `resetWindow: true`. That window is owned by the coordinator heartbeat (`CoordinatorClient.sendHeartbeat`), whose since-last deltas feed autotune demand ranking and **SPEC-023 payout-first scoring**. Because the control socket and the heartbeat share one `ProviderStatus` instance (injected in `MacProviderCLI`), a Malibu.app metrics poll (UI cadence) landing between heartbeats **drained the window**, truncating the next heartbeat's since-last figures to the post-poll sliver and under-reporting served requests to the coordinator.

## Fix

- The metrics poll now reads with `resetWindow: false`. The coordinator heartbeat remains the **sole** `resetWindow: true` caller and the sole owner of window rollover. All-time cumulative counters were never affected.
- Malibu.app consumes `metrics_response` as display state (verified), so cumulative-since-last-heartbeat window values are correct for the UI — no consumer relied on per-poll deltas.

## Test

`ControlMetricsBuilderTests.testMetricsBuildDoesNotDrainHeartbeatWindow`: record one finished request → poll metrics → assert the next heartbeat (`resetWindow: true`) still observes that request (would be 0 before the fix), then clears it. Existing `ProviderStatusTests` (21) green.

## Audit

Independent codex review (code + concurrency lane): **PASS, 0 Critical/High/Medium**. Confirmed the heartbeat is the only production `resetWindow: true` caller, the actor-isolated reset introduces no double-reset/leak, and Malibu.app display semantics are unaffected.

### DIFF
--- a/phase3-binary/Sources/macprovider-cli/ControlMetricsSnapshot.swift
+++ b/phase3-binary/Sources/macprovider-cli/ControlMetricsSnapshot.swift
@@ -80,7 +80,13 @@ enum ControlMetricsBuilder {
         guard let providerStatus else {
             return ControlMetricsSnapshot()
         }
-        let snapshot = await providerStatus.snapshot(resetWindow: true)
+        // CS-3: the control-socket metrics poll (Malibu.app UI cadence) must NOT
+        // reset the since-last window. That window is owned by the coordinator
+        // heartbeat (CoordinatorClient.sendHeartbeat), whose since-last deltas feed
+        // autotune demand ranking and SPEC-023 payout-first scoring. Draining it
+        // here — the two share one ProviderStatus instance — truncated the next
+        // heartbeat's window to the post-poll sliver. Read without resetting.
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Approved (user-authorized). CS-3 payout-adjacent metrics-window fix, Swift-only in phase3-binary. Codex code+concurrency lane 0 C/H/M; regression test proves the poll no longer drains the heartbeat window. The phase4-coordinator Postgres integration failure is unrelated flakiness (this PR touches no coordinator code); re-running.

PR #495, opened 2026-07-08
fix(deploy): OPS-1 remove superseded opoi-v0 systemd drop-in on malibu emission deploy

Augustas11:
## Summary

Fixes **OPS-1 (MEDIUM, deploy-safety)** from the 2026-07-08 audit. `deploy-malibu-emission-pearl.sh` installed `malibu-emission.conf` but left the earlier `opoi-v0.conf` systemd drop-in in place. systemd applies `.service.d` drop-ins in lexicographic filename order and the last `ExecStart=` reset+set wins; `opoi-v0.conf` sorts **after** `malibu-emission.conf`, so its `ExecStart` (`--config-overlay coordinator.opoi-v0-staging.yaml`) won and the coordinator ran against the OPoI-only overlay instead of the merged `coordinator.pearl-overlays.yaml` that step 7 `--validate-config` had just checked — a validate-vs-runtime divergence. Enabling `malibu_emission` in `pearl-overlays.yaml` then silently had no effect.

## Fix

- **Step 5:** `rm -f opoi-v0.conf` during install, so exactly one drop-in defines `ExecStart`. Its config is already folded into `pearl-overlays.yaml` by the step 3 merge (`merge-yaml-overlay.py` recursively merges `pool:` OPoI keys with `malibu_emission:`), so nothing is dropped. `rm -f` is idempotent and safe on fresh hosts / re-runs.
- **Step 7:** assert `systemctl show -p ExecStart` references `coordinator.pearl-overlays.yaml` after `daemon-reload` and **before** restart, failing closed (`exit 1`) — catches any future drop-in that would re-introduce the divergence.
- **Rollback guidance:** rewritten to disable `malibu_emission` in `pearl-overlays.yaml` while **keeping** the drop-in (it now loads the overlay carrying the merged OPoI config). Removing the drop-in would drop OPoI. Guidance is YAML-aware and explicitly warns against a blind `sed` (which could hit another `enabled:` key or break indentation).

## Verification

`bash -n` clean; no new shellcheck warnings (only the pre-existing line-166 `SC2145/SC2064`). Independent codex ops-deploy review: **PASS, 0 Critical/High/Medium** — confirmed OPoI preservation (rendered the merged overlay: `pool.canary_enabled=true` + `malibu_emission.enabled=false` both present), drop-in determinism, the fail-closed `ExecStart` assertion, and the corrected rollback. (A first review pass caught that the initial fix left the rollback guidance stale; that was fixed and re-verified.)

### DIFF
--- a/phase4-coordinator/dist/deploy-malibu-emission-pearl.sh
+++ b/phase4-coordinator/dist/deploy-malibu-emission-pearl.sh
@@ -180,6 +180,14 @@ if [ "$DRY_RUN" != "1" ]; then
     install -d -o root -g root -m 0755 /etc/systemd/system/macprovider-coordinator.service.d
     install -o root -g root -m 0644 $DEPLOY_TMP/malibu-emission.conf \
       /etc/systemd/system/macprovider-coordinator.service.d/malibu-emission.conf
+    # OPS-1: remove the superseded OPoI drop-in. Its ExecStart points at
+    # coordinator.opoi-v0-staging.yaml and, because systemd applies .service.d
+    # drop-ins in lexicographic filename order, opoi-v0.conf sorts AFTER
+    # malibu-emission.conf and its ExecStart would WIN — the coordinator would
+    # then ignore the merged coordinator.pearl-overlays.yaml. The OPoI config is
+    # already folded into pearl-overlays.yaml by the step 3 merge, so removing
+    # this drop-in drops no configuration.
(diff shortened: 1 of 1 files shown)


antfleet-ops (approved):
Approved (user-authorized). OPS-1 deploy-safety fix: removes superseded opoi-v0.conf drop-in, asserts effective ExecStart before restart, corrects rollback to preserve OPoI without blind sed. Codex ops-deploy lane 0 C/H/M (round 2 after fixing a rollback regression).

PR #500, opened 2026-07-08
Fix Tier2 relay sequencing for encrypted concurrency

antfleet-ops:
## What changed

- Fixed Tier2 encrypted relay handling so session-global AEAD p2c frames can arrive out of order under concurrent provider responses.
- Added a bounded sparse replay window (`4096`) with replay rejection and gap cleanup after contiguous advancement.
- Negotiated `response_chunk_plaintext_envelope` through `auth_request`/`auth_response` so new providers carry per-request app chunk `seq` inside authenticated plaintext only after coordinator acknowledgement.
- Preserved cap-false legacy raw encrypted chunk behavior, including wrapper-shaped model output.
- Deferred request-threshold rekey close until all active requests drain, including cancel/error removal paths.

## Why

Issue #379 live capacity/config raised qwen3-coder capacity to eight slots, but concurrent encrypted responses could be sealed/sent in valid non-contiguous AEAD order. The coordinator previously expected strict arrival order, causing `tier2_aead_decrypt_failed`, provider disconnects, and failed concurrent live requests.

Closes #379.

## Validation

- `git diff --check`
- `go test ./... -count=1` in `phase4-coordinator`
- `go test -race ./internal/ws -run 'TestEncryptedRelay(...)|TestDecodeEncryptedInferenceChunkPlaintextRejectsMalformedTypedEnvelope|TestProviderAuth' -count=1`
- `go vet ./...` in `phase4-coordinator`
- `swift test --package-path phase3-binary --filter 'Tier2ProviderSessionTests|InferenceRelayTests|CoordinatorClientTests/testAuthInitialUsesV2EncryptedLegCapabilitiesAndModelHash|CoordinatorClientTests/testAcceptedAuthResponsePairingMaterial_WritesClaimURL'`
- `FORCE_DIRTY=1 ./scripts/build-linux.sh` in `phase4-coordinator`
- 3 audit lanes rerun to 0 Critical/High/Medium findings: code correctness, security, architecture

Issue #379, opened 2026-07-05
[net-perf] C3: HTTP/2 (then HTTP/3) on buyer-facing nginx


Augustas11:
## Motivation

nginx currently serves buyer traffic over HTTP/1.1. Tools that pipeline multiple concurrent requests (multi-agent frameworks, IDE assistants) pay a head-of-line-blocking penalty on shared connections. HTTP/2 multiplexes; HTTP/3 removes the TCP HOL penalty entirely.

Expected win: modest for point-request buyers, meaningful for parallel-fan-out buyers.

## Proposal

Phase 1: HTTP/2. `listen 443 ssl http2;` in the vhost. Reload. Bench.

Phase 2 (later): HTTP/3 / QUIC. Requires nginx build with `--with-http_v3_module` and cert-serving under UDP/443. Ubuntu default nginx doesn't ship this — needs mainline or a source build.

## Success criteria (Phase 1)

- Parallel 8-request bench from one buyer shows lower p95 vs same bench today.
- Point-request bench unchanged (verify no regression).
- Existing openai-compatible clients keep working (Alpn negotiation is transparent).

## Blockers

- Every existing buyer client needs verification that HTTP/2 works with SSE streams. OpenAI Python SDK, litellm, and curl are known-good; less-common clients need probing.
- HTTP/3 requires a package change on Pearl (bigger operational lift).

## Refs

- 2026-07-04 network-perf memo (session)
- C2 (nginx keepalive tuning) is an independent lever — do C2 first.

antfleet-ops:
Reopened after PR #499 merge: the PR made the #379 harness capacity-aware, but it intentionally did not satisfy the final issue-closing condition. The restored MACPROVIDER_HTTP2_RUN_CHAT_BENCH=1 gate still requires a live model with total_slots >= 8 plus sufficient account concurrency; current live evidence at merge time showed qwen3-coder-30b-a3b-instruct advertising total_slots=1. Close this only after the chat benchmark and SDK compatibility runner pass against sufficient live capacity.

PR #501, opened 2026-07-08
Fix v2 provider auth version compatibility

antfleet-ops:
## Summary
- accept numeric-string top-level auth_request.version values while still enforcing decoded v2
- keep normal capacity/heartbeat numeric fields strict
- add bounded bad_field logging for rejected v2 initial auth requests without logging raw unauthenticated parse errors

## Context
PR #500 deployed the Tier2 encrypted concurrency fix, but the live provider is rejected before joining the pool with unrecognized auth message. This hotfix restores compatibility for providers that encode the auth envelope version as a JSON string.

## Validation
- go test ./internal/ws targeted auth compatibility tests
- go test ./... -count=1 in phase4-coordinator
- go vet ./... in phase4-coordinator
- git diff --check
- three audit lanes: code 0 C/H/M, security 0 C/H/M after logging fix, architecture 0 C/H/M

Follow-up to #500 / #379 live deploy validation.

### DIFF
--- a/phase4-coordinator/internal/ws/messages.go
+++ b/phase4-coordinator/internal/ws/messages.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"strconv"
 	"strings"
 
 	"github.com/augstar/macprovider-coordinator/internal/config"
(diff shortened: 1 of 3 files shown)


Augustas11 (approved):
Approved hotfix after checking numeric-string auth version compatibility, bounded unauthenticated rejection logging, local tests/vet, and three audit lanes with 0 C/H/M after the logging fix.

PR #505, opened 2026-07-08
Report catalog model IDs to coordinator

Augustas11:
## Summary
- report `model_catalog_model_id` to the coordinator for auth, hello, heartbeat, and supported-model validation when the served local model key is the configured boot model
- keep local serve/preflight model keys unchanged so live `--model qwen3-coder-30b-a3b-instruct` still passes catalog-key validation
- suppress idle-prewarm events and warm-swap completion heartbeats before coordinator session acceptance to avoid pre-auth non-auth frame races

## Root Cause
Live provider preflight requires the local model/rate-card key, while coordinator admission validates the catalog model ID. Using the local alias on the wire caused the coordinator to reject the provider as uncatalogued/autotune-evidence-invalid; changing the serve model to the full catalog ID broke local preflight.

## Validation
- `swift test --package-path phase3-binary --filter CoordinatorClientTests`
- `swift test --package-path phase3-binary`
- `swift build -c release --package-path phase3-binary`
- `git diff --check`
- Three local audit lanes after implementation: correctness/runtime, protocol/security, regression/test adequacy; 0 Critical/High/Medium findings. Native subagent spawn was unavailable due the existing thread limit, so the lanes were run locally.

## Live Follow-up
After merge, install the release binary on the live provider, keep the alias model in launchd/config with `model_catalog_model_id` present, restart launchd, and verify coordinator registration plus the live HTTP/2 chat benchmark.

### DIFF
--- a/phase3-binary/Sources/macprovider-cli/CoordinatorClient.swift
+++ b/phase3-binary/Sources/macprovider-cli/CoordinatorClient.swift
@@ -180,6 +180,7 @@ actor CoordinatorClient {
     private let maxBodyBytes: Int
     private let maxActiveRequests: Int
     private let supportedModels: [String]?
+    private let catalogModelIDForCoordinator: String?
     private let publishesSupportedModels: Bool
     private let warmSwapEnabled: Bool
     private let hardwareSummary: [String: Any]?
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Reviewed implementation, local validation, and three local audit lanes: 0 Critical/High/Medium findings.

PR #507, opened 2026-07-09
test(e2e): canary buyer probe for continuous network/perf measurement

antfleet-ops:
## Summary

Adds a **canary buyer probe** — a continuous synthetic buyer that measures the
live macprovider network from the buyer's perspective and exports Prometheus
metrics. This is **P1** from the 2026-07-09 e2e-testing review.

The month's testing pattern was that every serious finding came from live
contact with prod (KV-cache verify, W3 canary TTFT false-fail), yet the network
had **zero continuous buyer-perspective measurement**. A metallib-class silent
2× TPS regression, a thermal-throttle degradation, or a provider that
`/v1/status` calls "ready" but that 502s every completion, would be invisible
until a buyer complained.

While building this, the probe caught exactly that last case live: `/v1/status`
reported the sole provider `ready`, verified, 8 free slots — while every
`/v1/chat/completions` returned `502 upstream_provider_error` (3/3 retries). The
probe recorded `up=1, serviceable=0, {502:4, 5xx:1}`, which is the signal.

## What it adds (all under `test/e2e/canary-buyer/`, no product-code changes)

- `probe.mjs` — zero-dependency Node probe. Per model class per run records:
  TTFT p50/p95/p99, sustained decode TPS, sticky turn-2 KV-cache reuse ratio,
  serviceability, and request-outcome counts. Emits Prometheus text-exposition
  metrics + a per-run JSON artifact.
- `run-canary.sh` — wrapper: resolves the buyer token from the documented
  harness location, rotates artifacts, runs the probe. Token never echoed.
- `com.streamvc.canary-buyer.plist` — launchd agent template (30-min cadence).
- `README.md` — usage, metric reference, suggested alert rules.

Productionizes `test/e2e/malibu-console/smoke.mjs` and network-harness scenarios
`07_sustained_throughput` / `09_streaming_ttft_distribution` into a scheduled,
monitorable form. Generates the baseline data that P2 (cold/warm matrix), P3
(soak), and P4 (cache-reuse gate) need for their thresholds.

## Verification

- **Failure path** verified against live prod during an active
  `upstream_provider_error` incident (recorded the divergence correctly).
- **Success path** (TTFT percentiles, decode-TPS math, cache ratio, Prometheus
  emission) verified against a local mock SSE gateway.
- `node --check` + `bash -n` clean.
- Three-lane codex audit (code / security / architect), converged over 4 rounds
  to **0 CRITICAL / 0 HIGH / 0 MEDIUM** on every lane. Each round surfaced new
  territory (not repeats) and every finding was reproduced + fixed + re-verified:
  - R1: SSRF/token-leak on error bodies, README alert label-mismatch, launchd node PATH
  - R2: TTFT/TPS sample conflation; **2 HIGH** — IPv6 SSRF bypasses + token leaking via server-controlled fields; missing `stream_options.include_usage`
  - R3: terminal SSE `{error}` frames counted as healthy; dark-signal sample-count gauges
  - R4: code + architect PASS (security already PASS at R3)

  Carried LOW: pure DNS-rebinding TOCTOU without a pinning agent (documented in code; mitigated by https-only + private-host reject + startup resolve-check + `redirect: 'manual'`).

## Notes

- Buyer-side scope: splits latency into TTFT vs decode window. Finer server-side
  phase timing is in gateway completion logs (join by `x-request-id`), not the
  buyer response, so the probe does not claim it.
- Default exit 0 even on network failure (launchd shouldn't treat a bad-network
  run as a crash); `--fail-on-degraded` opts into exit 1 for CI/alerting.

### DIFF
--- a/test/e2e/canary-buyer/README.md
+++ b/test/e2e/canary-buyer/README.md
@@ -22,7 +22,8 @@ scheduled, dependency-free probe.
 | TTFT | `macprovider_canary_ttft_ms{model,quantile}` | p50/p95/p99 time-to-first-token. Feeds real numbers into the W3 `max_ttft_ms` canary gate instead of guesses. |
 | Decode TPS | `macprovider_canary_decode_tps{model,quantile}` | Sustained tokens/sec (first token excluded). Catches metallib/thermal silent regressions. |
 | KV-cache reuse | `macprovider_canary_cached_prompt_ratio{model}` | Turn-2 `cached_prompt_tokens / prompt_tokens`. Locks in the measured ~45% sticky-affinity win. |
-| Outcomes | `macprovider_canary_requests{model,outcome}` | 2xx / 502 / 5xx / timeout / empty counts (per-run gauge) → 502-rate signal. |
+| Outcomes | `macprovider_canary_requests{model,outcome}` | Per-run gauge over buckets `2xx / 502 / 5xx / timeout / network_error / stream_error / empty / other` → 502-rate and mid-stream-failure signal. |
+| Sample counts | `macprovider_canary_ttft_samples{model}`, `macprovider_canary_decode_tps_samples{model}` | Valid samples collected — alert on `== 0 while serviceable == 1` to catch a signal going dark (e.g. gateway stops returning usage). |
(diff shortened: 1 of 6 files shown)


Augustas11 (approved):
LGTM — canary buyer e2e harness

PR #510, opened 2026-07-09
fix(provider): accept coordinator-advertised catalog model id as serve alias

Augustas11:
## Summary

Fixes a live production outage: **100% of billed buyer completions have 503'd since ~2026-07-08 19:18 UTC** while `/v1/status` showed the sole provider `ready`.

**Root cause (verified against prod):** the provider advertises `config.modelCatalogModelID` (a catalog id like `mlx-community/Qwen3-Coder-30B-A3B-Instruct-4bit`) to the coordinator via `CoordinatorClient.coordinatorWireModelID`, but serves inference locally under `config.model` (the autotune rate-card key, e.g. `qwen3-coder-30b-a3b-instruct`). Its request validation (`ChatCompletionRequest.validateModelMatches`) only accepted the single served id — so the provider **404'd every request for the very id it advertised**. The coordinator relays the buyer's (catalog) id verbatim → provider 404 → relay NAK → coordinator returns `503 no_provider_available` "Selected provider is not reachable". Heartbeat and warmup use the served-id path, so the pool kept showing the provider `ready` (the "admits a provider that can't serve" symptom).

## Fix

Accept the coordinator-advertised catalog id as a **serve alias** in all three `validateModelMatches` sites — HTTP (`HTTPServer`), WS relay (`InferenceRelay`), and runtime (`ModelRuntime.acquireRequestHandle`) — **gated so the alias applies only while the configured model is the one currently loaded**, exactly mirroring `coordinatorWireModelID`'s `servedModelID == loadedModelID` predicate. After a warm-swap to a different model the alias no longer applies.

- Validation is **only relaxed**; the request model is not rewritten and response/receipt model echoing is unchanged. Same weights ⇒ same model hash, so proof-of-weights / settlement are unaffected.
- Default `aliases: []` keeps every non-serve caller (decode-bench, canary, self-test) behavior-identical.
- Alias is sourced from trusted local operator config only; length/case/whitespace handling mirrors `catalogModelIDForCoordinator` normalization.

Design doc: `specs/BUILD_SPEC_relay_serve_model_id_alias.md`.

## Tests

Added regression coverage (all passing):
- `ChatCompletionRequestTests`: default-empty regression, loaded-id pass, alias pass (case-insensitive), neither → 404, empty-alias-ignored.
- `HTTPServerSwapTests`: runtime accepts alias when configured model loaded; **rejects alias after warm-swap to a different model → 404**.
- `InferenceRelayTests`: relay accepts catalog alias when configured model loaded; rejects when no alias configured; **accepts under `warmSwapEnabled` while the configured model is still the served snapshot** (the audit-caught warm-swap gate case).

`swift build` clean; targeted suites pass (`ChatCompletionRequestTests|HTTPServerSwapTests|InferenceRelayTests`).

## Audit (three independent codex lanes, money-path bar 0 C/H/M)

Converged to **0 CRITICAL / 0 HIGH / 0 MEDIUM** across code, security, and architect lanes. Round 1 surfaced one HIGH (code + architect): the relay gated the alias on `warmSwapEnabled ? []`, which would wrongly reject the alias for a warm-swap-enabled provider still serving its configured model. Fixed by gating on `validationModelID == loadedModelID` (+ regression test); round 2 re-audit is clean.

Deferred (non-blocking LOW/INFO, documented per convention):
- The acceptable-id predicate is owned independently at the three validation sites; if a 4th validation path is added, consider centralizing it (e.g. exposed from the runtime snapshot).
- No explicit "matched via alias" telemetry; existing request telemetry still records the buyer-provided model id. Consider a low-cardinality alias-match marker only if operators need future mismatch forensics.

## Deploy note

This is the durable fix. Restoring buyers requires shipping a new provider binary (notarization/AMFI), which is a separate follow-up — buyers remain down until that lands.

### DIFF
--- a/phase3-binary/Sources/MacProviderCore/ChatCompletionRequest.swift
+++ b/phase3-binary/Sources/MacProviderCore/ChatCompletionRequest.swift
@@ -180,13 +180,17 @@ public struct ChatCompletionRequest: Sendable {
         self.conversationKey = conversationKey
     }
 
-    public func validateModelMatches(_ loadedModel: String?) throws {
+    public func validateModelMatches(_ loadedModel: String?, aliases: [String] = []) throws {
         guard let loadedModel else {
             throw APIError(status: 503, message: "Model not loaded", type: "server_error", code: "model_not_loaded")
         }
-        guard Self.asciiCaseInsensitiveEquals(model, loadedModel) else {
-            throw APIError(status: 404, message: "Model not found", code: "model_not_found")
+        if Self.asciiCaseInsensitiveEquals(model, loadedModel) {
+            return
         }
+        for alias in aliases where !alias.isEmpty && Self.asciiCaseInsensitiveEquals(model, alias) {
+            return
+        }
+        throw APIError(status: 404, message: "Model not found", code: "model_not_found")
     }
 
     public var allowsSpeculativeDecoding: Bool {
--- a/phase3-binary/Sources/macprovider-cli/CoordinatorClient.swift
+++ b/phase3-binary/Sources/macprovider-cli/CoordinatorClient.swift
@@ -846,6 +846,7 @@ actor CoordinatorClient {
             modelRuntime: modelRuntime,
             providerStatus: providerStatus,
             loadedModelID: loadedModelID,
+            catalogModelIDAlias: catalogModelIDForCoordinator,
             warmSwapEnabled: warmSwapEnabled,
             maxActiveRequests: maxActiveRequests,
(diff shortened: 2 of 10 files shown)


antfleet-ops (approved):
Approved per operator authorization. Root cause verified against prod; validation-only relaxation gated to configured-model-loaded (mirrors coordinatorWireModelID); 3-lane codex audit converged 0 C/H/M; swift build+tests green; CI green (13/13 incl. phase3-binary swift test).

PR #511, opened 2026-07-09
fix(gateway): clamp over-cap provider usage instead of dropping the frame

Augustas11:
## Summary

Fixes a confirmed prod billing + observability bug: **streaming (and non-streaming) requests whose prompt exceeds the gateway's usage-validation cap silently lost their `usage` frame and were billed on `gateway_estimated` byte estimates instead of provider-reported token counts.**

Discovered by the canary buyer probe (#507) against `api.streamvc.live`, model `qwen3-coder-30b-a3b-instruct`, and root-caused end-to-end (probe → gateway settlement code → Pearl coordinator logs + config).

## Root cause

The gateway's usage-validation cap `maxUsageTokens` is derived from a `ceil(bytes/4)` byte heuristic (`promptCapTokens`) and, by design invariant, **equals the buyer's reservation hold**. `ceil(bytes/4)` is prose-calibrated and under-counts token-dense prompts (code, JSON, CJK) — and prod serves a **coder** model. When the provider's honest `prompt_tokens + completion_tokens` exceeded the cap, [`usageFromJSON`](phase5-gateway/internal/router/chat_proxy.go) returned an error, and the streaming relay set the usage line to `nil` (dropping it from the buyer stream) and fell back to `gateway_estimated`.

Confirmed live:
- Probe: usage frame present ≤~302 prompt tokens, absent ≥~330 (deterministic); small prompt + long completion → present, large prompt + short completion → absent (isolates it to prefill/prompt size).
- Pearl coordinator log for a probe request: `provider reported prompt tokens exceeded independent bound ... provider_reported_prompt_tokens=502 charged_prompt_tokens=355`.
- Pearl config: no `verified_model_settlement_mode` set → prod runs the default `observe` mode, where the gateway's own stream-parsed usage is authoritative for billing. So this affects **billing**, not just buyer-visible usage.

## Fix (A)

On over-cap, `usageFromJSON` now **clamps** the reported usage to the reservation cap (reduce prompt first, then completion; floor at 0; clamp cached ≤ prompt) and returns it as valid — mirroring the coordinator's existing "charge bounded prompt tokens" behavior. Effect:
- The buyer keeps a `provider_reported` usage frame, forwarded with the provider's **actual** counts (only `cached_prompt_tokens` is normalized).
- Billing settles `provider_reported`, bounded to the reservation — **never over the hold** (over-billing defense preserved).
- Provider self-consistency (`total_tokens` mismatch) is still validated on the original values before any clamp; malformed / negative / overflow / completion-over-max still hard-reject.

## Deferred (B)

Widening the cap heuristic was intentionally **not** included: the validation cap equals the reservation hold by invariant, so widening it widens every request's quota hold (it 429'd a demo request in tests). A alone fully fixes the bug; token-dense prompts simply bill at the bounded cap (buyer-favorable, same as the coordinator already does). Any cap widening is a separate quota-behavior decision.

## Tests

- `usageFromJSON` clamp unit coverage; the Air5 abusive-usage case now asserts clamping (bounded), not rejection.
- New streaming regression `TestStreamingOverCapUsageClampsInsteadOfDroppingFrame` — usage frame forwarded + settled `provider_reported`.
- New non-streaming regression `TestNonStreamingOverCapUsageClampsInsteadOfRejecting`.
- Full gateway suite + `go vet` green.

## Audit

Three-lane codex audit (code / security / architect), money-path, **PASS 0 CRITICAL / 0 HIGH / 0 MEDIUM on the first round**. Security lane specifically verified a malicious provider cannot use the clamp to charge beyond the reservation hold.

### DIFF
--- a/phase5-gateway/internal/router/chat_proxy.go
+++ b/phase5-gateway/internal/router/chat_proxy.go
@@ -2255,11 +2255,37 @@ func usageFromJSON(body []byte, maxUsageTokens, maxCompletion int64, allowComple
 		return tokenUsage{}, true, fmt.Errorf("usage completion_tokens exceeds request max_tokens")
 	}
 	sum := usage.PromptTokens + usage.CompletionTokens
-	if sum > maxUsageTokens {
-		return tokenUsage{}, true, fmt.Errorf("usage token total exceeds request maximum")
-	}
+	// Validate provider self-consistency on the ORIGINAL reported values before
+	// any clamp.
 	if rawUsage.TotalTokens != nil && usage.TotalTokens != sum {
 		return tokenUsage{}, true, fmt.Errorf("usage total_tokens does not match prompt_tokens plus completion_tokens")
+	}
+	if sum > maxUsageTokens {
+		// The provider's honest token count exceeded the reservation-derived
+		// validation cap — common for token-dense prompts the byte heuristic
+		// under-counts. Clamp to the cap (reduce prompt tokens first, then
+		// completion) instead of rejecting: this mirrors the coordinator's
+		// "charge bounded prompt tokens" behavior so the buyer keeps a
+		// provider_reported usage frame and bounded billing, instead of losing
+		// usage visibility and being silently billed on gateway_estimated byte
+		// estimates. Never charges beyond the buyer's reservation hold. The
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
approved

PR #512, opened 2026-07-09
feat(opoi): W3 canary cold-start grace for the TTFT/TPS wall-time gates

Augustas11:
### Problem
A cold 30B model load produces ~8s TTFT on the first canary probe after a
provider (re)connects. That false-fails a static `max_ttft_ms` — the reason the
live Pearl tune had been forced to pad `max_ttft_ms` from 3500 up to 7000. The
padded value weakens the steady-state latency gate for every subsequent probe
just to survive one cold probe.

### Fix
New `pool.canary_cold_start_grace_s` (default `0` = off). For that many seconds
after a provider connects, the **wall-time** latency gates (`max_ttft_ms` **and**
`min_sustained_tps` — canary probes are non-streaming, so both are measured over
wall clock and are cold-contaminated) are **waived**. This lets `max_ttft_ms`
stay at the real production target (3500) instead of a padded 7000. Staging
overlay sets `300`.

The **nonce-correctness gate is never relaxed**, so anti-downgrade is unchanged.

### Sanction-safety model (hardened across 4 audit rounds)
The grace is on a security-sensitive sanction gate, so the model is deliberately
conservative:

- **Correctness first**: a wrong-nonce answer is recorded as a failure even
  inside the grace window — grace only ever waives *latency*, never correctness.
- **Graced probe is NEUTRAL**: a correct-but-slow graced probe is not recorded —
  it neither counts as a failure nor **clears** failures accrued on enforced
  probes.
- **Graced probe FORCES the next probe enforced**: `enforceNextCanary` is keyed
  by `ProviderID` and **persists across reconnect**; it is cleared only after a
  *current, recorded* enforced probe. So a chronically-slow provider cannot
  arrange (via reconnect / due-timing churn) to only ever be probed under grace —
  the interleaved enforced probes still sanction it.
- **Grace is connect-window bounded**: `canaryColdStartActive` returns false once
  `ConnectedAt` age ≥ window, on zero/empty `ConnectedAt`/`ProviderID`, or when
  `enforceNextCanary` is set. A provider connected longer than the window gets no
  grace at all.

A waived probe logs `provider canary TTFT gate waived during cold-start grace
window` with `canary_ttft_ms` + `connected_age`.

### Audit convergence (three-lane codex, bar 0 C/H/M)
| Round | Lane | Finding | Resolution |
|---|---|---|---|
| R1 | security | reconnect-churn resets grace; HTTP TPS gap | provider-ID anchoring |
| R2 | security | graced pass clears failure counter; false decode-window flag | graced-neutral; honest non-streaming metrics |
| R3 | security/architect | all-graced via due-timing; wrong-nonce neutralized | enforced-probe alternation; correctness-gated neutrality |
| R4 | security/architect | flag cleared before `Current=true` | clear enforce-next only after a current recorded probe |
| R5 | architect | — | **LANE PASS 0/0/0** |
| R5 | security | reconnect-mid-enforced-probe drops result | see below → R6 |
| R6 | security (attribution) | — | **PRE-EXISTING BASELINE**, not introduced/worsened by this diff |

### Carried finding (pre-existing baseline — NOT introduced by this change)
R5 security raised: a provider can reconnect mid-enforced-probe so
`RecordCanaryResult` returns `Current=false`, dropping the result without
recording a failure. R6 attribution audit confirmed this is **pre-existing
baseline** behavior of the canary session-currency model:
`origin/main` already contains the identical `if !result.Current { return false }`
drop (server.go 2218-2221) and the non-current return in `RecordCanaryResult`
(provider.go 1027-1029). This diff does **not** make the evasion easier, cheaper,
or more powerful — grace is connect-window bounded, wrong nonce still fails, and
`enforceNextCanary` survives reconnect so the forced enforced probe cannot be
re-armed into grace. A future subsystem-wide hardening (strike-on-stale-disconnect
during an in-flight enforced probe) would need careful false-positive handling
(honest Mac-sleep / WiFi-blip mid-probe disconnects) and is out of scope for W3.

### Config
```yaml
pool:
  canary_cold_start_grace_s: 300  # 0 = off (default)
```
Validation rejects negative values. Examples updated:
`coordinator.yaml.example`, `dist/coordinator.yaml.example`,
`coordinator.opoi-v0-staging.yaml` (staging → 300), runbook §5.

### Tests
- `canary_probe_test.go`: `evaluateCanaryProbe` grace cases;
  `TestCanaryColdStartActiveIsChurnSafe` (reconnect-alternation model).
- `config_test.go`: `TestCanaryColdStartGraceValidation`.
- `go build ./...`, `go vet`, `go test ./internal/ws ./internal/config` green.

### DIFF
--- a/ops/runbooks/proof-of-weights-implementation.md
+++ b/ops/runbooks/proof-of-weights-implementation.md
@@ -81,6 +81,16 @@ pool:
 
 Full downgrade cheat smoke (30B claim + 8B serve) requires a dedicated lab provider loading the wrong weights; latency gates exercise the same sanction path.
 
+**Cold-start grace (2026-07-09):** a cold 30B load produces ~8s TTFT on
+reconnect, which false-fails a static `max_ttft_ms` (the reason the live tune
+had bumped 3500→7000). The real fix is `pool.canary_cold_start_grace_s` (default
+0): for that many seconds after a provider connects, ONLY the `max_ttft_ms` gate
+is waived — nonce-correctness and `min_sustained_tps` stay enforced, so the
+anti-cheat guarantee is unchanged. Staging overlay sets `300`, so `max_ttft_ms`
+can stay at the real production target (3500) instead of a padded value. A waived
+probe logs `provider canary TTFT gate waived during cold-start grace window` with
+`canary_ttft_ms` + `connected_age` so a genuinely-slow provider stays visible.
+
 ---
 
 ## §6 — W4 telemetry drift alerts
--- a/ops/runbooks/proof-of-weights-implementation.md
+++ b/ops/runbooks/proof-of-weights-implementation.md
@@ -83,13 +83,22 @@ Full downgrade cheat smoke (30B claim + 8B serve) requires a dedicated lab provi
 
 **Cold-start grace (2026-07-09):** a cold 30B load produces ~8s TTFT on
 reconnect, which false-fails a static `max_ttft_ms` (the reason the live tune
-had bumped 3500→7000). The real fix is `pool.canary_cold_start_grace_s` (default
-0): for that many seconds after a provider connects, ONLY the `max_ttft_ms` gate
-is waived — nonce-correctness and `min_sustained_tps` stay enforced, so the
-anti-cheat guarantee is unchanged. Staging overlay sets `300`, so `max_ttft_ms`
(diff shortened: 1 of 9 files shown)


antfleet-ops (approved):
LGTM — W3 canary cold-start grace; R1–R6 audit converged, CI green 13/13.

PR #513, opened 2026-07-09
fix(opoi): observe-only canary latency gates + fail-reason logging

Augustas11:
### Why
The coordinator pool canary sends `stream:false` (non-streaming). Its wall-time
latency metrics are therefore **structurally unreliable** — measured
`canary_ttft_ms` swings ~125ms↔7000ms and `canary_sustained_tps` ~7↔27000 for the
**same healthy provider**, depending on relay chunk timing. The streaming buyer
path shows the true numbers (TTFT ~1200ms, ~30 TPS); the canary does not.

This bit us in prod on 2026-07-09: tightening Pearl `max_ttft_ms` to 3500 (to
match the buyer-probe TTFT) made a warm provider fail the latency gate 3-in-a-row
→ sanctioned → `pool_ready:0` → intermittent buyer 503s. Mitigated live by
reverting to 7000. This PR is the durable fix.

### What
- **`pool.canary_latency_enforcement`** (default **`observe`**): a nonce-correct
  probe that breaches a wall-time latency gate (`max_ttft_ms` / `min_sustained_tps`)
  is **logged** (`provider canary latency breach observed (not enforced)`) but
  **not failed**. `enforce` preserves the exact prior W3 cold-start-grace
  behavior. The **nonce-correctness gate is ALWAYS enforced** in both modes, so
  anti-downgrade is unchanged. TPS is also watched observe-only via
  `proof_of_weights.telemetry_drift`.
- **`canary_fail_reason`** on the `provider canary failed` log
  (`nonce_mismatch` | `ttft_breach` | `tps_breach` | `incomplete` | `relay_error`)
  — prod canary failures were previously undiagnosable (no reason field).
- Grace (`canaryColdStartActive` / `LatencyGraced`) is evaluated **only** in
  enforce mode.
- Config examples / staging overlay / runbook §5 updated: default observe, keep
  `max_ttft_ms >= 7000` for the non-streaming round-trip, and **startup-only**
  (SIGHUP reloads tier2/billing, not pool — a restart is required to flip).

### Security invariants preserved (enforce mode)
Nonce wrong → always fail; graced probe neutral for the sanction counter AND
forces the next probe enforced; enforce-next flag cleared only after a current
recorded probe; reconnect/churn cannot arrange all-graced. Observe mode only
relaxes the (unreliable) latency sanction — never nonce.

### Audit (three-lane codex, bar 0 C/H/M)
| Round | Code | Security | Architect |
|---|---|---|---|
| R1 | **PASS 0/0/0** | 2 MEDIUM (SIGHUP-reload doc gap; stale runbook smoke) | **PASS 0/0/0** |
| R2 | — (unchanged) | **PASS 0/0/0** (both fixed, docs-only) | — (unchanged) |

### Tests
`TestEvaluateCanaryProbeLatencyGates` (enforce + observe + all fail reasons),
`TestCanaryLatencyEnforcementValidation`. `go build ./...`, `go vet`,
`go test ./internal/ws ./internal/config` green.

### Follow-up for the operator
This ships `observe` as the default, so once deployed the Pearl latency gates
stop sanctioning (nonce still enforces). At that point `min_sustained_tps` can
stay configured for the observe signal without ban risk. Flipping to `enforce`
later requires validating metric stability first (and a restart).

### DIFF
--- a/ops/runbooks/proof-of-weights-implementation.md
+++ b/ops/runbooks/proof-of-weights-implementation.md
@@ -68,8 +68,9 @@ pool:
     qwen3-coder-30b-a3b-instruct:
       - prompt: What is the code {nonce}? Reply with only the code.
         expected: '{nonce}'
-        max_ttft_ms: 3500
+        max_ttft_ms: 7000        # non-streaming round-trip; keep generous
         min_sustained_tps: 20
+  canary_latency_enforcement: observe   # observe (default) | enforce
 ```
 
 **Export:** `model_class_opoi_pass` on `/poolz` when a model-class bank was used.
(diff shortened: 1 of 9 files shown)


antfleet-ops (approved):
LGTM — observe-only canary latency gates + fail-reason logging; 3-lane audit converged 0/0/0, CI green 13/13.

PR #515, opened 2026-07-09
fix(malibu): show coordinator connection status, not just local ready

Augustas11:
## Summary
- Malibu no longer shows **Serving** from local `/v1/health` alone when `macprovider-cli` reports `coordinator.connected=false`.
- Wire existing `/v1/status` coordinator session data into `AgentSnapshot` and reconcile UI state from both local readiness and coordinator connectivity.
- When the model is loaded locally but the coordinator session is down, the menu bar shows **Sync** and the dashboard shows **Local only** with explicit copy (`Model loaded locally · not connected to coordinator`).

## Test plan
- [x] `xcodebuild -project Malibu.xcodeproj -scheme Malibu test -only-testing:MalibuTests` (123 tests, 0 failures)
- [ ] Manual: provider with local `ready` but coordinator auth-reject loop shows **Local only / Sync**, not **Serving**
- [ ] Manual: fully connected provider still shows **Serving** and earnings subtitle

### DIFF
--- a/phase3-binary/app/Sources/Malibu/Agent/AgentSnapshot.swift
+++ b/phase3-binary/app/Sources/Malibu/Agent/AgentSnapshot.swift
@@ -60,6 +60,9 @@ struct AgentSnapshot: Equatable {
     var lastError: String?
 
     var cliVersion: String?
+    /// Whether macprovider-cli reports an active coordinator WebSocket session.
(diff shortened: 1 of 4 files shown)


antfleet-ops (approved):
Coordinator-aware Serving state looks correct; local ready without coordinator session now surfaces Local only / Sync instead of misleading Serving.

PR #527, opened 2026-07-10
Recover providers wedged by rejected proofs

### LINKED ISSUE #519: Provider CLI wedges on rapid coordinator restarts (malformed v2 proof auth, silent pool drop)
## Summary
Rapid coordinator restarts within a short window wedge the `macprovider-cli` provider so its **SPEC-001 v1.6 §6.7 v2 proof-stage auth frame** comes out malformed, and the coordinator rejects every reconnect. The provider drops out of the pool (`pool_size:0`) and buyers get `no_provider_available` — while the Malibu app still shows "Serving" (local state only).

## Evidence (2026-07-09, prod)
Coordinator restarted 4× in ~1h during a deploy/incident. After the churn, `macprovider-cli` (a long-running process that never *exited*, so launchd KeepAlive never replaced it) began emitting proof frames the coordinator rejected:
```
provider auth_request proof rejected   bad_field:"type"   -> close 4001 invalid_auth_request: type
provider first auth message rejected   bad_field:"missing version" -> close 4000
```
Rejection is at frame-parse level (`ParseAuthRequest`, `phase4-coordinator/internal/ws/messages.go:465`), *before* the policy/exemption check — the same coordinator binary had accepted this provider minutes earlier, and its migration signature exemption was still valid. So the wedge is in the CLI's proof/ControlSocket machinery, not the coordinator or auth policy.

## Impact
Full 30B outage until a human noticed and hand-restarted the CLI. Restarting the *Malibu app* did **not** fix it (the app only adopts the running CLI — see companion issue). Only `pkill` + a fresh CLI process recovered it.

## Fix direction
`macprovider-cli` (phase3-binary) should either (a) detect a persistently-failing v2 proof handshake and self-restart the proof/ControlSocket subsystem (or the process), or (b) exit non-zero on repeated proof-frame build failures so launchd KeepAlive replaces it. Today the process stays *alive but broken*, which defeats KeepAlive.

## Repro
Restart the coordinator ≥3× within a couple minutes while a provider is connected; observe the provider fail to re-establish the v2 proof handshake.

Augustas11:
## Summary
- Recover providers wedged by repeated pre-session `invalid_auth_request` / malformed auth protocol failures.
- Map coordinator close diagnostics for pre-accept Tier2 auth failures into fatal auth/protocol classifications.
- Reset the auth failure streak on successful sessions, drain/auth-upgrade reconnects, and transient reconnect failures.
- Sanitize coordinator-controlled diagnostic text before it reaches fatal watchdog logs.

## Root cause
A provider that repeatedly produced an invalid proof before coordinator session acceptance could stay alive and back off forever. launchd only restarts the process when the provider exits, so the provider-side reconnect loop needed a bounded fatal path for repeated pre-accept auth/protocol failures.

## Validation
- `swift test --filter CoordinatorClientTests` passed: 70 tests, 0 failures.
- `git diff --check` passed.
- 3-lane audit rerun to completion: code correctness, security/auth, and architecture all PASS with 0 Critical/High/Medium and 0 Low on the final diff.

Fixes #519

### DIFF
--- a/phase3-binary/Sources/macprovider-cli/CoordinatorClient.swift
+++ b/phase3-binary/Sources/macprovider-cli/CoordinatorClient.swift
@@ -9,6 +9,8 @@ protocol ProviderWebSocketTask: AnyObject, Sendable {
     func sendPing() async throws
     func receive() async throws -> URLSessionWebSocketTask.Message
     func cancel(with closeCode: URLSessionWebSocketTask.CloseCode, reason: Data?)
+    var closeCodeRawValueForDiagnostics: Int? { get }
+    var closeReasonTextForDiagnostics: String? { get }
 }
 
 // M1-1 follow-up (codex security audit 2026-06-11): refuse HTTP redirects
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
accept

PR #529, opened 2026-07-10
Enforce one provider runtime owner

### LINKED ISSUE #520: Two launchd managers double-spawn macprovider-cli (duplicate --provider-id, port :61919 collision)
## Summary
The provider install registers **two** launchd managers for a single `macprovider-cli` singleton, and they race to (re)spawn it — producing duplicate `--provider-id mac` processes that fight over `--port 61919` (only one can bind; the other idles broken).

## Evidence (2026-07-09, prod Mac)
```
$ launchctl list | grep -iE 'malibu|macprovider|streamvc'
-       0    live.streamvc.macprovider-watchdog
22654  -15   live.streamvc.macprovider
```
Both re-launch the CLI. During incident recovery, every `pkill` was immediately followed by a respawn, and we repeatedly ended up with 2–3 concurrent `macprovider-cli` processes, e.g.:
```
14090 ... macprovider-cli --provider-id mac ...   (nohup/manual)
82959 ... macprovider-cli --provider-id mac ...
13742 ... macprovider-cli --provider-id mac ...
```
Only one holds `:61919` and serves; the rest are harmless-but-wasteful can't-bind idlers. The coordinator saw a single stable session, so it's not (currently) causing serving churn — but it's fragile and confusing.

## Fix direction
One manager for the singleton. Either fold the watchdog's responsibility into the main job's `KeepAlive`, or make the watchdog *adopt-and-verify* (check for an existing healthy CLI on `:61919` / a live coordinator session before spawning) instead of unconditionally launching. A port-bind/coordinator-session guard in `macprovider-cli` startup (refuse to start if another instance for the same `--provider-id` is already serving) would also prevent the duplicate class.

## Related
Companion to the Malibu-app adopt-mode gap (a *third* spawner) and the CLI-wedge issue.

antfleet-ops:
## Summary

Fixes #520 by making the main `live.streamvc.macprovider` LaunchAgent the sole routine runtime manager for the local provider singleton.

- Add a port-scoped `ProviderServeLock` so duplicate `serve` startups fail before model artifact preflight or model load.
- Add an early `127.0.0.1:<port>` listener probe after the lock so legacy or foreign listeners also fail before heavy startup work.
- Change the companion watchdog normal liveness path to observe and log only; bounded rollback recovery still bootstraps/kickstarts after restoring the prior binary.
- Update standalone and inlined installer watchdog copies, docs, and tests.

## Root Cause

The install flow could leave two launchd-side managers influencing one `macprovider-cli` singleton. Both could try to run the provider on the same local port, causing duplicate `--provider-id mac` processes to fight over `--port 61919`.

## Validation

- `swift test --filter ServeCommandTests` - 29 tests, 0 failures
- `swift test --filter ProviderServeLockTests` - 5 tests, 0 failures
- `swift test --filter ProviderConflictDetectorTests` - 8 tests, 1 expected env-gated skip, 0 failures
- `bash phase3-binary/dist/test/watchdog_health_scope.test.sh`
- `bash phase3-binary/dist/test/watchdog_rollback_paths.test.sh`
- `bash scripts/test-watchdog-inline-drift.sh`
- `bash scripts/test-install-launchd-enable.sh`
- `git diff --check`
- Three audit lanes re-run after fixes: implementation, architecture, and verification all reported 0 Critical / High / Medium / Low findings.

## Not Run

- `MACPROVIDER_INTEGRATION_TEST=1` real `launchctl` integration test.

### DIFF
--- a/ops/macprovider-watchdog/README.md
+++ b/ops/macprovider-watchdog/README.md
@@ -1,18 +1,26 @@
 # macprovider-watchdog
 
-External LaunchAgent that verifies the installed Mac provider process
-is alive and locally healthy. It ships alongside every install of
-`macprovider-cli` via the public `get.streamvc.live/install.sh` flow.
+External LaunchAgent that observes the installed Mac provider process
+and runs auto-update rollback recovery. It ships alongside every
+install of `macprovider-cli` via the public
+`get.streamvc.live/install.sh` flow.
 
 ## What it does
 
 Every 60 seconds, `watchdog.sh` checks that exactly one installed
 `macprovider-cli` process is running and that the provider's local
-`/v1/health` endpoint responds successfully. If either check fails,
-it issues `launchctl kickstart -k gui/$UID/live.streamvc.macprovider`
-so launchd restarts the provider. Coordinator TCP state is logged as
-advisory only; another process reaching the coordinator is not treated
-as proof that the provider is healthy.
+`/v1/health` endpoint responds successfully. If either check fails, it
+logs the condition and leaves restart ownership to the main
+`live.streamvc.macprovider` LaunchAgent's `KeepAlive` policy. This is
+intentional: #520 made the companion watchdog observer-only so there is
+one routine runtime manager for the provider singleton. Coordinator TCP
+state is logged as advisory only; another process reaching the
+coordinator is not treated as proof that the provider is healthy.
+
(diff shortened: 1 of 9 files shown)


antfleet-ops:
Closing and recreating under Augustas11 so antfleet-ops can provide the required independent approval.

PR #526, opened 2026-07-10
test(e2e): P2 cold/warm TTFT matrix harness — buyer-stream vs canary-nonstream

Augustas11:
## P2 — cold/warm TTFT matrix harness

Second e2e harness from the 2026-07-09 e2e-testing review (spec `RESEARCH_234`). P1 (`test/e2e/canary-buyer/`) measures steady-state serving quality; **P2 attacks cold-start behavior and the latency gates that depend on it.**

### Forcing function
A cold 30B model load produced a **30,827 ms** canary TTFT in prod. The W3 `max_ttft_ms` gate was hand-guessed (3500→7000) with no measured basis; tightening it to 3500 — calibrated off the *streaming buyer* ~1200 ms TTFT — banned a healthy provider 3-in-a-row → buyer 503s. Fix shipped was observe-only gates (#513), a safety valve, not a calibration. This harness produces the real numbers.

### The subtlety it gets right
The pool canary is **`stream:false`**; its "TTFT" is the full non-streaming round-trip. The buyer path **streams** (true first-token latency). These are **not interchangeable**, and the `max_ttft_ms` gate is evaluated against the *canary non-streaming* regime. The harness measures **both regimes side by side**, per model class, in three states (`warm` / `cold` / `post_reboot`):

| Regime | `stream` | "TTFT" | Whose number |
|---|---|---|---|
| `buyer_stream` | true | first content token | the buyer SLO |
| `canary_nonstream` | false | full round-trip (mirrors coordinator `canaryMetricsFromTiming`) | the `max_ttft_ms` gate input |

Cold is inherently **one sample per cycle** (the first request warms the model), so samples accumulate into an append-only NDJSON store; `--build-matrix` aggregates → Prometheus textfile + JSON with **sufficiency-gated advisory calibration**.

### What's here
- `coldwarm-probe.mjs` — zero-dep probe; reuses P1's security helpers **verbatim** (redact, SSRF guards, atomic writes) so they share the same audit.
- `cold-cycle.sh` — operator cold-induction (idle / provider-CLI restart / reboot). **Refuses a prod base** and **never restarts the coordinator** (issue #519 / the hour-long 2026-07-09 outage).
- systemd `coldwarm-warm.{service,timer}` + `coldwarm-matrix.{service,timer}`, launchd plist, `run-coldwarm.sh`, `mock-gateway.mjs`.
- `README.md` + `CALIBRATION.md` (maps matrix output → the three P2 deliverables).

[...]

PR #533, opened 2026-07-10
fix(canary): pin temperature 0 so providers echo the nonce exactly

Augustas11:
## Problem

The OPoI canary asks each provider to **echo a random nonce exactly** — that's the anti-downgrade / model-identity proof. But the canary request body (`canary_probe.go`) set only `max_tokens` + `stream:false` and **never pinned `temperature`**. So providers sample the nonce-echo at their own default temperature, and a model that serves normal traffic perfectly hallucinates the random string:

```
Reply with exactly: CANARY-A1B2C3   ->  CANARY-A1By2C3
Reply with exactly: CANARY-DEADBEEF ->  CANARY-DEADBEOF\n\nWait, let me correct that...
```

→ `canary_fail_reason=nonce_mismatch` → trips the failure threshold → `canary_trip=degraded` → a **healthy** provider is benched and buyers get `503 no_provider_available` on a provider that is serving correctly.

## Evidence (live mac provider, qwen3-coder-30b-a3b-instruct, 4-bit, 2026-07-10)

| temperature | exact nonce echo |
|---|---|
| default (unset) | **0 / 8** |
| `0` | **6 / 6** |

At default temperature the provider flapped ~35% ready / 65% degraded purely from canary nonce hallucination, despite streaming real completions fine.

## Fix

Add `"temperature": 0` to the canary request body so decoding is greedy/deterministic and an honest provider reproduces the nonce exactly. The nonce-correctness gate is unchanged — this only removes the sampling noise that was failing honest providers. Test asserts the canary body pins `temperature: 0`.

## Scope / gates

- `phase4-coordinator/internal/ws/canary_probe.go` (+ test). `go vet` clean; full `internal/ws` suite passes.
- The canary is a security/anti-downgrade gate → this should go through the **three-lane codex audit** (code / security / architect) before merge, and the deploy is the usual gated coordinator rebuild + restart.

Surfaced by the P2 cold/warm TTFT harness carry (#526).

### DIFF
--- a/phase4-coordinator/internal/ws/canary_probe.go
+++ b/phase4-coordinator/internal/ws/canary_probe.go
@@ -107,6 +107,14 @@ func buildCanaryBodyFromRandomWithChallenge(modelID string, maxTokens int, chall
 		}},
 		"max_tokens": maxTokens,
 		"stream":     false,
+		// Pin greedy decoding. The canary challenge requires the provider to echo
+		// a random nonce EXACTLY (anti-downgrade identity proof); without this the
+		// provider samples at its own default temperature and a model that serves
+		// fine otherwise (e.g. a 4-bit qwen3-coder-30b) hallucinates the nonce
+		// ("CANARY-A1B2C3" -> "CANARY-A1By2C3" -> garbage), producing spurious
+		// nonce_mismatch failures that bench a healthy provider. Measured
+		// 2026-07-10: temperature 0 -> 6/6 exact echo; default temp -> 0/8.
+		"temperature": 0,
 	})
 	if err != nil {
 		return nil, "", config.CanaryChallengeConfig{}, err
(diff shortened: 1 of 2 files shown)


Augustas11:
## Three-lane codex audit: PASS (0 CRITICAL / 0 HIGH / 0 MEDIUM / 0 LOW)

Ran the three independent codex lanes (`omc ask codex`) per repo convention.

- **Code** — `"temperature":0` marshals to valid JSON; the MacProvider parser accepts the integer JSON number and converts to Double (ChatCompletionRequest.swift:514); coordinator validation accepts 0 in [0,2] (server.go:4470); both HTTP and WS canary paths forward the body; test detects omission + nonzero. 0 C/H/M/L.
- **Security** — does not weaken the anti-downgrade gate: nonce + expected are generated server-side (crypto/rand), the nonce gate is evaluated before latency/grace, and `temperature` reveals nothing new to the provider. The canary is behavioral evidence, not crypto identity proof — a stub could always echo regardless of temperature; greedy decoding only removes *honest*-model sampling failures. 0 C/H/M/L.
- **Architect** — correct layer (coordinator canary request); no bad interaction with model_class_challenges, cold-start grace, or the losslessness probe. 0 C/H/M/L.

Bar (0 C/H/M) met on all lanes. Remaining gates: CI + review, then the standard gated coordinator rebuild + restart deploy.

antfleet-ops (approved):
Canary temperature:0 fix. Three-lane codex audit passed 0 C/H/M (posted above). Coordinator checks green. The phase3-binary swift-test red is pre-existing on main (Swift-6 concurrency error in InferenceRelay.swift, untouched by this coordinator-only change). Approving.

PR #538, opened 2026-07-10
Restore providers after optional-attestation and canary failures

antfleet-ops:
## What changed

- Treat absent optional SE/MDA attestation as `not_required` for coordinator-driven provider updates.
- Add an authenticated, idempotent operator recovery endpoint for persisted admission rejections and canary sanctions.
- Fence stale/in-flight canary results during recovery and keep canary persistence wired when probes are disabled.
- Advertise provider binary 1.8.26.
- Align Pearl monitor state and credential permissions with its hardened systemd sandbox.

## Root cause

An exact-match production canary sanctioned the only provider after the model transformed the nonce. The sanction persisted into admission state, but production had no inverse operator action. Separately, the optional-attestation trust state made older clients unable to consume the update containing the fix.

The production verification also found that the monitor wrote state outside its systemd `StateDirectory` and could not read root-only `monitor.env` credentials.

## Impact

Operators can clear false sanctions without database surgery, recovered sessions must reconnect and pass warmup before routing, and optional unprovisioned attestation no longer deadlocks self-update. The hardened monitor can persist state and read optional credentials without broadening its filesystem sandbox.

## Validation

- `go test ./...`
- `go vet ./...`
- `go test -race ./internal/ws ./internal/pool -count=1`
- `swift test --filter AutoUpdateTests` (32 passed)
- `check_monitor_sandbox_test.sh`
- `check_deploy_config_test.sh` (53 passed)
- Production Pearl: coordinator hotfix healthy, provider 1.8.26 ready/routing-eligible, warmup passed
- Production gateway: non-streaming HTTP 200 `PONG`; streaming SSE completed with `[DONE]`
- Production monitor: one-shot completed successfully; `systemctl --failed` is empty
- Final code, security, and architecture audits: no critical/high/medium/low findings for the provider recovery changes

## Deployment note

The exact coordinator build from commit `64083ef` and monitor asset from commit `c5e8877` are already deployed to Pearl as incident hotfixes. Canary probing is disabled pending replacement of the brittle exact-match probe.


Augustas11 (approved):
Hotfix recovery path verified: all required CI checks passed, production provider connectivity was restored, and public inference smoke tests succeeded.

PR #539, opened 2026-07-10
chore(specs): evacuate audit + prompt exhaust from specs/ (P1+P2)

Augustas11:
## What (REORG_PLAN.md P1 + P2)

Evacuates review exhaust out of `specs/`. It held **1,571** `.md` but only **24 canonical specs** — the rest was audit/prompt output drowning the normative surface. This moves **1,469 files** out; `specs/` now holds **103** (24 canonical + ~79 impl/step/notes + odd work-item docs left for P3/P4).

**All 1,469 are 100%-similarity renames** — no content changes, git history preserved. Verified before moving that **nothing anywhere references the audit paths**, so no link rewrites were needed. `spec-index --check` stays green (canonical specs untouched).

## Moves

- **1,309 `*_audit.md`** →
  - `audits/spec-NNN/` when the filename or a *unique* in-file spec reference identifies a spec (e.g. spec-017 gets 169, spec-018 114, spec-019 74),
  - `audits/item-NNN/` for work-item-numbered bundles (082, 125, 126, 127),
  - `audits/misc/` for **528** work-item audit bundles (`ARC_3_*`, `BUILD_SPEC_RATE_CARD_*`, `ISS-266-*`, `M2_1E_*`) that don't map to a single spec — honestly unsorted rather than fake-bucketed.
- **160 `*_PROMPT.md` / `ABSORB_*`** → **`audits/_prompts/`**.
  - ⚠️ Deviation from the plan: it first said `.omc/prompts/`, but `.omc` is **gitignored** — moving there would have silently dropped all 160 from version control. Caught during staging; rerouted to a tracked dir and `REORG_PLAN.md` corrected.

## Not in this PR

P3 (normalize the odd `125_`/`82_ITEM3_`/`ARC_3_` names still in `specs/`) and P4 (backfill headers for SPEC 007/012/021/025-027 + fix the `SPEC-012-source.md` title-says-010 bug) — separate follow-up, they need per-file judgment.

## Review tip

`git log --stat` is huge but every line is a `rename … (100%)`. To sanity-check nothing changed content: `git show --stat --diff-filter=M HEAD` should be empty (no modifications, only renames).


antfleet-ops (approved):
Verified: pure renames, zero loss, blocking test path fixed + CI green

Issue #524, opened 2026-07-09
Wire canary-buyer heartbeat (active paging) + adopt deploy-gate operating mode


Augustas11:
## Summary
The canary buyer harness (#507) is deployed and running on Pearl (`canary-buyer.timer`, every 30 min) but currently only writes to journald + local artifacts — **it does not alert anyone**. Two things are needed to make it deliver: (1) wire a heartbeat so failures actively page, and (2) decide its operating mode.

## 1. Wire the BetterStack heartbeat (turns logging into paging)
The harness is built for a dead-man's-switch: `run-canary.sh` runs the probe with `--fail-on-degraded`, and on a healthy (exit-0) run it pings `CANARY_HEARTBEAT_URL`; a degraded/failed/missed run leaves the heartbeat stale so the upstream monitor alerts. No Prometheus required.

Steps:
1. Create a BetterStack **Heartbeat** monitor with an expected period ≥ 35 min (timer is 30 min + jitter).
2. Add its URL to `/etc/macprovider/canary-buyer.env` on Pearl (mode 0640 root:macprovider):
   ```
   CANARY_HEARTBEAT_URL=https://uptime.betterstack.com/api/v1/heartbeat/<token>
   ```
3. `systemctl start canary-buyer.service` once and confirm the monitor goes green.

Until this is done the harness is inert — it accumulates journald logs no one reads. Today's provider flap/outage is exactly what a paging harness would have caught automatically instead of by manual discovery.

## 2. Decide operating mode — recommendation: **deploy-gate + heartbeat**
The benchmark mission is already achieved (baselines measured, bugs #511/#513 fixed). The remaining value is regression detection, whose ROI depends on mode:
- **24/7 tripwire** — cheap (~24k tok/day) insurance, but with 1 provider / 1 model it re-measures a narrow, stable slice; value scales with pool growth.
- **Deploy-gate (recommended)** — run it before + after every Pearl/provider deploy and treat red as a rollback signal. This is the single highest-value use and directly prevents today's class of incident (a deploy silently breaking serving).

Recommendation: keep the 30-min timer as cheap insurance **once the heartbeat is wired**, AND adopt it as a deploy gate. Revisit full 24/7 emphasis when the pool grows beyond 1 provider/1 model (routing/failover/per-model SLO coverage).

## Alert thresholds (already documented, for whoever wires dashboards later)
From `test/e2e/canary-buyer/README.md` — current prod passes all comfortably:
[...]

PR #542, opened 2026-07-10
Make buyer canary authoritative for deploys and paging

Augustas11:
## What changed

- makes every request outcome, TTFT, decode TPS, cache reuse, and missing signal authoritative for strict canary success
- requires an external heartbeat in the production systemd unit and accepts only an explicit HTTPS 2xx delivery
- isolates buyer and heartbeat credentials with systemd credentials and a dynamic service identity
- permits one complete bounded retry for the known mandatory AEAD handoff window; partial-success attempts never count as healthy
- adds Node and shell regression coverage and records the operational decision

## Why

The 2026-07-10 incident showed that the legacy probe could exit successfully after partial buyer-visible failure and had no active dead-man paging path. This makes the same external buyer canary the deploy gate and paging authority without re-enabling the false-sanctioning internal challenge bank.

## Validation

- 8 Node probe policy tests
- heartbeat wrapper, retry, timeout, credential, redirect, and malformed-input shell tests
- Node syntax, bash syntax, shellcheck, diff check
- three audit lanes converged to 0 critical / 0 high / 0 medium
- rebased patch is byte-equivalent to the audited commit

## Operational gate

Production activation is intentionally not part of this PR. Issue #524 remains blocked until a Better Stack heartbeat and rotated buyer credential are available; the legacy environment file must be removed atomically during that migration. No production or release state was changed.


antfleet-ops (approved):
Reviewed the strict canary deploy-gate patch and its rebased range-diff. CODE, SECURITY, and ARCHITECTURE audits are 0 critical/high/medium; both full CI runs passed. Production activation remains separately blocked on Better Stack and credential rotation.

PR #548, opened 2026-07-11
fix(buyer): retryable=true on transient errors + HEAD /healthz (H2)

Augustas11:
## Summary

Closes audit finding **H2** (buyer error envelopes always stamped `retryable:false`) and the **healthz HEAD-405** latent bug.

Before this change, the coordinator's `spec018RetryableByCode` table listed only permanent/cap error codes, so every transient availability/timeout envelope fell through to Go's zero value (`false`). Buyers that honor `retryable` therefore refused to retry recoverable transient 502/503/504s (`no_provider_available`, `provider_timeout`, `provider_error`, `provider_disconnected`, `provider_failed`) — the exact class that occurs during canary-incomplete degrade windows and provider churn.

Separately, `/healthz` accepted only `GET` and returned `405` for `HEAD`, tripping `curl -I`, Kubernetes HTTP probes, and UptimeRobot.

### Retryability (finding H2)
- `phase4-coordinator/internal/buyer/server.go`: added the transient availability/upstream codes to `spec018RetryableByCode` as `retryable:true`, and pinned the permanent/client codes to `false` explicitly (documented so the lookup is auditable rather than relying on the zero value).

Final transient (`retryable:true`) set, each verified as a buyer-emitted code:
  - `no_provider_available` (503)
  - `provider_timeout` (504)
  - `provider_error` (502)
  - `provider_disconnected` (502) — **added this round**
  - `provider_failed` (502) — **added this round**

`provider_disconnected` and `provider_failed` are siblings of `provider_error`/`provider_timeout` — both are emitted to buyers via `writeError`/`writeSSEError` with explicit "buyer should retry" messaging, but were absent from the table and so wrongly reported `retryable:false`.

`upstream_provider_error` was checked and is **not** a buyer-visible error `code` — it exists only as an error *type* returned by `spec018ErrorType` (e.g. for `byte_cap_exceeded`, `malformed_tool_call_final_json`). No table entry is required for it.

### healthz HEAD (latent 405 bug)
- `phase4-coordinator/internal/buyer/server.go`: registered `r.Head("/healthz", s.handleHealthz)`.
- `phase4-coordinator/internal/ws/server.go`: `handleHealthz` now allows `HEAD` (and advertises it in `Allow`).
- `phase5-gateway/internal/router/server.go`: `handleHealthz` now allows `HEAD`.

The net/http server discards the response body for `HEAD`, so probes get `200` with correct headers and no body.

### SPEC-006 §5.2
- Added a tight additive normative note to the error-envelope section documenting the `retryable` field, its transient-vs-permanent classification, and the default-`false` rule for unclassified codes. No ACs renumbered.

### Tests
- `internal/buyer/server_internal_test.go`: `TestRetryableByCodeClassification` and `TestWriteErrorEnvelopeRetryableField` now cover `provider_disconnected` and `provider_failed`.
- `internal/buyer/server_test.go`: `TestHealthzAcceptsHEAD` drives a real `httptest.NewServer` (so the transport strips the HEAD body, matching production) and asserts `200` + empty body.
- `internal/ws/server_test.go`: `TestProviderHealthzAcceptsHEAD` covers the provider healthz HEAD path.

## Verification

```
phase4-coordinator: go build ./... -> OK
phase4-coordinator: go vet ./...   -> OK
phase4-coordinator: go test ./internal/buyer/... ./internal/ws/... -> ok
phase5-gateway:     go build ./... -> OK
phase5-gateway:     go test ./internal/router/... -> ok
```


antfleet-ops (changes requested):
antfleet-ops exact-SHA review of 2fc9766820d6a46cdef72008c70ab1cfad8f6c25: REQUEST CHANGES. Architecture/spec audit found C0/H2/M3/L1. Blocking: (1) SPEC-006 treats every X-RateLimit-Reset header as a retry signal even though daily quota reset headers remain on permanent downstream errors; narrow the normative rule. (2) the coordinator streaming writer discards SPEC-019 per-response retryable=false overrides, so deterministic empty structured-output failures can be advertised retryable=true. Also close legacy missing-field body normalization, completeness-guard drift, invalid_tools explicit classification, the response_byte_cap_exceeded cross-spec contradiction, and the normative JSON example before re-review. Security audit was C0/H0/M0/L0 and runtime tests are green, but passing tests do not resolve these contract defects. This PR is excluded from the v1.8.30 operational recovery release until corrected.

antfleet-ops (approved):
Buyer-error `retryable` contract + healthz HEAD. Codex 4-round audit discipline: r1 (fix not end-to-end) → r2 (same-class contradiction) → r3 (coordinator completeness false-guarantee HIGH) → r4 closing audit = 0 LIVE-WRONG HIGH/CRITICAL. All remaining HIGH/MEDIUM are pre-existing/carried or test-doc completeness (none ships a wrong runtime value); coordinator completeness list verified exhaustive; both modules green. Carried follow-ups documented (route_snapshot_failed no-charge, signup_rate_limited→false, streaming end.Retryable override, SPEC wording). Approving.

PR #553, opened 2026-07-11
Prebuild v1.8.30 before creating its immutable tag

Augustas11:
## Outcome

Makes the complete unsigned release build the acceptance gate before an immutable production tag exists, while keeping every protected signing and publication transition fail-closed on the exact tag and captured commit.

## Root cause

The v1.8.27-v1.8.29 attempts created immutable tags before the hosted release build had exercised toolchain parsing, architecture checks, and Swift package resolution. The v1.8.29 build then proved mlx-swift 0.31.6 requires Swift 6.3 while the reviewed Xcode 16.4 runner provides Swift 6.1.

## Changes

- adds an explicit false-by-default candidate dispatch mode available only to the secret-free contents-read build job;
- keeps protected signing strict on an existing exact-target tag and allows the captured tagged commit to remain an ancestor if main advances during review/notarization;
- binds all six unsigned inputs to repository, tag, commit, prerelease state, exact toolchain, and SHA-256 hashes before they cross into signing;
- requires exactly antfleet-ops as environment reviewer, with self-review and admin bypass disabled;
- pins mlx-swift 0.31.4 at dc43e62d7055353c7f99fa071a4e71d29dfddc44 for Xcode 16.4 / Swift 6.1 compatibility;
- advances CLI, Malibu build, ledger, and coordinator advertisement to v1.8.30/build30;
- makes package failures retain the useful final 200 log lines in CI;
- aligns release and Pearl updater runbooks with tag-late operation.

## Validation

- make test-dist
- swift test --parallel: 1,135 tests passed
- local package.sh v1.8.30 succeeded; binary reports 1.8.30 and packaged resources were inspected
- release tag/race, security posture, publication provenance, and version cohesion regressions passed
- shellcheck and git diff --check passed
- live production-release posture guard passed: sole antfleet-ops reviewer, self-review disabled, admin bypass disabled, main-only, immutable releases, protected v* tags

Exact-SHA audits for 1d47bf3e1c94fa2c43d9008bd7f46097c95e0f8f:

- code: C0/H0/M0/L0 — approve
- security: C0/H0/M0/L1 — pass
- architecture: C0/H0/M0/L1 — pass

The remaining lows are non-blocking hardening: lightweight exact-target tags remain accepted although the runbook uses a signed annotated tag, and two ad-hoc negative probes are not yet committed fixtures.

## Release boundary

This PR does not create v1.8.30, publish a release, or activate Pearl. After merge, dispatch candidate=true from fresh main, wait for the unsigned build to succeed, create signed annotated v1.8.30 at the captured SHA, then approve sign_publish as antfleet-ops.


antfleet-ops (approved):
Exact-SHA release audit accepted: code C0/H0/M0, security C0/H0/M0, architecture C0/H0/M0. Candidate build remains unprivileged; protected signing requires the exact tag and antfleet-ops environment approval.

PR #554, opened 2026-07-11
Keep draft release publication recoverable

Augustas11:
## Why

The v1.8.30 release proved two production gaps: GitHub REST lookup by tag does not return a draft release, and the Malibu publication verifier used a construct incompatible with macOS Bash 3.2. The release artifacts were valid, but the workflow stopped before public transition and required an exact numeric-release recovery.

## What changed

- discover drafts through `gh release view`, then bind the numeric release ID, tag, target commit, and draft state independently
- bind exact release title, reviewed notes, numeric asset IDs, and asset digests into the draft publication manifest
- re-fetch the immutable public release and require byte-identical draft/public manifests before Malibu publication
- make the Malibu publication-set verifier portable to Bash 3.2
- add mutation regressions for release identity, presentation, asset IDs, workflow ordering, and the portable verifier path

## Validation

- `make test-dist` passed at exact head `8c215ecb3b1a6baba45f8c2a493b23fca618ac71`
- targeted publication, security-posture, and Malibu tests passed under `/bin/bash` 3.2
- ShellCheck, Bash/Python syntax, workflow YAML parse, and `git diff --check` passed
- three independent audits: code C0/H0/M0/L0 APPROVE; security C0/H0/M0/L1 COMMENT; architecture C0/H0/M0/L1 APPROVE

## Residual risk

GitHub does not offer an atomic verified-draft GET-to-PATCH transition. A concurrent mutation can still burn the public release, but the immutable post-publication re-fetch and manifest-equality gate detect it before Malibu publication, limiting impact to availability/recovery rather than undetected downstream publication.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -995,6 +995,10 @@ jobs:
             --target "$release_commit" \
             --verify-tag \
             --draft
+          gh release view "$tag" \
+            --json databaseId,isDraft,tagName,targetCommitish > draft-release-cli.json
+          python3 scripts/verify-release-draft-identity.py cli \
+            draft-release-cli.json "$tag" "$release_commit" > draft-release-id.txt
 
       - name: Verify draft release assets by numeric ID
         env:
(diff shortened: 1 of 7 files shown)


antfleet-ops (approved):
Reviewed exact head 8c215ecb. Independent code, security, and architecture audits report C0/H0/M0; full test-dist and targeted release publication regressions pass.

PR #555, opened 2026-07-11
fix(coordinator): implement SPEC-010 R-3.3.4 seenModels union of supported_models (404->503 for declared-but-cold)

Augustas11:
## What & why

Implements **SPEC-010 v1.5 R-3.3.4** (audit finding: never implemented). R-3.3.4 requires the coordinator's `seenModels` index to be the **UNION** of each provider's served `ModelID` **and** every entry in its `SupportedModels[]`, so `ModelKnown()` returns `true` for a model that some provider **declares** supporting but is not currently serving (cold).

Before this change, only `ModelID` was recorded into the seen index at registration and heartbeat, so `SupportedModels` never fed `ModelKnown()`. A buyer request for a declared-but-cold model returned **404 `model_not_found`** instead of the spec'd **503 `no_provider_available`** (transient/retryable — a provider may warm it).

This is **buyer-visible**: the 404→503 substitution appears only when at least one provider explicitly opts in by sending `supported_models` with more than its `model_id` (R-3.3.4). Legacy providers carry `SupportedModels == [model_id]` (R-3.1.5 synthesis), so the union collapses to `{model_id}` and behavior is byte-identical (L-1 / R-3.5.1).

## The change

`phase4-coordinator/internal/pool/provider.go`
- New `recordSeenModelsUnionLocked(providerID, modelID, supportedModels)` records the served `model_id` and then each `SupportedModels[]` entry.
- Called at **both** seed sites:
  - **Registration** — `RegisterAtDetailed` (was `recordSeenModelLocked(p.ProviderID, p.ModelID)`).
  - **Heartbeat** — `applyHeartbeatLocked` (was `recordSeenModelLocked(p.ProviderID, hb.ModelID)`). Heartbeat frames don't carry `supported_models`, so `p.SupportedModels` (populated at registration) is the authoritative declared set.

Each `SupportedModels[]` entry flows through the existing `recordSeenModelLocked`, so it shares the served `model_id`'s **normalization** (`strings.ToLower` canonical key for the pool-lifetime accumulator; raw id for the per-session attribution set) — the same case-folding the registration path already applies.

## 404 vs 503

The buyer branch in `internal/buyer/server.go` keys the distinction entirely off `ModelKnown()`:

```
if !s.pool.ModelKnown(req.Model) && s.resolveModelClass(req.Model) == nil {
    // 404 model_not_found
}
```

- **Declared, cold** → `ModelKnown()` now true → 404 branch skipped → routing finds no candidate serving that `model_id` → **503 `no_provider_available`**.
- **Undeclared by any provider** → `ModelKnown()` still false → **404 `model_not_found`** (unchanged). The union does **not** turn genuinely-unknown models into 503.

No buyer-side code change was needed.

## Lifecycle / removal handling

Union entries age **exactly as `model_id`** because they reuse `recordSeenModelLocked`:
- **Per-session attribution** (`seenModelsByProvider`): dropped on disconnect (`RemoveIfSession`/`RemoveIfSessionState`) and on session-replacement — M2-5 / PERF-5. No leak into the per-session index.
- **Pool-lifetime accumulator** (`seenModelsLifetime`): append-only, retained for the coordinator process lifetime — SPEC-002 § 7.2 cold-start-race survival, identical to `model_id`. This is intentional (declared-cold stays 503 across the cold-start window), not a stale-forever bug.

No separate removal path is introduced.

## Tests

- `TestModelKnownUnionsDeclaredSupportedModels` (pool) — declared-but-cold model is `ModelKnown`-true (incl. case-folding); undeclared model stays false; per-session index carries then drops the entry on disconnect while the lifetime accumulator survives.
- `TestModelKnownUnionsSupportedModelsOnHeartbeat` (pool) — union re-applied on the heartbeat path.
- `TestChatCompletionsDeclaredButColdModelReturns503` (buyer) — 503 for the declared-cold model, 404 for an undeclared model.

All three fail against the pre-fix behavior (buyer test returned 404) and pass after.

```
go build ./... && go vet ./... && go test ./internal/pool/... ./internal/buyer/... ./internal/routing/...
ok  internal/pool
ok  internal/buyer
ok  internal/routing
ok  internal/routing/sticky
```
Full `go test ./...` for the module is green; `gofmt -l` clean.

## SPEC

SPEC-010 is **LOCKED v1.5**: no normative text changed. The status "known open gap" note is marked **CLOSED (2026-07-11)** and a non-normative R-3.3.4 implementation note added in §3.3.

---

## Round 2: codex 3-lane audit findings addressed (1 HIGH, 1 MEDIUM; security lane all-LOW/bounded, no blocker)

**HIGH — fixed.** `ModelKnown()`'s live-provider fallback scanned only each connected provider's current `ModelID`, never its declared `SupportedModels`. The seen-index union above (`recordSeenModelsUnionLocked`) is a best-effort accumulator bounded by `maxSeenModelsPerProvider` (32, per-session), `maxLifetimeContribPerProvider` (128, per-provider lifetime), and `maxSeenModelsLifetime` (4096, global lifetime) — a provider whose declared catalog exceeds those caps could have entries silently dropped from the seen index, so a declared model beyond the cap returned a **permanent 404** even though a live provider currently declares it. A served `model_id` never had this gap (the live `ModelID` scan always covers it); declared models needed the same unconditional guarantee. Fix: `internal/pool/provider.go`'s `ModelKnown()` gains a new fallback (1b) that scans every live provider's `SupportedModels` (same `EqualFold` normalization as the `ModelID` scan), independent of seen-index cap state. This is the correctness core of R-3.3.4, not optional hardening.

**MEDIUM — carried, not resolved in this PR.** Cross-spec inconsistency: SPEC-006 §17.2 defines the 404 condition as "not in any provider's served or recently seen model list" without an explicit reference to declared-but-cold models, and SPEC-002 v1.3.5 R-3.X.6 calls the identical `seenModels` union `MAY` (optional) while SPEC-010 v1.5 R-3.3.4 says `MUST`. The code correctly follows the more specific, LOCKED SPEC-010 `MUST`, and changes no dispatch outcome (R-3.4.1 / SPEC-002 R-3.X.6's "MUST NOT change dispatch outcomes" both still hold — only the buyer error code for declared-but-cold requests changes, 404→503). SPEC-002 and SPEC-006 are large LOCKED specs and are intentionally **not** amended by this PR. `specs/SPEC-010-model-catalog.md` §3.3 now carries an explicit follow-up note flagging this cross-reference reconciliation for a future pass (operator: track as a runbook follow-up item).

**LOW — test adequacy fixes, guarding the HIGH:**
- `TestModelKnownUnionsSupportedModelsOnHeartbeat` was ineffective: it declared the supported model at *registration* time, so the assertion stayed green even with the heartbeat-site union reverted. Rewritten to append the supported-model entry to the live `Provider` *after* registration (white-box, same package) and assert *after* disconnect — at that point the live-scan fallback (the HIGH fix above) no longer applies, so the assertion can only pass if the heartbeat call itself recorded the entry into the seen index.
- Added `TestModelKnownFindsDeclaredModelBeyondSeenIndexCaps` (pool) and `TestChatCompletionsDeclaredModelBeyondSeenIndexCapsReturns503` (buyer): a provider declaring 200 supported models, with an entry deliberately placed past both the 32-entry per-session cap and the 128-entry per-provider lifetime cap, asserting it is still `ModelKnown` and still returns 503.
- Buyer test now asserts `error.retryable == true` for the 503 case (and `false` for the 404 case) and that the response body is *exactly* the OpenAI error envelope shape (a single top-level `"error"` key) — ruling out any forwarded provider body.
- Fixed fixtures so the served `ModelID` is included in `SupportedModels`, per R-3.1.4 (the original PR's pool and buyer fixtures violated this).

Every new/changed assertion above was manually verified to **fail** when its corresponding fix is reverted (not just written as a positive check).

```
go build ./... && go vet ./... && go test ./internal/pool/... ./internal/buyer/... ./internal/routing/...
ok  internal/pool
ok  internal/buyer
ok  internal/routing
ok  internal/routing/sticky
```
Full `go test ./...` for the module remains green; `gofmt -l` clean.

### DIFF
--- a/phase4-coordinator/internal/buyer/server_test.go
+++ b/phase4-coordinator/internal/buyer/server_test.go
@@ -5556,6 +5556,56 @@ func TestChatCompletionsColdStartRaceReturnsNoProviderAvailable(t *testing.T) {
 	assertOpenAIErrorEnvelope(t, unseen, "model_not_found", "invalid_request_error")
 }
 
+// TestChatCompletionsDeclaredButColdModelReturns503 pins the buyer-
+// visible half of SPEC-010 v1.5 R-3.3.4: a request for a model that a
+// connected provider DECLARES supporting (supported_models) but is not
+// currently serving (cold) must return 503 no_provider_available
+// (transient/retryable — the provider may warm it), NOT 404
+// model_not_found. A model that no provider declares still returns 404.
+func TestChatCompletionsDeclaredButColdModelReturns503(t *testing.T) {
+	registry := pool.NewRegistry([]config.ProviderConfig{{ProviderID: "p1", EndpointURL: "http://p1.example"}})
+	// Provider serves model-served but declares support for
+	// model-declared-cold (it is warm-capable but not currently loaded).
+	registry.Register(&pool.Provider{
+		ProviderID:            "p1",
+		AssignedID:            "s1",
+		Hostname:              "p1.local",
+		ModelID:               "model-served",
+		SupportedModels:       []string{"model-declared-cold"},
+		ModelParamsB:          7,
+		RAMGB:                 16,
+		MaxContextTokens:      20000,
+		MaxConcurrency:        1,
+		SlotsFree:             1,
+		SlotsTotal:            1,
+		ThroughputTPSEstimate: 20,
+		EndpointURL:           "http://p1.example",
+		Tier:                  pool.TierPinned,
+		InferencePath:         pool.InferencePathHTTPForwarding,
+		State:                 pool.StateReady,
+		LastHeartbeatAt:       time.Now().UTC(),
+		ConnectedAt:           time.Now().UTC(),
+		BinaryVersion:         "0.1.0",
+	}, nil)
+	server := buyer.NewServer(registry, zerolog.Nop(), time.Unix(1716768000, 0))
+
+	// (b) Declared-but-cold model: no provider currently serves it, but
+	// it is in the seen-model union (R-3.3.4) → 503, not 404.
+	cold := postChat(t, server, []byte(`{"model":"model-declared-cold","messages":[{"role":"user","content":"hello"}]}`), nil)
+	if cold.Code != http.StatusServiceUnavailable {
+		t.Fatalf("declared-but-cold model status = %d, want 503; body=%s", cold.Code, cold.Body.String())
+	}
+	assertOpenAIErrorEnvelope(t, cold, "no_provider_available", "service_unavailable")
+
+	// (c) A model no provider declares stays 404 model_not_found — the
+	// union must not turn genuinely-unknown models into 503.
+	unknown := postChat(t, server, []byte(`{"model":"model-nobody-declares-9000","messages":[{"role":"user","content":"hi"}]}`), nil)
+	if unknown.Code != http.StatusNotFound {
+		t.Fatalf("undeclared model status = %d, want 404; body=%s", unknown.Code, unknown.Body.String())
+	}
+	assertOpenAIErrorEnvelope(t, unknown, "model_not_found", "invalid_request_error")
+}
+
 // assertOpenAIErrorEnvelope decodes the response body and verifies
 // the full OpenAI error envelope shape (error.code, error.type,
 // error.message non-empty, error.param is null). Used by the cold-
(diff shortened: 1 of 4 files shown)


antfleet-ops (approved):
SPEC-010 R-3.3.4 seenModels union (404→503 for declared-but-cold models). Codex 3-lane audit: r1 (1 HIGH cap-pressure 404 + 1 cross-spec MEDIUM) → r2 fix (ModelKnown live-scan of SupportedModels, independent of seen-index caps; load-bearing tests verified) → CODE+ARCHITECT re-audit PASS 0 C/H/M; SECURITY passed r1 (bounded, no billing/routing exploit). Cross-spec MUST/MAY inconsistency carried documented. Unknown models still 404 (test-verified). Approving.

Issue #536, opened 2026-07-10
Add a signed Pearl release updater for coordinator and gateway binaries


Augustas11:
## Context

Pearl currently requires a manual SSH/deploy-script rollout for coordinator and gateway binaries. During the current provider outage, this created a recovery dependency: the Mac provider was already running locally but could not self-update because the coordinator advertised a version that the provider rejected during trust eligibility, and the provider was later removed from the network pool by canary failures.

The GitHub release workflow currently publishes Mac/provider assets and signed checksums, but Pearl has no supported mechanism to fetch and install the latest coordinator/gateway Linux binaries.

## Proposed outcome

Implement a Pearl-local updater that can be triggered manually and/or by a guarded systemd timer/webhook. It should select the newest eligible GitHub Release, download the coordinator and gateway Linux amd64 artifacts, verify them, and perform a transactional rollout with health-gated rollback.

## Requirements

- Extend the release workflow to publish coordinator-linux-amd64, coordinator-cli-linux-amd64 if required for operations, and gateway-linux-amd64, plus a SHA-256 manifest, detached signature, and release metadata identifying the exact tag/version and supported artifact set.
- Pin the GitHub repository and a release-signing public key in Pearl configuration.
- Never execute downloaded content before validating the release tag as semver, rejecting revoked/downgrade versions, verifying the signed checksum manifest, and verifying each artifact checksum, architecture, and embedded --version.
- Stage artifacts in a root-owned temporary directory, preserve the current binaries/configuration, and install atomically.
- Drain or use the existing connected-provider guard before coordinator restart.
- Restart coordinator and gateway in a safe order, then verify systemd active state, local health endpoints, public TLS endpoints, advertised binary version, gateway/coordinator compatibility, provider pool recovery, and canary pass.
- Automatically restore the previous binary/configuration if any post-restart health gate fails.
- Use a lock, bounded retry/backoff, timeout, and an auditable journal/event record for every check, install, rollback, and skipped release.
- Keep the trigger disabled by default until an explicit production enablement flag is set; support a dry-run/plan mode.
- Add hermetic tests for signature failure, checksum mismatch, downgrade/revocation, partial download, restart failure, rollback, concurrent triggers, and provider-connected drain protection.
- Document the operator runbook and recovery procedure.

## Acceptance criteria

1. A single guarded command can plan and install an eligible release without manually copying binaries to Pearl.
2. A signed release with both Linux artifacts can be installed atomically.
[...]

PR #557, opened 2026-07-11
fix(updater): validate alert config against named service group

Augustas11:
## Outcome

Pearl updater preflight now validates the independent alert configuration against the explicitly named `macprovider` systemd/installer group, not the service user primary GID. Candidate validation remains isolated on `macprovider-updater-validate`.

## Regression coverage

- proves a divergent passwd primary GID cannot override the named group GID
- rejects missing or root service accounts/groups
- asserts the alert unit keeps `Group=macprovider`
- preserves exact root:macprovider 0750/0640 policy

## Verification

- 65 Pearl updater tests pass (1 root-only environment skip)
- `make test-dist` passes
- Python compile, Bash syntax, and `git diff --check` pass
- CODE audit: C0/H0/M0/L0
- SECURITY audit: C0/H0/M0/L0
- ARCHITECTURE audit: C0/H0/M0/L0

Operationally discovered while advancing the signed Pearl updater preflight from #536 toward the credential-gated #524 activation. This PR does not enable the updater or alter Pearl production state.

### DIFF
--- a/ops/pearl-updater/macprovider-pearl-update
+++ b/ops/pearl-updater/macprovider-pearl-update
@@ -12,6 +12,7 @@ import contextlib
 import dataclasses
 import errno
 import fcntl
+import grp
 import hashlib
 import http.client
 import json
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Reviewed as antfleet-ops: exact-head code, security, and architecture audits are C0/H0/M0/L0; named-group production binding and regression coverage verified.

PR #559, opened 2026-07-11
fix(updater): permit one-way bootstrap from clean legacy builds

Augustas11:
## Outcome

Allows Pearl to perform its first signed transactional updater rollout from the two clean pre-updater git-describe binaries, without seeding false durable provenance or replacing daemons manually.

## Safety contract

- Accepts only clean `vX.Y.Z-N-g<lowercase hex>` identities when durable updater state is absent.
- Uses the highest installed base tag as the downgrade floor.
- Requires a strictly newer signed, checksummed, revocation-gated release.
- Rejects dirty, ambiguous, commit-only, same-base, downgrade, and durable-state conflict cases.
- Keeps legacy identities permanently ineligible for already-current reconciliation.
- Corrects manual alert preflight to use the same named user and group as systemd.
- Makes the negative signature test deterministic.

## Verification

- Pearl updater: 69 tests passed; 1 environment-only root UID-drop test skipped.
- `make test-dist`: passed.
- Python compile, Bash syntax, and `git diff --check`: passed.
- Exact-head independent audits: CODE C0/H0/M0/L0; ARCHITECTURE C0/H0/M0/L0; SECURITY C0/H0/M0/L1.
- Security low residual is limited to generic clean legacy shape admission for root-owned trusted binaries; Pearl exact lineage was verified safe.

Operational follow-up for #536; enables completing the production updater rollout after #524 monitoring activation.

### DIFF
--- a/ops/pearl-updater/macprovider-pearl-update
+++ b/ops/pearl-updater/macprovider-pearl-update
@@ -73,6 +73,9 @@ CONTROL_ASSETS = (
     "checksums.txt.sig",
 )
 SEMVER_RE = re.compile(r"^(?:v)?(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)$")
+LEGACY_GIT_DESCRIBE_RE = re.compile(
+    r"^v?(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)-([1-9][0-9]*)-g([0-9a-f]{7,40})$"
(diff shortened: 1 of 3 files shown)


antfleet-ops (approved):
Independent operational review: exact head a9bcbbd audited code C0/H0/M0/L0, architecture C0/H0/M0/L0, security C0/H0/M0/L1. The medium runbook identity mismatch was fixed before this head; full updater and distribution validation passed. Approved for merge after all required checks are green.

PR #560, opened 2026-07-11
fix(updater): snapshot shared stats database once

Augustas11:
## What changed

- allow only the production stats-mirror database to alias the coordinator database
- snapshot and journal that physical database once
- continue rejecting gateway/coordinator, stats/gateway, payout, and other duplicate database paths
- add focused production-shaped and rejection regressions

## Why

Pearl correctly configures the read-only stats mirror against the coordinator SQLite database. The updater preflight treated the two logical consumers as distinct physical databases and aborted before creating a transaction journal or mutating services.

## Validation

- updater suite: 72 passed, 1 expected root-only skip
- make test-dist: passed
- exact-head code audit: C0/H0/M0/L0
- exact-head architecture audit: C0/H0/M0/L0
- exact-head security audit: C0/H0/M0/L1; low residual is limited to root-controlled lexical path aliases and does not affect the production literal path
- Python compilation, Bash syntax, and git diff checks passed

### DIFF
--- a/ops/pearl-updater/macprovider-pearl-update
+++ b/ops/pearl-updater/macprovider-pearl-update
@@ -1642,11 +1642,15 @@ class Updater:
             gateway = self._yaml_section_scalar(gateway_text, gateway_config, "storage", "db_path", required=True)
             assert gateway is not None
             stats_unit_inputs, stats_database = self.stats_mirror_runtime()
-            resolved = [
-                self._database_path(gateway, "gateway storage.db_path"),
-                self._database_path(storage, "coordinator storage.db_path"),
-                stats_database,
-            ]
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Independent exact-head review accepted. Code, architecture, and security gates are clear at be16cda2ba0d4a53ba555b92e53e50b077cc728b; C0/H0/M0.

antfleet-ops (approved):
Independent review refreshed after the required rebase. Exact head 1ae12932ffe83afa8007a17bec0aa8a3ca80d504 is patch-equivalent to the audited change; code, architecture, and security gates are C0/H0/M0.

PR #561, opened 2026-07-11
fix(updater): recover legacy health without advertised field

Augustas11:
## Outcome

Allows an interrupted updater rollback to prove the exact legacy coordinator identity when that older health schema genuinely omits `recommended_binary_version`.

## Safety boundaries

- The exception applies only to anchored legacy git-describe coordinator identities.
- A present `recommended_binary_version` must still match exactly; explicit `null` fails.
- Signed release candidates remain strict semantic-version identities.
- Local/public gateway identity checks remain exact.
- Full serving recovery still requires coordinator pool readiness, gateway readiness, public identity, and the external canary.

## Production root cause

The signed v1.8.30 candidate correctly failed because database migrations 15-17 had not yet been applied. Rollback restored v1.8.26, whose older health response omits the advertised-version field; the updater then could not finish its otherwise valid serving proof.

## Validation

- 74 updater tests passed; 1 environment-dependent root/UID-drop test skipped.
- `make test-dist` passed, including canary wrapper E2E.
- `git diff --check` passed.
- Exact-head code, architecture, and security audits: C0/H0/M0/L0.

### DIFF
--- a/ops/pearl-updater/macprovider-pearl-update
+++ b/ops/pearl-updater/macprovider-pearl-update
@@ -2471,11 +2471,17 @@ class Updater:
             advertised_version=self.config_update.next_version,
         )
 
-    def local_coordinator_identity_ready(self, identity: RuntimeIdentity, require_pool: bool) -> bool:
-        health = self.get_json("http://127.0.0.1:8444/healthz")
+    @staticmethod
+    def coordinator_identity_ready(health: dict[str, Any], identity: RuntimeIdentity) -> bool:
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Reviewed exact head 04447bf82e0d83a48a764c46703effaeb1f13fbf. Code, architecture, and security audits are C0/H0/M0/L0; the legacy omission is narrowly bounded and candidate/gateway checks remain strict.

PR #588, opened 2026-07-14
gateway: forward SPEC-018 AC-45 X-MacProvider-Streaming-Mode diagnostic (item 15)

Augustas11:
## Item 15 — forward the SPEC-018 AC-45 `X-MacProvider-Streaming-Mode` diagnostic

Resolves runbook item 15. Gateway + spec only; **no coordinator, billing, or
money-path code change**.

### The drift (verify-before-design)
SPEC-018 AC-45 requires every v0.2 tool-calling response to carry the buyer-visible
diagnostic header `X-MacProvider-Streaming-Mode` (`incremental` /
`buffered_kill_switch` / `buffered_provider_downgrade`); its fail condition includes
"the header is absent."

The coordinator **already fully implements** AC-45 — the operator kill switch
(`COORDINATOR_STREAMING_FORCE_BUFFERED=1`), the per-(buyer, provider) auto-downgrade
after 3 malformed streams within 5 min, and the 10-min recovery
(`phase4-coordinator/internal/buyer/streaming_downgrade.go`) — and **sets the header**
on the streaming HTTP 200 path (`buyer/server.go:3052,3404,3570,3893`). But the public
gateway's blanket `X-MacProvider-*` strip (`copyCleanHeadersWithReceipt` →
`isMacProviderHeader`) removed it before the buyer on `api.streamvc.live`, so AC-45's
"header absent" fail condition was **live in production**.

### The fix (Option 1 — gateway allowlist)
- **`phase5-gateway/internal/router/server.go`**: `copyCleanHeadersWithReceipt` now
  forwards `X-MacProvider-Streaming-Mode` before the blanket strip, but **only** when
  the value is byte-exactly one of AC-45's three canonical values
  (`buyerVisibleStreamingModeHeader`) — any other value is dropped, not normalized or
  forwarded verbatim (defense-in-depth against header-content injection past the
  strip). The passthrough is unconditional across the receipt-eligible and
  non-receipt copy paths, so it is future-proof for a later coordinator non-streaming
  emission.
- **SPEC-006 → v0.9.9**: added the header to the §5.4 response-pass-through allowlist
  (both the §5.4 normative list **and** its outbound-strip-summary duplicate — swept
  both sides), with the drop-non-canonical requirement and observation-only note.
- **SPEC-018 v0.2.4**: reconciled the "Known open gap" status note to "Resolved gap"
  and corrected its stale `§8.3` → `§5.4` reference. **Documentation-only** — AC-45's
  normative text is untouched, so this is not a lock amendment (SPEC-018 is LOCKED).
- Regenerated `specs/README.md`.

### Chosen over Option 2 (scope AC-45 to the coordinator surface)
Buyers hit the gateway (`api.streamvc.live`), so scoping AC-45 to the coordinator
would have left its buyer-visible promise structurally unmet.

### Tests
`TestStreamingModeHeaderForwarded` (all three canonical values pass through on the
streaming path; sibling `X-MacProvider-*` still stripped) and
`TestStreamingModeHeaderInvalidValueDropped` (table-driven: CRLF-injection,
whitespace-wrapped, case-variant, empty, unknown-token — all dropped). Full
`phase5-gateway/internal/router` package green.

### Documented residual (out of scope for the gateway-strip fix)
The coordinator currently emits the diagnostic only on the **streaming** success
path. Non-streaming AC-45 emission is a separate coordinator-side completeness item;
the gateway passthrough already covers it if/when the coordinator adds it.

### Audit
Codex 3-lane audit (code-fidelity / security-correctness / architect). Round 1:
security 0/0/0/0, architect 0/0/0/0, code 0/0/0/**1 LOW** (the trim-then-accept
normalization above). Fixed to strict byte-exact matching in round 2; code lane
re-audited to confirm. Security + architect held (not re-fired — a stricter
value-check cannot regress a passed lane). Final: all three lanes **0 C/H/M**.


antfleet-ops (approved):
Item 15 (SPEC-018 AC-45 streaming-mode gateway passthrough). Codex 3-lane audit converged 0 C/H/M (code R2, security, architect). Gateway+spec only; phase5-gateway CI green. The two CI reds are confirmed pre-existing and unrelated: (1) 'check' spec-lint rejects specs/RESEARCH_23x files committed to main in d03a75d; (2) phase4-coordinator AgeThresholdRekey date-threshold flake reproduced on unchanged base. Approving.

Issue #585, opened 2026-07-13
G3 epic: physical acceptance for Option 2 lifecycle + updater-cluster Partials (#610/#612/#616/#658)


Augustas11:
## Status — G3 / physical-acceptance epic

**Role:** Umbrella tracking issue for remaining **physical acceptance** of the Option 2 provider lifecycle and the updater-cluster Partials. This is **not** a request to redesign the lifecycle again.

| Gate | Meaning | Reality (2026-07-21) |
|---|---|---|
| G1 — implementation merged | Option 2 lifecycle + updater-cluster Partials on `main` | **Done** for the updater cluster (see below). Option 2 design/decision basis: Entries [139](https://github.com/Augustas11/macprovider/blob/main/beta/DECISION_CRITERIA.md)–[143](https://github.com/Augustas11/macprovider/blob/main/beta/DECISION_CRITERIA.md). |
| G2 — exact candidate accepted | One signed private acceptance candidate bound to an exact commit | **Signed (private only)** — v1.8.57 `@ea9093e4` + successor v1.8.58 `@8c4c57d8` (PATH re-proof). Not a public release. |
| G3 — physical rollout complete | Both Macs prove the matrix + soak on that candidate (or successor) | **Open — this issue's remaining work** |

Historical incident narrative (2026-07-13 redesign brief) is preserved in the comment thread. Treat it as background, not current ground truth: two legacy Macs were already working on the latest **public** CLI/Malibu before this private candidate. "Nothing works" is stale.

---

### Already shipped on `main` (updater cluster)

| Partial | Squash tip | What landed |
|---|---|---|
| [#616](https://github.com/Augustas11/macprovider/issues/616) PATH/set convergence | [`ec6f5f23`](https://github.com/Augustas11/macprovider/commit/ec6f5f23) via [#672](https://github.com/Augustas11/macprovider/pull/672); regular-file via [#678](https://github.com/Augustas11/macprovider/pull/678) | Stale PATH symlink + regular-file entrypoints converge to activated compatibility set |
| [#612](https://github.com/Augustas11/macprovider/issues/612) recommendation continuity | [`3bd54131`](https://github.com/Augustas11/macprovider/commit/3bd54131) via [#671](https://github.com/Augustas11/macprovider/pull/671) | CLI marketing-version bumps no longer invalidate recommendation evidence |
[...]

PR #595, opened 2026-07-14
Keep acceptance version checks reliable under pipefail

Augustas11:
## Outcome

Keeps the private Issue #585 acceptance build reliable under macOS `set -o pipefail` while preserving the embedded-version assertion for the coordinator and gateway binaries.

## Root cause

Acceptance run `29357448110` built the binaries successfully, but Xcode 16.4's `strings` reported `failed to flush output` after `grep -q` exited on the first match. With `pipefail`, that expected early pipe close failed the job before any protected environment or secret was reached.

## Change

- Materialize each binary's `strings` output before checking for `v1.8.33`.
- Add a workflow-contract regression that rejects `strings | grep -q` under pipefail and requires both version checks to remain present.
- Leave the protected signer, secret boundary, ref/SHA checks, signing, notarization, and no-publication behavior unchanged.

## Validation

- `bash -n scripts/test-acceptance-candidate-security.sh`
- `git diff --check`
- `actionlint .github/workflows/acceptance-candidate.yml` when available

Refs #585


antfleet-ops (approved):
antfleet-ops exact-head operational approval under the repository dual-account house workflow, authorized by the account owner. Reviewed 778c4726d0d87b39aa95467e6beb5b87c222a17f: the fix materializes macOS strings output before grep -q, preserving pipefail and both embedded-version assertions; the added contract test rejects recurrence. No protected signer, secret, environment, ref/SHA, signing, notarization, retention, or publication boundary changed. All 13 required checks are green. Approved for squash merge and acceptance redispatch.

PR #598, opened 2026-07-14
Keep dependency-free acceptance builds valid on macOS Bash

Augustas11:
## Outcome

Fixes the Issue #585 acceptance control workflow failure seen in run 29361066508 before the protected environment gate.

The macOS runner uses Bash 3.2, where expanding an initialized empty array under `set -u` fails. The xcodebuild argument array now always contains the `archive` action and conditionally appends the resolved-package flag.

## Validation

- acceptance security contract passed
- Bash 3.2 dependency-free and lockfile argument regressions passed
- workflow YAML parse passed
- shell syntax passed
- diff check passed
- read-only audit: 0 critical, 0 high, 0 medium

No candidate code, signer secret handling, publication behavior, or protected-environment policy changes.

### DIFF
--- a/.github/workflows/acceptance-candidate.yml
+++ b/.github/workflows/acceptance-candidate.yml
@@ -161,12 +161,12 @@ jobs:
           set -euo pipefail
           cd candidate/phase3-binary/app
           "$RUNNER_TEMP/xcodegen-2.45.4/xcodegen/bin/xcodegen" generate
-          package_resolution_args=()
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Approved exact head 73cff0a6bd0aa3156b177f75006381724f2b3458 after Bash 3.2 regression, workflow integrity, and trust-boundary review.

PR #599, opened 2026-07-14
Keep acceptance signing in an owned user keychain

Augustas11:
## Summary

- create the protected Developer ID keychain in the standard user keychain directory used by the proven release signer
- capture and restore the prior user default fail-closed
- delete only a keychain created by this invocation, using a run/attempt/random scoped name
- regression-test query failure, create collision/failure, and downstream restore-before-delete behavior

## Failure evidence

Acceptance run 29365725391 imported the Developer ID identity, then failed in `security set-key-partition-list` with `The specified item could not be found in the keychain`. The successful release signer uses a standard user/default keychain; the failed acceptance signer used an absolute keychain under `RUNNER_TEMP`.

## Validation

- ShellCheck and Bash syntax on all three changed files
- code, security, and architecture/operations audits: C0/H0/M0

Advances #585. This control PR publishes nothing and retains read-only workflow permissions.

### DIFF
--- a/scripts/acceptance-signing-keychain.sh
+++ b/scripts/acceptance-signing-keychain.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+# Main-owned lifecycle helpers for the transient Developer ID signing keychain.
+# This file is sourced by sign-acceptance-candidate.sh.
+
+acceptance_previous_default_keychain=""
+acceptance_keychain_created=0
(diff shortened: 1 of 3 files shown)


antfleet-ops (approved):
Approved exact control head 19234d3243bf2921da363379e0bdec0d48d85d2e after all required CI checks and code/security/architecture audits passed at C0/H0/M0.

PR #602, opened 2026-07-15
Make v1.8.35 activation durable and rollback-safe

Augustas11:
## Summary

- publish the v1.8.35 CLI/Malibu/coordinator compatibility boundary
- replace an executing provider binary with a fresh same-directory inode, then fsync the file and containing directory
- keep rollback armed unless the exact restored session proves assigned identity, buyer serving, provider-reported catalog evidence, and legacy-bridge admission
- preserve Keychain credentials, provider identity, config, model selection, and recovery custody; no wipe/re-enrollment is required

## Root cause

The immutable v1.8.34 installer could copy new signed bytes over the vnode already being executed. On macOS that can leave the refreshed process subject to an AMFI SIGKILL even though the resulting bytes validate. The previous emergency rollback path could also encounter the sparse v1.8.30 pool envelope, whose `ready` state is not authoritative session-bound serving proof.

## Deployment order

1. Preserve both currently ready v1.8.30 providers and their recovery assets.
2. Deploy the bridge-capable coordinator/backend while that capacity remains online.
3. Install the signed v1.8.35 compatibility set atomically on each Mac.
4. Require exact buyer-serving admission before committing each cutover or rollback.
5. Remove obsolete binaries, launch items, and the legacy Malibu credential copy only after the new pair is proven.

## Validation

- `make test-dist`
- atomic/AMFI replacement suite: 44/44
- version pinning suite: 61/61
- Pearl updater suite: 103 passed, 1 root/systemd integration skip
- targeted Swift handshake test
- Bash syntax, ShellCheck, and `git diff --check`
- three independent final audits: 0 critical, 0 high, 0 medium, 0 low

## Remaining operational gates

After merge: protected signing/notarization and immutable v1.8.35 publication, bridge-first Pearl rollout, physical two-provider upgrade/rollback/restart/login/reboot matrix, exact-fleet canary, and the required soak. Issue #585 should remain open until those gates complete.

Relates to #585.


antfleet-ops (approved):
Approved exact head 61ccb8e9a420ac3f38388bd9cee6f04036c87c65 after CI run 29389853676 and aggregate ci-required completed successfully, the full local distribution suite passed, and independent code, security, and architecture audits reported C0/H0/M0/L0. This approval covers the reviewed v1.8.35 durable-activation change; Issue #585's protected signing, bridge-first two-provider rollout, physical recovery matrix, canary, and soak gates remain tracked separately and require signed evidence.

PR #594, opened 2026-07-14
item 20: signup_rate_limited retryable:false + streaming SSE end.Retryable override symmetry

Augustas11:
## Runbook item 20 — two consistency fixes carried from the #548 r4 closing audit

Buyer-visible **contract** change (the `retryable` field on error envelopes). No settlement / money-amount change.

### (1) `signup_rate_limited` → `retryable:false` (gateway)

`signup_rate_limited` (`phase5-gateway/internal/router/oauth.go` `createSignupAccount` — a per-IP daily account-creation cap, `SignupAccountsPerIPPerDay` over a 24h window) was emitted `retryable:true` with **no** `Retry-After`/reset header, and sits **outside** the `/v1/chat/completions` 30-RPS account clamp. That is the exact shape a prior round-3 SECURITY-MEDIUM revert moved to `gatewayPermanentCodes` (`false`) for its three siblings `feedback_rate_limited` / `oauth_state_rate_limited` / `demo_session_rate_limited`: a headerless, unthrottled 429 marked `retryable:true` tells a conforming SDK to auto-retry with zero backoff — a DoS amplifier. `signup_rate_limited` was left `true` only because that revert never named it.

Moved from `gatewayRetryableByCode` to `gatewayPermanentCodes`. Retrying the identical signup within the 24h window cannot succeed anyway.

### (2) Coordinator streaming SSE honors the provider `end.Retryable` override (SPEC-019 §5)

A provider WS terminal frame `inference_response_end` carries an optional `Retryable *bool` that overrides the static default `retryable` for a structured-output terminal outcome. The **non-streaming** JSON writer `writeProviderStructuredOutputError` already honored it; the **streaming** SSE writer `writeSSEError` did not — the synthesized-terminal-error site (`forwardWSStreaming`, gated on `isSpec019ProviderDetailCode`) dropped `end.Retryable` and defaulted to `spec018Retryable(code)`. So the two transports could report a different `retryable` for the identical terminal outcome.

New `writeSSEErrorWithRetryable(w, message, code, retryableOverride *bool, requestID)` threads the override; `writeSSEError` delegates with `nil` (byte-identical for its ~14 other call sites). The synthesis site now passes `end.Retryable`.

### Tests
- Gateway: `signup_rate_limited` moved from the `retryable` list to the `permanent` list in the classification test; `TestGatewayErrorCodeCompleteness` still green (exactly-one-map invariant holds).
[...]

PR #603, opened 2026-07-15
item 21: AST-based future-code completeness guard for emitted error codes

Augustas11:
## Runbook item 21 — AST-based future-code completeness guard

**Test-only. No production code change.**

### The gap (carried from #548 r4)
Each module's retryable-completeness test guards only a **hand-curated inventory** (`coordinatorEmittedErrorCodes` / `gatewayEmittedErrorCodes`). A new emitted code omitted from *both* the classification map and that list still passes CI — and an unclassified code defaults to `retryable:false` via Go's map zero value (the original H2 class of bug). A prior round even found the coordinator inventory had itself gone stale, letting 3 codes ship unclassified.

### The guard (both modules)
`TestCoordinatorEmittedLiteralCodesAreClassified` / `TestGatewayEmittedLiteralCodesAreClassified` parse the package source at test time and assert every **string-literal** `code` is classified — so a new emitted literal code **fails CI merely by existing**, with no hand-curated list edit. Coverage:

- **Emitter call sites** — a registry of writer functions; each function's `code` argument index is **derived from its real signature** (the parameter named `code`), never hand-maintained, so index drift is structurally impossible.
- **Inline classified envelopes** — the classifier functions (`gatewayRetryable`/`spec018Retryable`) are registered, and a composite-literal pass checks any map with both a literal `"code"` field and a `"retryable"` sibling (e.g. `writeStructuredOutputTimeoutSSE`), so inline emissions and code/classifier decoupling are both caught.
- **Fail-closed registry** — every registered emitter that is declared must resolve a `code` param; an unknown name must be an explicit, self-expiring forward-reference exemption (only item 20's `writeSSEErrorWithRetryable`, which the guard forces removed once #594 declares it).
- **Staleness cross-check** — every discovered literal must also be in the hand-curated inventory, catching the inventory rotting.

**Variable-routed residual** (documented): codes reaching an emitter through a variable — validation-helper returns, `routeError.code`, tier-2 `blockReason`, `end.Status`, `concurrencyErrCode` — can't be resolved by static AST. The finite ones (`isSpec019ProviderDetailCode` switch cases, `concurrencyErrCode` assignments) are pinned **directly from source**, so new members are auto-covered.

Mechanism tests prove the extraction actually fires (synthetic new codes are extracted and shown unclassified; variable args ignored); the guard's firing was also verified end-to-end by injecting unclassified codes into real source.

### Audit
Codex 3-lane converged **0 C/H/M** over 5 rounds. Notable catches, each fixed at the root:
- **R1 (architect HIGH):** the first cut hand-indexed `writeSpec019PreflightError` wrong and silently skipped every preflight code → replaced hand-indices with **signature-derived** indices.
- **R2:** inline `writeStructuredOutputTimeoutSSE` bypassed the registry; derivation failed open on rename → classifier registration + strict resolution + source-derived pins.
- **R3–R5:** stale forward-ref exemption, one-directional pins, inline code/classifier decoupling, and mechanism-test rigor — all closed.

LOWs ship documented. SPEC-006 §5.2 (5)'s "unimplemented follow-up" wording is now superseded-for-literals by this guard; that one-line doc update is deferred to avoid a merge conflict with item 20 (#594), which is concurrently editing the same §5.2 line, and a version-race with its in-flight v0.9.11 — to be folded when #594 lands. No DECISION-log entry (test-infrastructure, per the item-19 precedent).


antfleet-ops (dismissed):
Approving after codex 3-lane audit convergence (0 CRITICAL / 0 HIGH / 0 MEDIUM across code-fidelity, security-correctness, architect lanes).

antfleet-ops (approved):
Re-approving after rebase onto main. The AST guard's own stale-exemption branch fired on rebase (writeSSEErrorWithRetryable now declared post-#594); removed the transitional allow-absent entry as the guard demands. Test-only; guard green. Codex 3-lane 0 C/H/M.

PR #558, opened 2026-07-11
docs(spec-020): amend provisional trust table to match shipped crypto-gated reality (G2/c)

Augustas11:
## What & why (Wave B decision gate G2, disposition (c))

SPEC-020's normative trust table declared `provisional → notify-only`, but the
**shipped code auto-applies coordinator-recommended updates to provisional
providers by default** (`acceptProvisional` defaults to `true` via
`AutoUpdateConfig.acceptProvisional` reading `autoUpdateAcceptProvisional != false`).
The G2 runtime investigation confirmed the spec was stale, **not** the code.

**Flipping the default was rejected** as the fix:
- The prod `mac` provider is `tier: provisional` (bearer-validated via
  `provider_token`) with **no opt-in** flag set.
- The whole fleet is provisional by design — the coordinator ships with an
  empty `providers: []` pin list, so every self-service `curl | bash` install
  is admitted at `tier: provisional`.
- Flipping the default to `false` would break prod auto-update (it ran the
  2026-07-10 incident-recovery update to CLI 1.8.21) and defeat SPEC-020's
  entire purpose: delivering signed fixes to self-service providers without
  operator SSH.

So this PR **amends the SPEC (and fixes stale code comments) to match the
shipped, crypto-gated reality**. There is **no behavior change**.

## Crypto-gating + T-3 rationale (why the code is safe as-is)

Binary replacement is independently crypto-gated by the pre-existing
`SelfUpdate` path: GitHub-host URL validation, ECDSA P-256 signature
verification of `checksums.txt`, SHA-256 tarball verification, tar
path-traversal rejection, new-binary `self-test`, drain, and single-step
rollback. Per **threat model T-3**, a hostile or compromised coordinator can
at most *accelerate* delivery of a **legitimately signed, newer GitHub
release** (subject to opt-out, drain, cooldown) — it cannot substitute
unsigned, non-GitHub, downgrade, or arbitrary code. A bearer-validated
provisional session (validated `provider_token` + successful encrypted leg +
satisfied Tier2 attestation where required) is a materially stronger trust
signal than raw first-contact TOFU.

## Changes (spec + comments only — no logic touched)

- **`specs/SPEC-020-provider-autoupdate.md`** (normative):
  - Split the single `provisional → notify-only` trust-table row by auth
    sub-state (see wording below).
  - Added a "Rationale for provisional eligibility" paragraph (fleet is 100%
    provisional; crypto-gating; T-3).
[...]

PR #605, opened 2026-07-15
item 23: propagate coordinator auth_state so the bearerless-duplicate notify-only floor is client-enforceable

Augustas11:
## Runbook item 23 — propagate coordinator `auth_state` (SPEC-020 tokenless race-loser residual)

**Security-sensitive** (provider auth / autoupdate trust). Two-module code + SPEC docs. No production behavior change for bearer-validated providers (prod `mac` unaffected).

### The gap (SPEC-020 "Known residual", from #558 audit)
The SPEC-020 trust table assigns a **bearerless-duplicate** race-loser session `notify-only`, but that was coordinator-side *intent* only. The accept ack carried neither `assigned_provider_token` nor any `auth_state`, so the CLI's `AutoUpdateTrustState.fromCoordinatorPayload` computed `tokenConfigured=false` → `bearerlessDuplicate=false` and could reach `.eligible` instead of notify-only.

### The fix
- **Coordinator** (`internal/ws/messages.go` / `server.go`): both accept acks (`HelloAck` v1, `AuthResponse` v2) now carry `auth_state` (the `pool.AuthState` verdict: `bearer_validated` / `self_minted` / `bearerless_duplicate`; `mint_failed`/rejects close the connection and never ride an ack), populated from the value `resolveProvisionalToken` already returns.
- **Binary** (`AutoUpdateTrustState.swift`): reads `auth_state` **authoritatively** for the bearerless floor, with **fail-closed** handling — a key that is absent uses the legacy heuristic (unchanged), but any *present* value that is not a recognized non-empty string (unknown string, empty, `null`/NSNull, number, bool, object) fails **closed** to notify-only. The signal is used only to hold a *more* restrictive floor; it never relaxes a verdict (`tokenValidated` stays heuristic).

### Ownership (following the `assigned_provider_token` precedent)
- **SPEC-003 v0.10.3** FR-C9.2a — coordinator emission policy (+ corrected the stale FR-C9.1 generic-mint-error clause to the shipped fail-closed behavior).
- **SPEC-001 v1.8.4** §6.5.1 — wire field shape/domain.
- **SPEC-020 v0.1.7** — autoupdate interpretation; residual → RESOLVED.

### Tests
Coordinator asserts the ack carries `auth_state` for v1 self_minted + bearer_validated **and** v2 bearer_validated admissions. Binary: 8 trust-state tests including the tokenless-residual closure, heuristic-false-positive suppression, legacy fallback, unknown-value and malformed-type (empty/NSNull/number/bool/object) fail-closed. Full ws suite + 52 Swift AutoUpdate tests pass.

### Audit
Codex 3-lane converged **0 C/H/M** over 5 rounds — a security-sensitive fail-closed surface: each round tightened the handling of an anomalous `auth_state` (unknown string → non-string/malformed → reserved value) and corrected doc/ownership drift the change exposed.


antfleet-ops (dismissed):
Approving after codex 3-lane audit convergence (0 CRITICAL / 0 HIGH / 0 MEDIUM across code-fidelity, security-correctness, architect lanes).

antfleet-ops (approved):
Re-approving after rebase onto main. SPEC-001 v1.8.4 / SPEC-003 v0.10.3 / SPEC-020 v0.1.7 changelog order preserved (all +1 over main, no collision), spec-index regenerated. Local: coordinator ws auth_state tests pass, swift AutoUpdateTests 52/0. Codex 3-lane 0 C/H/M.

Issue #613, opened 2026-07-16
P0: enforce signed physical lifecycle acceptance before any stable release


Augustas11:
## Summary

The repository protects source, signing, notarization, checksums, tag identity, immutable GitHub releases, and environment approval. It does **not** currently require machine-verifiable physical lifecycle acceptance before a stable release becomes public and installable.

Issue #585 demonstrated the consequence: extensive component tests and security controls passed, but the complete Malibu + CLI + launchd + Keychain + coordinator + catalog + model journey failed repeatedly on real Macs. The written v1.8.40 gate was amended so remaining checks moved after publication. That made the stable release channel an integration-test surface again.

This issue implements the permanent recurrence-prevention control:

> No exact signed physical journey evidence, no stable release.

A narrative checklist, issue comment, decision-log entry, or reviewer assertion is not sufficient evidence.

### Private candidates stop before physical acceptance

`.github/workflows/acceptance-candidate.yml` builds, signs, notarizes, binds, and uploads a short-lived private acceptance candidate. It does not execute the candidate on the physical Mac matrix and does not emit a signed physical pass/fail result.

### Physical acceptance is manual

`phase3-binary/dist/release-signing-runbook.md` contains manual clean-install, upgrade, and rollback commands. Those commands do not produce a canonical signed evidence object that the release workflow can verify.

### Stable publication does not consume acceptance evidence

`.github/workflows/release.yml` verifies release source, signatures, checksums, tag identity, GitHub environment posture, draft identity, and uploaded asset hashes. The final `draft=false` transition has no required acceptance-candidate ID, physical evidence digest, journey matrix, hardware-role proof, soak result, or Malibu/buyer-serving assertion.

### Tested candidate bytes are rebuilt

The v1.8.40 private candidate was produced from `f10da4b3`; the stable release was rebuilt from `18638472`. The intervening diff was documentation-only, but the process still relied on source equivalence instead of promoting the exact accepted binary/app/catalog/coordinator bytes.

### Written gates can be amended

Decision Entry 162 converted remaining pre-publication requirements to publication-then-verify because no private staging-Pearl path existed. That escape hatch must not exist for a stable channel.

### 1. One canonical lifecycle journey contract

Add a normative provider lifecycle journey specification plus a versioned machine-readable manifest.

Each required journey MUST define:

- journey ID and contract version;
- actual starting-state profile;
- authoritative state writer and source of truth for every asserted field;
- candidate components under test;
- required physical hardware role;
[...]

PR #647, opened 2026-07-19
Preserve exact CLI bytes in signed Malibu candidates

Augustas11:
## Outcome

Unblocks the exact signed referral/X release candidate by preserving the already-signed standalone CLI bytes when that CLI is embedded in Malibu.

The protected signer previously installed the signed CLI into Malibu and then used `codesign --deep`, which re-signed the nested executable. Both executables still reported 1.8.49, but their hashes differed, so the transactional updater correctly failed closed with `embedded_cli_mismatch` before cutover.

## Fix

- sign the nested CLI first, as before
- copy that exact binary into Malibu
- sign only the outer Malibu bundle
- verify the nested CLI hash is unchanged across outer signing
- continue to run deep signature verification after signing

No updater verification is relaxed.

## Verification

- signer/test Bash syntax
- `git diff --check`

{
  "behavior_change": "yes",
  "contract_change": "none",
  "issue": "https://github.com/Augustas11/macprovider/issues/613",
  "specs": ["SPEC-020", "SPEC-025"],
  "requirements": ["SPEC-020-R003"],
  "authority_domains": ["provider-autoupdate", "native-app-lifecycle"],
  "arbitration": ["CODE_BUG"],
  "tests": ["Acceptance-candidate security regression", "Independent Malibu release regression"],
  "journeys": ["Exact signed referral/X production activation tracked by #613"]
}

### DIFF
--- a/scripts/sign-acceptance-candidate.sh
+++ b/scripts/sign-acceptance-candidate.sh
@@ -213,13 +213,16 @@ rm -f "$cli_notary"
 install -m 0755 "$cli_work/macprovider-cli" "$app/Contents/MacOS/macprovider-cli"
 install -m 0644 "$cli_work/compatibility-set.json" "$app/Contents/Resources/compatibility-set.json"
 cmp "$cli_work/compatibility-set.json" "$app/Contents/Resources/compatibility-set.json"
-codesign --force --deep \
+embedded_cli_sha256="$(shasum -a 256 "$app/Contents/MacOS/macprovider-cli" | awk '{print $1}')"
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Reviewed release-security invariant: preserve the already-signed nested CLI bytes; outer-only signing retains deep verification and the updater exact-hash gate remains unchanged.

Augustas11:
Refreshing the governance event payload after correcting contract_change metadata; code head is unchanged.

Issue #616, opened 2026-07-16
P0: converge legacy upgrades to one complete signed compatibility set


Augustas11:
## Status (2026-07-22) — OPEN

Close criteria doc: [`audits/2026-07-22/CLOSE_CRITERIA_616.md`](https://github.com/Augustas11/macprovider/blob/main/audits/2026-07-22/CLOSE_CRITERIA_616.md) ([#691](https://github.com/Augustas11/macprovider/pull/691)).

| Layer | State |
|---|---|
| Code Partials on `main` | **DONE** — [#672](https://github.com/Augustas11/macprovider/pull/672) (PATH symlink converge) + [#678](https://github.com/Augustas11/macprovider/pull/678) (PATH regular-file converge) |
| Physical J4 (PATH strand) | **PASS** — private `Augustas11/macprovider:v1.8.58@8c4c57d87a2c59d7f7dd93556e05e3204ee4b093`; PATH regular-file → symlink/payload/set converge; buyer HTTP 200 |
| Public / Pearl | **NOT DONE** — no public tag containing #678; Pearl advertise still ~1.8.49 |
| Fleet PATH cohort | **NOT DONE** — legacy PATH shapes remain until hop through public release |
| Non-PATH ACs | **UNPAID, not waived** — mixed Malibu/manifest/watchdog fixtures, interrupted rollback matrix, 8 GB Air fixture, full rollback evidence pack |

### Remaining before close (exact gates)

1. **R1 — Public release/tag** containing merged #678 lineage; Pearl promote to that set ID
2. **R2 — Fleet hop** — every provider off divergent PATH copies onto the public set
3. **R3 — Unpaid whole-set ACs** — legacy watchdog, mixed-member fail-closed/converge, interruption rollback, 8 GB Air fixture, full before/after + rollback evidence

Merged Partials + private J4 PATH proof ≠ closed. Parent: [#585](https://github.com/Augustas11/macprovider/issues/585).

---

## Production ground truth

The main physical Mac is not one coherent installed release even though the installed CLI reports 1.8.40.

Observed on 2026-07-16:

- macprovider-cli --version: 1.8.40
- installed Malibu: 1.8.39
- install_manifest.json: version v1.8.3
- compatibility-set.json: signed v1.8.40 set declaring provider CLI 1.8.40 and Malibu 1.8.40
- lifecycle state also claims the v1.8.40 compatibility-set ID
[...]

Issue #612, opened 2026-07-16
P0: CLI update invalidates a healthy model and autotune rejects the feasible 8 GB fallback


Augustas11:
## Summary

A successful `macprovider-cli update` can turn an already-running, buyer-serving provider into a "no eligible model / donor mode only" workflow even though the installed model still loads and completes inference.

Observed on an Apple M1 MacBook Air with 8 GB unified memory after upgrading to macprovider-cli v1.8.40:

- The provider had already been running and serving with `mlx-community/Llama-3.2-3B-Instruct-4bit`.
- The update completed, then declared the prior recommendation stale and told the user to run `macprovider-cli autotune --recommend`.
- Autotune reported the Llama 3.2 3B probe as **feasible**, with `swap detected`.
- It then returned **No catalog model currently fits this Mac for network serving**, **Best compatible option: none**, and **Recommendation: donor mode only**.

This is not a missing-model or model-load failure. It is a composed update/freshness regression plus an autotune policy regression.

Related: #585. Regression source: #565.

## User impact

A routine CLI software update invalidates a known-good model decision, forces broad retuning, and can strand a working provider. The user sees an internal implementation label (`spec-023 probe`) and a misleading swap warning without the actual measured threshold that caused rejection.

A provider update must preserve service. A CLI version change alone must not make a working model ineligible.

### 1. Every CLI version change invalidates the previous recommendation

`RecommendationStateStore.isStale` returns stale whenever `stored.binaryVersion != current.binaryVersion`:

- `phase3-binary/Sources/macprovider-cli/AutotuneRecommend.swift:2019-2027`

The update command then explicitly tells the user to rerun broad autotune:

- `phase3-binary/Sources/macprovider-cli/MacProviderCLI.swift:1753-1759`

### 2. Cached successful benchmark evidence is also tied to the CLI version

`cachedBenchmarkAdmitted` requires `benchmark.binaryVersion == request.hardware.binaryVersion`:

- `phase3-binary/Sources/macprovider-cli/AutotuneRecommend.swift:1678-1694`

This discards known-good evidence after a software-only version bump even when the hardware, model artifact, catalog row, runtime compatibility, and serving configuration are unchanged.

### 3. Swap is not the rejection reason

The probe records successful inference as `.feasible` and separately records pageouts as `swapDetected`:

- `phase3-binary/Sources/macprovider-cli/AutotuneRecommend.swift:2113-2119`
[...]

Issue #610, opened 2026-07-16
P0: make CLI updates recover disconnected or coordinator-rejected providers


Augustas11:
## Status (2026-07-22) — OPEN

| Layer | State |
|---|---|
| Code Partial on `main` | [#673](https://github.com/Augustas11/macprovider/pull/673) (first-hop bridge) + PATH regular-file repair via [#678](https://github.com/Augustas11/macprovider/pull/678); contract [#630](https://github.com/Augustas11/macprovider/pull/630) |
| Physical J1 | **PASS** on private `Augustas11/macprovider:v1.8.58@8c4c57d87a2c59d7f7dd93556e05e3204ee4b093` — Shape A two-step (stranded public `1.8.48` PATH regular-file → exact set + buyer HTTP 200). Evidence: [comment](https://github.com/Augustas11/macprovider/issues/610#issuecomment-5043211212); cache `…/physical-reproof/2026-07-22-j1-j4-v1.8.58-071839Z` |
| Public / Pearl | **Not** released; Pearl advertise still ~**1.8.49**; no public `v1.8.58` tag |

**Merged Partial + private physical J1 PASS ≠ closed #610.**

Authoritative close checklist: [`audits/2026-07-22/CLOSE_CRITERIA_610.md`](https://github.com/Augustas11/macprovider/blob/main/audits/2026-07-22/CLOSE_CRITERIA_610.md) (in-repo after docs PR merges).

### Remaining before close

1. **Public release + Pearl advertise** — ship and promote the recovery Partial (#673 + #678) so production operators and stranded public 1.8.48 hosts can consume it (not private acceptance candidates).
2. **Production first-hop bridge** — if any provider remains on public 1.8.48 set, enable reviewed Pearl `first_hop_bridge_ids` per `exc-v1848-first-hop-update-bridge` and prove ordinary PATH `macprovider-cli update` on live Pearl (not acceptance overlay only); remove bridge when cohort empties.
3. **Deferred physical ACs (unpaid, not waived)** — disconnected-only autoupdate discovery, opt-out, rollback replay; rejected-old-version matrix partially paid via Shape A only. Automated tests on `main` do not substitute.

Parent umbrella: [#585](https://github.com/Augustas11/macprovider/issues/585) (G3 physical acceptance). #610 may close independently when the gates above pass; #585 stays OPEN until its full matrix completes.

---

## Problem

Provider recovery currently has a circular dependency:

[...]

PR #671, opened 2026-07-21
Stop CLI marketing-version bumps from invalidating recommendation evidence

Augustas11:
Partial #612

Note on SPEC-023 text: `specs/SPEC-023-installer-autotune-recommend.md` §9 / AC-25
currently documents `binary_version` as one of the inputs that should trigger a
stale-recommendation warning. This PR intentionally deviates from that specific
line per issue #612's required behavior (a CLI marketing-version-only change
must not invalidate evidence) and tracks under the existing `SPEC-023-R001`
gap (verdict `CODE_BUG`, owner `@Augustas11`, issue #612) already recorded in
`specs/CONFORMANCE.json`. The spec prose itself is not edited in this PR to
keep the diff minimal and scoped to the autotune/recommendation code + tests;
reconciling §9/AC-25 wording is left to the existing pending-reconciliation
tracking for SPEC-023.

## Summary

`RecommendationStateStore.isStale` and `cachedBenchmarkAdmitted` both keyed
evidence freshness on `binaryVersion` — the independently-versioned CLI
marketing release number. Any `macprovider-cli update` therefore:

- Discarded a healthy incumbent recommendation and its cached benchmark
  solely because the CLI version string changed.
- Forced a broad `autotune --recommend` nudge (surfaced from both `status`
  and `update`) on a pure software-only bump.
- Could collapse a feasible Llama 3.2 3B 8 GB fallback to `donor mode only`
  purely because the stale-cache check rejected the version string, even
  though hardware, model artifact, and catalog identity were unchanged.

## Fix

Bind both checks to the inputs that actually affect validity instead of the
CLI marketing version:

- `cachedBenchmarkAdmitted`: catalog row identity / digest, model ID +
  artifact SHA-256, hardware identity hash, and the existing evidence-age
  window (`maxBenchmarkAge`). `binaryVersion` no longer participates.
- `RecommendationStateStore.isStale`: rate card version, demand rank
  version, candidate catalog version + digest, hardware identity hash, and
  benchmark age. `binaryVersion` no longer participates.

Since the "Recommendation stale" nudge in both `status` and `update`
(`SelfUpdate.swift`, `MacProviderCLI.swift`) is derived entirely from
`RecommendationFreshnessChecker` → `RecommendationStateStore.isStale`, fixing
staleness at the source suppresses the spurious nudge on version-only bumps
without touching the messaging call sites directly.

Already-fixed items from the issue that this PR verifies but does not
re-touch: TPS/TTFT/swap remain advisory-only (thermal throttle is the sole
hard runtime gate), and the Llama 3.2 3B 8 GB catalog row is intact.

## Tests added (`AutotuneRecommendTests.swift`)

- `testIsStaleIgnoresCLIVersionOnlyChange` — a version-only bump with all
  compat inputs unchanged keeps `isStale` returning `false`.
- `testIsStaleDetectsCompatibilityInputChanges` — catalog digest, catalog
  version, and hardware identity changes still mark evidence stale.
- `testCachedBenchmarkAdmittedAcrossCLIVersionOnlyBump` — a cached benchmark
  recorded under an older CLI version is still admitted (and still drives
  `recommend()`) after a version-only bump.
- `test8GBLlama32FeasibleFallbackSurvivesCLIVersionOnlyBump` — an 8 GB Mac
  with a feasible (swap-flagged) Llama 3.2 3B benchmark from an older CLI
  version still recommends that model after a version bump, instead of
  collapsing to `donor mode only`.
- Updated `testCachedBenchmarkAdmissionRejectsEveryMismatch` — flipped the
  `binaryVersion`-mismatch case from reject to admit, since that dimension is
  intentionally no longer a compatibility input; all other mismatch
  dimensions (catalog digest, model ID, artifact SHA-256, hardware identity,
  evidence age) are unchanged and still reject.

I verified all four new/changed tests fail against the pre-fix source (by
temporarily stashing only the source change) and pass with it applied.

## Test results

```
swift test   # full phase3-binary suite
Executed 1519 tests, with 8 tests skipped and 0 failures (0 unexpected)
```

## Scope / out of scope

- Does not touch #616 (PATH/compat-set convergence) or #610/#658 (updater
  discovery) — no files under that scope were modified.
- Does not change TPS/TTFT/swap advisory behavior or the Llama 3.2 3B
  catalog row — verified intact, not re-implemented.
- Does not rename `spec-023`-labeled internal identifiers to
  "Model readiness check" user-facing language — out of scope for this
  minimal fix; `benchmarkID` prefix (`spec-023-…`) is not printed in normal
  (non-advanced/non-JSON) operator output today, so this is lower risk to
  leave for a follow-up.

## Remaining physical closure evidence (why this is "Partial", not "Closes")

Per the issue's closure criteria, this still needs physical proof on the
affected 8 GB Air:

1. Start the existing Llama 3.2 3B provider and prove it serves a request.
2. Run a signed CLI update.
3. Prove the same incumbent model remains selected and the provider remains
   available after restart.
4. Run recommendation diagnostics and prove swap/TPS/TTFT are reported
   without eliminating the feasible fallback.
5. Prove a second request is served after the update and diagnostics.

The code + unit test evidence in this PR is intended to unblock that
physical verification pass, not replace it.

### DIFF
--- a/phase3-binary/Sources/macprovider-cli/AutotuneRecommend.swift
+++ b/phase3-binary/Sources/macprovider-cli/AutotuneRecommend.swift
@@ -1703,8 +1703,11 @@ struct AutotuneRecommendEngine {
         } else {
             catalogEvidenceMatches = benchmark.candidateRowIdentity == request.candidateCatalog.rowIdentity(for: modelKey)
         }
+        // Evidence freshness binds to compatibility inputs (catalog row identity,
+        // model artifact, hardware identity) and an explicit evidence lifetime —
+        // never to the independently-versioned CLI marketing release number.
+        // A CLI version-only bump must not discard a known-good cached benchmark.
         guard catalogEvidenceMatches,
-              benchmark.binaryVersion == request.hardware.binaryVersion,
               benchmark.modelID == request.candidateCatalog.rows[modelKey]?.modelID,
               benchmark.artifactSHA256 == request.candidateCatalog.rows[modelKey]?.modelSHA256,
               benchmark.hardwareIdentityHash == request.hardware.hardwareIdentityHash,
@@ -2025,11 +2028,16 @@ enum RecommendationStateStore {
     }
 
     static func isStale(stored: LastRecommendationState, current: LastRecommendationState, now: Date) -> Bool {
+        // Staleness binds to compatibility inputs — rate card, demand rank, and
+        // candidate catalog identity/digest, plus hardware identity and evidence
+        // age. `binaryVersion` is the independently-versioned CLI marketing
+        // release number and intentionally does not participate here: a
+        // software-only version bump with unchanged compat inputs must not
+        // invalidate an otherwise-healthy recommendation.
         if stored.rateCardVersion != current.rateCardVersion { return true }
         if stored.demandRankVersion != current.demandRankVersion { return true }
         if stored.candidateCatalogVersion != current.candidateCatalogVersion { return true }
         if stored.candidateCatalogSHA256 != current.candidateCatalogSHA256 { return true }
-        if stored.binaryVersion != current.binaryVersion { return true }
         if stored.hardwareIdentityHash != current.hardwareIdentityHash { return true }
         guard let benchmarkGeneratedAt = stored.benchmarkGeneratedAt else { return true }
         return now.timeIntervalSince(benchmarkGeneratedAt) > AutotuneRecommendEngine.maxBenchmarkAge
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
CI green (all checks passing including phase3-binary swift test with the 5 new/updated unit tests). Reviewed the isStale/cachedBenchmarkAdmitted diff: removes binaryVersion from both compatibility checks, keeps rate card/demand rank/candidate catalog/hardware identity/benchmark-age checks intact. Verified locally that the new tests fail on pre-fix source and pass post-fix. Approving per issue #612 scope.

PR #695, opened 2026-07-22
Reclaim orphaned control sockets on serve startup instead of failing forever

Augustas11:
## Summary

Fixes today's production wedge (2026-07-22): launchd `KeepAlive` restarts of `macprovider-cli serve` failed forever with

```
control socket .../T/macprovider-cli/ctl.sock already exists; remove the stale file and restart serve
```

Process stayed "running" under launchd but never bound HTTP (`:61920`), leaving Malibu stuck on Starting.

**Incident facts (this Mac):**
- After J1/J4 physical update re-proof (~07:35), the watchdog logged `no validated PID` from ~07:40 while `KeepAlive` skipped restarts.
- The stale `ctl.sock` (mtime ~09:34) then blocked every subsequent `serve` start.
- Manual recovery required `bootout` + `rm -f ctl.sock` + `bootstrap`.

**Root cause:** `ControlSocketServer.start()` treated *any* pre-existing socket path as fatal (`staleSocket`), with no way to distinguish "another live serve owns this" from "a dead serve process left this behind." A process that dies before its watchdog exit cleanup runs (SIGKILL, OOM, power loss) leaves an orphaned socket file that then wedges every future restart — exactly the failure mode `KeepAlive` is supposed to recover from.

**Fix:** `ControlSocketStaleSocketReclaimer` reuses the same identity-verification and atomic quarantine-then-unlink discipline already proven in `ControlSocketWatchdogCleanup` (the exit-side cleanup), but runs it at startup instead:

1. Verify the path is genuinely our own control socket — owned by our euid, `S_IFSOCK`, mode `0600`, with a `0700` euid-owned parent directory. Any mismatch (regular file, symlink, wrong owner/mode, unsafe parent) is never touched and stays fail-closed with the existing `staleSocket` error.
2. Probe liveness with a real `connect()`. If a listener accepts, the socket is live — stay fail-closed exactly as before; some other `serve` process currently owns it.
3. Only `ECONNREFUSED` / `ENOTSOCK` / `EPROTOTYPE` / `ENOENT` (no listener, or the file is already gone) is treated as an orphan. Reclaim it via the same rename-verify-unlink sequence as the watchdog path: atomically rename to a quarantine name, re-verify the quarantined file's device/inode/owner/mode against what was observed before reclaiming, then unlink. If a live replacement won a race and now owns the name, the identity check fails and the rename is reversed — the replacement is never clobbered.

Same security posture as the existing reclaim path throughout: never unlink an unknown or live path, only ever act on something proven to be both ours and dead.

## Scope

Only `phase3-binary/Sources/macprovider-cli/ControlSocket.swift` (+ its tests). No changes to serve's HTTP/model runtime, watchdog exit cleanup, launchd plist, or coordinator admission. No Pearl mutation, no canary change, no release/promote.

## Test plan

- [x] `swift build` — clean.
- [x] `swift test --filter ControlSocketTests` — **51/51 passing**, including 3 new cases:
  - `testServerReclaimsOrphanedControlSocketOnStart` — an orphaned socket (bound+listened, then closed without unlinking, simulating a crashed serve) is reclaimed and the new server binds and serves normally.
  - `testServerFailsClosedWhenControlSocketIsLive` — a second `serve` attempt against a path with a live listener still fails closed with `staleSocket`, and the live peer is unaffected.
  - `testStaleSocketReclaimPreservesLiveReplacementOnIdentityRace` — quarantining with a stale (pre-race) identity against a path now owned by a live replacement fails and leaves the replacement's socket completely untouched.
- [x] Full `swift test` suite run and compared against a clean `origin/main` baseline (via `git stash`) — the pre-existing 26 `CoordinatorClientTests` failures + 2 `SelfUpdateTests`/`EndToEndAcceptanceTests` failures are identical on both, confirming they're environment-dependent (local MDM enrollment, missing keychain-access-groups entitlement, version-drift fixtures) and unrelated to this change.

## Residual risk

- Reclaim only fires when the socket file exists at `start()`; a still-running healthy `serve` is unaffected since its own socket connect always succeeds.
- The liveness probe treats any *unexpected* `connect()` errno (not in the reclaim allowlist) as fail-closed/`unverified`, matching the existing conservative posture rather than guessing.
- Does not address the separate watchdog `no validated PID` gap noted in today's incident timeline (watchdog skipping restarts before the stale-socket wedge) — that is tracked separately and out of scope for this fix per the assigned task.

Incident reference: `ops/runbooks/pearl-exception-clearance-20260722.md`. Extends the lifecycle-reliability line of work tracked under #585.

### DIFF
--- a/phase3-binary/Sources/macprovider-cli/ControlSocket.swift
+++ b/phase3-binary/Sources/macprovider-cli/ControlSocket.swift
@@ -480,7 +480,7 @@ public enum ControlSocketPaths {
     }
 }
 
-private struct ControlSocketFileIdentity: Equatable, Sendable {
+struct ControlSocketFileIdentity: Equatable, Sendable {
     let device: dev_t
     let inode: ino_t
     let owner: uid_t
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Approves orphan ctl.sock reclaim on serve startup — fixes KeepAlive wedge (Malibu stuck Starting). Fail-closed if live peer still owns the socket.

PR #719, opened 2026-07-23
fix(tools): stop null tool-schema defaults from 503ing as model_not_loaded (#718)

Augustas11:
## Summary

Fixes https://github.com/Augustas11/macprovider/issues/718.

OpenAI tool requests whose parameters schema contains a literal JSON `null` (commonly `"default": null`) failed with:

```json
HTTP 503 {"error":{"code":"model_not_loaded","message":"Model inference failed",...}}
```

even when the model was loaded and healthy. Root cause: `mlxToolsForTemplate` / `jsonAny` materialised `.null` as `NSNull()`, then swift-jinja's `Value(any:)` threw (no `NSNull` case). The HTTP residual catch stamped that throw as `model_not_loaded`.

### Changes

1. **Primary (correctness)** — `ModelRuntime.mlxToolsForTemplate`:
   - recursively **omit** JSON-null object members and array elements when building the chat-template tools dict (never emit `NSNull`)
   - omit `description` when missing or null (was `?? NSNull()`)
2. **Secondary (diagnosability)** — `HTTPServer` residual catch:
   - Jinja/NSNull/chat-template style failures → `400 chat_template_error`
   - other unexpected non-streaming failures keep historical `503 model_not_loaded` (error-receipt eligibility unchanged)
3. **Tests**
   - `testTemplateToolsOmitJSONNullsAndMissingDescription` (issue #718 regression)
   - mapper unit tests for Jinja/NSNull → `chat_template_error` and generic → `model_not_loaded`

### Line-number verification against `origin/main` (`7bc5ba87`)

Issue pointers matched current main closely:

| Site | Issue pointer | Actual on main |
|---|---|---|
| `mlxToolsForTemplate` | ~L2630 | L2630 |
| `description ?? NSNull()` | ~L2646 | L2646 |
| `jsonAny` `.null → NSNull()` | ~L2693–2694 | L2693–2694 |
| non-stream residual catch | ~L639–644 | L639–644 |
| stream residual catch | ~L1012 | L1012 |
| `validateTools` loose parse | ~L834–856 | L834–856 |

## Test plan

- [x] `swift test --filter ToolCallParserTests` (41 tests, including new null-strip case)
- [x] `swift test --filter HTTPServerReceiptTests/testUnexpectedInference*`
- [x] `swift test --filter HTTPServerReceiptTests/testHTTPGenericInferenceFailureGetsNullUsageReceipt`
[...]

Issue #718, opened 2026-07-23
Tool requests with `"default": null` in schema fail as 503 model_not_loaded (NSNull breaks swift-jinja chat template)


Augustas11:
## Summary

Any OpenAI `tools` request whose tool `parameters` schema contains a literal JSON `null` value (e.g. `"default": null`) fails with:

```json
HTTP 503
{"error":{"type":"server_error","code":"model_not_loaded","message":"Model inference failed","inference_ran":false,"settlement_ran":false,"retryable":false,"request_id":null,"param":null}}
```

The model is **loaded and healthy** — plain chat and null-free tool schemas return `200`. `model_not_loaded` is a **mislabel**: a `null` in the tool schema is materialized as `NSNull()` and thrown by the swift-jinja chat-template engine, then caught by a blanket handler that stamps it `model_not_loaded`.

## Impact

This blocks **all standard OpenAI function-calling / agent harnesses**, because real-world tool schemas routinely emit `"default": null` for optional parameters. Concretely it makes MacProvider unusable as the backend for Block's **Buzz** `buzz-agent` harness (tool-calls-as-output): 5 of its 6 built-in tools carry a `"default": null` property, so every agent turn 503s. Applies to the **provider CLI serving path** (both the local `/v1` endpoint and, by extension, any request the coordinator routes to a provider) — so it ships as a provider binary release, not a coordinator change.

## Environment

- `macprovider-cli` **1.8.58**, `serve` mode
- Model `qwen3-coder-30b-a3b-instruct` (MLX, mlx-community/Qwen3-Coder-30B-A3B-Instruct-4bit)
- Reproduced against the local loopback endpoint (`127.0.0.1:<port>/v1`), coordinator not in path

## Minimal reproduction

**FAILS → 503 `model_not_loaded`:**
```bash
curl -sS http://127.0.0.1:61920/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model": "qwen3-coder-30b-a3b-instruct",
  "messages": [{"role":"user","content":"hi"}],
  "max_completion_tokens": 8,
  "tools": [{"type":"function","function":{"name":"f","description":"d",
    "parameters":{"type":"object","properties":{"timeout":{"type":"integer","default":null}}}}}]
}'
```

**WORKS → 200 (identical, minus the `"default": null`):**
```bash
curl -sS http://127.0.0.1:61920/v1/chat/completions -H 'Content-Type: application/json' -d '{
[...]

PR #725, opened 2026-07-23
fix(tools): render JSON-null tool-schema values as native Jinja nulls (#718); revert flawed #719

Augustas11:
## What & why

Reimplements the #718 tool-schema `null` fix **correctly**, after reverting the flawed **#719**.

**Root cause (#718):** a tool-schema value of JSON `null` became Foundation `NSNull`, which swift-jinja's `Value(any:)` rejects — surfacing to buyers as a mislabeled `503 model_not_loaded` instead of a template error.

**Why revert #719 rather than patch on top:** #719 was admin-merged without the standard 3-lane audit and took a shortcut that introduced collateral defects:
- It **deleted** the null-valued keys instead of preserving them — silently mutating the buyer's tool schema (dropping `enum:[null,…]`, `const:null`, positional defaults).
- It introduced an **ungoverned `chat_template_error`** code outside the SPEC-001 taxonomy.

This PR is `revert #719` + a clean correct implementation, so history and review see the whole fix rather than a slice-on-a-slice.

## The correct fix

- **Native Jinja nulls** — `ModelRuntime.jsonAnyForTemplate` renders every JSON `null` as `Jinja.Value.null`, preserving object keys and array positions (shape-faithful, no key deletion).
- **Signed-hash parity for `description`** — an absent `function.description` renders as a native Jinja null too, matching `PromptCanonicalizer.canonicalTool` (via `jcsOrNull`), which already collapses absent and explicit-null to the same JCS null. The rendered template can no longer diverge from the signed prompt hash.
- **Depth guard at the real boundary** — `ChatCompletionRequest.validateTools` rejects over-deep tool schemas `400 invalid_tools` via an **iterative** (non-recursive) per-member depth-1 probe against `JSONSchemaValidator.maxDepth`, **before** the recursive `JSONValue.parse`. It's a raw stack-safety backstop and covers **every** tool member (no `tools[i].<extra>` bypass). Exact 32/33 boundary is tested.
- **Taxonomy** — drops #719's `chat_template_error`; stays on the established SPEC-001 set.
- Pins `swift-jinja 2.3.6` (already resolved transitively) for the `import`.

## Deliberately deferred (not silently changed)

The residual `model_not_loaded` mapping for **post-load inference failures** is **preserved**. It's pre-existing behavior that **SPEC-015 §M.5 (AC-31)** relies on to issue a null-usage (0-token) error receipt = proof of no charge. An audit lane flagged the taxonomy as impure; changing it alters signed-receipt economics and is left to a governed SPEC decision, documented in-code — not folded into a null-handling fix.

## Verification

- **80 tests pass, 0 failures**; builds green.
- **3-lane codex audit (code / security / architect)** against the full combined fix diff, iterated to **0 Critical/High/Medium**.
- Pre-existing, unrelated failures (`SelfUpdateTests` version-string assertion; MLX `default.metallib` in headless env) fail identically on clean `main` — out of scope.

## Governance note — advisory `spec-index / check` is red BY DESIGN

`SPEC-018` / `agentic-tool-calling` (which governs tool-schema rendering) has **no requirements defined** in `specs/CONFORMANCE.json`, so the declaration below cites the real spec and domain but an **empty `requirements` array**. The advisory `spec-index / check` therefore fails on "requires at least one requirement" — an **honest** red reflecting the governance gap, tracked in #726. Unlike #719 (which invented a non-existent `SPEC-018-R001` and was admin-bypassed), this PR does not fake a requirement. The **required** gate `ci-required` is unaffected.

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-018"],
  "requirements": [],
  "authority_domains": ["agentic-tool-calling"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "phase3-binary/Tests/macprovider-cliTests/ToolCallParserTests.swift",
    "phase3-binary/Tests/macprovider-cliTests/HTTPServerReceiptTests.swift"
  ],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/718"
}

### DIFF
--- a/phase3-binary/Package.swift
+++ b/phase3-binary/Package.swift
@@ -25,6 +25,13 @@ let package = Package(
             url: "https://github.com/huggingface/swift-transformers.git",
             exact: "1.0.0"
         ),
+        // swift-transformers 1.0.0 resolves swift-jinja 2.3.6 transitively; pin
+        // it directly so `import Jinja` (native null tool-schema rendering,
+        // issue #718) binds to the same reviewed version.
+        .package(
+            url: "https://github.com/huggingface/swift-jinja.git",
+            exact: "2.3.6"
+        ),
         .package(
             url: "https://github.com/apple/swift-nio.git",
             exact: "2.101.2"
@@ -57,6 +64,7 @@ let package = Package(
                 .product(name: "MLXLMCommon", package: "mlx-swift-lm"),
                 .product(name: "MLXHuggingFace", package: "mlx-swift-lm"),
                 .product(name: "Tokenizers", package: "swift-transformers"),
+                .product(name: "Jinja", package: "swift-jinja"),
                 .product(name: "NIO", package: "swift-nio"),
                 .product(name: "NIOHTTP1", package: "swift-nio"),
                 .product(name: "ArgumentParser", package: "swift-argument-parser"),
--- a/phase3-binary/Sources/MacProviderCore/ChatCompletionRequest.swift
+++ b/phase3-binary/Sources/MacProviderCore/ChatCompletionRequest.swift
@@ -852,7 +852,56 @@ private func validateTools(_ raw: Any?) throws {
         guard function["parameters"] is [String: Any] else {
             throw APIError(status: 400, message: "Invalid tools[\(index)]: missing function.parameters", code: "invalid_tools")
         }
+        // The whole tool object is later walked recursively by `JSONValue.parse`
+        // and the chat-template converter, both otherwise depth-unbounded. Bound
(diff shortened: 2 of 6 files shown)


antfleet-ops (approved):
Review-required gate: substantive review was the 3-lane codex audit (code/security/architect) against the full revert-#719+redo diff, iterated to 0 C/H/M; 80 tests pass. Advisory spec-index check red by design (honest empty-requirements declaration; gap tracked in #726).

Issue #608, opened 2026-07-16
P0: unify autotune and Tier-2 model catalogs under one release-bound authority


Augustas11:
## Problem

The production system has two independently deployed catalogs that repeat model identity and artifact hashes:

- The autotune/network catalog selects which model should run and binds the model artifact hash, policy, and recommendation data.
- The Tier-2 attestation catalog independently checks whether a provider-reported model hash is approved for trust and reward enforcement.

This duplication allows the two catalogs to drift. During recovery work for #585 on 2026-07-16, production autotune was correctly advanced to the signed 2026-07-10 release while Tier-2 had independently regressed to macprovider-tier2-model-catalog-2026-05-31, despite a newer signed 2026-07-07 P2 artifact being available. The result was model_hash_uncatalogued for a Qwen model whose exact identity and hash were already present in the active network catalog.

Tier-2 currently returned decision: allow, so this drift did not directly cause the outage. It is nevertheless an architectural defect: if enforcement changes, a valid release can strand providers, and operators must coordinate two overlapping mutable authorities.

## Target architecture

Create one signed, release-bound model identity authority. Autotune and Tier-2 policy may remain separate consumers with separate decisions, but Tier-2 must consume the same canonical signed model identity rows instead of maintaining a second model/hash file.

Canonical model identity should include at least:

- catalog_key
- model_id
- model_revision
- artifact_sha256
- policy_version or release identity

## Implementation requirements

1. Define one canonical schema and source of truth for signed model identity rows.
2. Bind autotune recommendations and Tier-2 attestation policy to one immutable network release identity.
3. Expose the active release identity and component digests through the coordinator.
4. Have providers report or resolve the exact release and model-row identity used for admission.
5. Remove the independently editable Tier-2 model/hash catalog after a bounded migration period.
6. Make publish and rollback atomic: tooling must never restore Tier-2 model identity separately from the active network release.
7. Support a current-plus-previous release compatibility window during migration without permitting arbitrary stale catalogs.
8. Add invariant checks at build, publish, coordinator startup, and provider admission boundaries.

## Acceptance criteria

- Autotune and Tier-2 cannot disagree about model identity or artifact hash because both resolve the same signed row.
- Deployment fails before mutation if any legacy or generated source duplicates conflicting model/hash data.
- Rollback restores one complete release and cannot independently regress Tier-2 identity data.
- A model present in the active release, including the Qwen row observed in #585 recovery, is catalogued consistently by both consumers.
[...]

PR #737, opened 2026-07-24
fix(release): accept reviewed OpenSSL formula vintages in release seal

Augustas11:
## Summary

- Allow the sealed release OpenSSL preflight to accept exactly the two reviewed `openssl@3` formula text digests observed across the GitHub `macos-15-intel` runner fleet.
- Preserve the exact OpenSSL `3.6.3` version, sequoia bottle tag/rebuild, pinned bottle sha256 `5477285c4ebec45713873ae4002affece39e427c5f1b655c6a3df49c6b90f924`, forced bottle install, and receipt checks.
- Update `scripts/test-release-security-posture.sh` so the guard requires both reviewed formula digests and fails if either digest or the membership check is removed.

## Root cause

Homebrew changed only the `openssl@3` formula recipe text while GitHub runners still expose both recipe vintages. A single formula-text hash therefore makes the release pipeline depend on runner-image vintage even though both formula states map to the same pinned bottle digest.

## Validation

- `bash -n scripts/install-sealed-release-openssl.sh scripts/test-release-security-posture.sh`
- `scripts/test-release-security-posture.sh`
- `git diff --check`

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-020"],
  "arbitration": ["CODE_BUG"],
  "tests": ["scripts/test-release-security-posture.sh"],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/608"
}

### DIFF
--- a/scripts/install-sealed-release-openssl.sh
+++ b/scripts/install-sealed-release-openssl.sh
@@ -8,7 +8,12 @@ readonly expected_formula_revision="0"
 readonly expected_bottle_rebuild="1"
 readonly expected_bottle_tag="sequoia"
 readonly expected_bottle_sha256="5477285c4ebec45713873ae4002affece39e427c5f1b655c6a3df49c6b90f924"
-readonly expected_formula_sha256="00e19cdcb1b7d99058a8a15f316e5dce2e4b5cd2afee14b272e7f5448624801d"
+# GitHub macos-15-intel runners currently expose either reviewed Homebrew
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Reviewed release-seal scope: dual reviewed formula-text digest allowlist keeps the OpenSSL bottle digest exact and updates the release-security guard in lockstep.

PR #740, opened 2026-07-25
Sign Malibu nested payloads without mutating provider CLI

Augustas11:
## Summary

Fixes the v1.8.62 production release failure in `Sign + notarize + staple Malibu.app`.

The prior hardening correctly removed recursive app signing so the embedded provider CLI would not be mutated after standalone signing. The production run then exposed the next missing step: `mlx.metallib` is a nested code-like payload and must be signed explicitly before the outer app is signed/verified.

This PR:

- signs `Contents/MacOS/mlx.metallib` explicitly before outer Malibu.app signing;
- signs copied SwiftPM `.bundle` directories explicitly before outer Malibu.app signing;
- keeps `macprovider-cli` untouched after the standalone signed CLI is copied into Malibu.app;
- keeps the SHA-256 byte-identity guard between standalone CLI and embedded CLI;
- extends the posture test to require explicit nested payload signing before outer app codesign.

## Validation

- `bash scripts/test-tier2-provider-release.sh`
- YAML parse for `.github/workflows/release.yml`
- `git diff --check`

## Release note

This is release-pipeline-only. It does not change product code or the v1.8.62 release tag payload. After merge, rerun `release.yml` for `v1.8.62` from `main`; the workflow file changes should apply while the release source remains the signed `v1.8.62` tag commit.

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-020"],
  "arbitration": ["CODE_BUG"],
  "tests": ["scripts/test-release-security-posture.sh", "scripts/test-tier2-provider-release.sh", "scripts/test-malibu-independent-release.sh"],
  "journeys": ["provider-release-verification"],
  "issue": "https://github.com/Augustas11/macprovider/issues/608"
}

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -758,6 +758,21 @@ jobs:
             exit 1
           fi
 
+          # Sign nested copied payloads explicitly before signing the outer
+          # app. Do not use codesign --deep here: the standalone provider CLI
+          # is already signed and must remain byte-identical to the tarball
+          # CLI after it is embedded in Malibu.app.
+          codesign --force \
+            --timestamp \
+            --sign "$SIGNING_ID" \
+            "$APP/Contents/MacOS/mlx.metallib"
+          while IFS= read -r -d '' nested_bundle; do
+            codesign --force \
+              --timestamp \
+              --sign "$SIGNING_ID" \
+              "$nested_bundle"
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Approved release-pipeline fix: signs non-CLI nested Malibu payloads explicitly while preserving standalone/embedded provider CLI byte identity and keeping recursive app signing removed.

PR #747, opened 2026-07-25
Mute provider-churn alert email on the Pearl monitor

Augustas11:
## Problem

The Pearl-side `macprovider-monitor` timer emailed **237 alerts in 7 days**.
A 3-day journald tally shows what they were:

| count | alert |
| --- | --- |
| 34 | `[WARN] gateway status -> idle` |
| 34 | `[CRITICAL] pool has 0 ready providers (idle)` |
| 25 + 18 | `[WARN] provider … dropped from pool (was ready)` |
| 10 + 6 | `[WARN] provider … -> unavailable` |
| 3 | genuine service-down (coordinator `/healthz`, gateway `/v1/status`, SPEC-023 static feeds 502) |

On a single-Mac fleet one provider oscillating produces the whole top block
every few minutes, and each transition is CRITICAL/WARN by severity even
though none of it is operator-actionable. The 3 alerts that *are* actionable
were buried under ~98% noise.

## Change

Every alert now carries a **kind**, and email is routed per kind rather than
per severity:

| kind | emailed by default |
| --- | --- |
| `provider`, `pool`, `gateway_status` | no — journal-only |
| `service`, `static_feed` | **yes** |

`EMAIL_MUTED_KINDS` in `/etc/macprovider/monitor.env` overrides the default:
an explicit list, `all` (email off entirely), or `none` (pre-mute behaviour).
Unknown names are logged and ignored so a typo cannot silently suppress a page.

Muting is a **delivery** decision only:

- journald still receives every alert, now prefixed `[SEV] (kind) message`
- muted alerts still count as alerting transitions for the state machine
- the "SMTP failed → do not seal state" guarantee is unchanged for the alerts
  that are still emailed

`ops/pearl-updater/macprovider-pearl-updater-alert`, which shares
`monitor.env`, is unaffected — this adds a key rather than removing one, so
its `REQUIRED_KEYS` check still passes and updater-failure mail still sends.

## Tests

New `phase4-coordinator/dist/test/check_monitor_email_mute_test.sh`:

- unit: `muted_kinds()` across absent/empty/`none`/`all`/whitespace/case/unknown
- unit: unconfigured email returns `None` (journal-only), not `False`
- source contract: every `alerts.append` site is tagged with a declared kind,
  and an untagged 2-tuple site fails the test
- end-to-end against a stubbed coordinator: a provider drop reaches journald
  but never SMTP; the same transition with `EMAIL_MUTED_KINDS=none` does
  email; a coordinator outage still emails CRITICAL

Existing `check_monitor_sandbox_test.sh` still passes.

## Audit

Three codex lanes against the full diff (prompts in `audits/2026-07-25/`):

- CODE — `READY | C=0 H=0 M=0 L=2`
- ARCHITECT — `READY | C=0 H=0 M=0 L=3`
- SECURITY — `NEEDS REVISION | C=0 H=0 M=1 L=0`

**SEC-M-1 carried, accepted, not fixed.** Muting `pool` means a *sustained*
pool-zero outage (not just flapping) is now journal-only as well, and journald
is the only sink on Pearl. That is the explicit intent of the request — these
alerts had stopped carrying signal. Restore paging for it at any time with
`EMAIL_MUTED_KINDS=provider,gateway_status` in `monitor.env`, no redeploy.

LOWs carried (no email-routing impact): CODE-L-1 — a mixed muted+emailable
round whose SMTP send fails re-prints the muted lines to journald next poll,
because state sealing is all-or-nothing; the service-alert retry guarantee is
what that protects. CODE-L-2 — the untagged-alert-site check is a source
regex, so a future alert assembled indirectly could evade it. ARCH LOWs are
about kind taxonomy and where the default policy should live.

## Governance declaration

`phase4-coordinator/dist/monitor/**` is a non-governance path, so
`behavior_change: "none"` is rejected — but no domain in `specs/AUTHORITY.json`
governs the Pearl-side monitor script (`provider-connection-diagnostics` /
SPEC-035 covers the coordinator's own diagnostics journal and `/admin/providers*`
endpoints, not this alert-mail router). Rather than fabricate a spec link, the
declaration below is filed honestly with empty spec arrays and
`spec-index / check` is expected to go red. That job is advisory; only
`ci-required` + 1 approval gate merge.

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": [],
  "requirements": [],
  "authority_domains": [],
  "arbitration": [],
  "tests": ["phase4-coordinator/dist/test/check_monitor_email_mute_test.sh"],
  "journeys": ["not-required"],
  "issue": ""
}

### DIFF
--- a/audits/2026-07-25/AUDIT_MONITOR_EMAIL_MUTE_R1_ARCHITECT_PROMPT.md
+++ b/audits/2026-07-25/AUDIT_MONITOR_EMAIL_MUTE_R1_ARCHITECT_PROMPT.md
@@ -1,0 +1,87 @@
+# AUDIT_MONITOR_EMAIL_MUTE_R1_ARCHITECT_PROMPT
+
+You are auditing the ARCHITECT lane of a change to the Pearl-side observability
+monitor in this repository (branch `fix/monitor-email-mute-provider-churn`).
+
+## Audit target — the FULL fix as it will land
+
+```
+git diff origin/main -- phase4-coordinator/dist/monitor phase4-coordinator/dist/test/check_monitor_email_mute_test.sh
+```
+
+Files:
+
+- `phase4-coordinator/dist/monitor/macprovider-monitor.py`
+- `phase4-coordinator/dist/monitor/README.md`
+- `phase4-coordinator/dist/test/check_monitor_email_mute_test.sh` (new)
+
+Related unchanged context you should read:
+
+- `phase4-coordinator/dist/monitor/macprovider-monitor.service` / `.timer`
+- `phase4-coordinator/dist/test/check_monitor_sandbox_test.sh`
+- `phase4-coordinator/dist/deploy-pearl-vps.sh` (monitor.env perms block)
+- `ops/pearl-updater/macprovider-pearl-updater-alert` (a SEPARATE tool that
+  reads the same `/etc/macprovider/monitor.env` and requires the keys
+  `ALERT_EMAIL`, `GMAIL_USER`, `GMAIL_APP_PASSWORD`)
+
+## What the change does and why
+
+The monitor runs on a 3-minute systemd timer and emails every alerting state
+transition. On the current single-Mac fleet, provider churn dominates: Pearl
+journald shows 237 emailed alerts in 7 days, ~95% of them
+`provider … dropped from pool`, `pool has 0 ready providers`, and
+`gateway status -> idle`. The operator asked for those to stop while real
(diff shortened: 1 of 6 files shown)


antfleet-ops (approved):
Approved. Scope is the Pearl monitor's mail routing only; journald record is unchanged apart from the kind prefix, and the shared monitor.env contract with the pearl-updater alert is preserved (key added, none removed). SEC-M-1 carried as documented.

Issue #658, opened 2026-07-20
P0: make signed release discovery compatible with immutable GitHub releases


Augustas11:
## Status (2026-07-22) — OPEN

| Layer | State |
|---|---|
| Code Partial on `main` | [#674](https://github.com/Augustas11/macprovider/pull/674) |
| Close criteria doc | [`audits/2026-07-22/CLOSE_CRITERIA_658.md`](https://github.com/Augustas11/macprovider/blob/main/audits/2026-07-22/CLOSE_CRITERIA_658.md) |
| Physical J2 | **PASS** on private `v1.8.57@ea9093e4` matrix (`1.8.55` → candidate, buyer 200). See status rollup comment. |
| Public / Pearl | Private candidate only; public immutable discovery on production releases unpaid |

### Remaining before close

Authoritative checklist: `audits/2026-07-22/CLOSE_CRITERIA_658.md`.

- **Gate 1:** Public immutable discovery journey on production vNext (private ≠ public promote)
- **Gate 2:** Renew-head-before-expiry ops for append-only transports (`renew-release-discovery-head.yml`)
- **Gate 3:** Stop condition — physically installed public v1.8.55 → supported bridge → exact **public** vNext, one launchd process, buyer HTTP 200

Merged Partial #674 + private physical J2 ≠ closed. Parent: [#585](https://github.com/Augustas11/macprovider/issues/585).

---

## Problem

The signed-release updater and GitHub immutable releases are not currently compatible across stable releases.

During the v1.8.55 rollout:

- exact candidate run `29777720073` and promotion run `29779612270` passed;
- public v1.8.55 was immutable and byte-identical to the physically accepted candidate;
- `macprovider-cli update` nevertheless failed with `GitHub release tag not found`;
- root cause: `.github/workflows/promote-acceptance-candidate.yml` published the signed discovery assets inside v1.8.55 but did not publish the separate `release-discovery` transport release read by `SignedReleaseDiscoveryHead.transportReleaseTag`;
- publishing that missing non-installable transport from the exact signed v1.8.55 assets restored the normal command:
  `Already up to date (v1.8.55)`.

The resulting `release-discovery` release is itself reported by GitHub as `immutable: true`. The older path in `.github/workflows/release.yml` expects to update it with `gh release upload --clobber`, which cannot be relied on for the next stable release under immutable-release enforcement.

## Why this is P0

[...]

PR #753, opened 2026-07-25
Fix the discovery-head renewal failing on every run under bash 3.2

Augustas11:
## Problem

`renew-release-discovery-head.yml` has **never completed successfully** — two runs in
its entire history, one cancelled (2026-07-24) and one failed (2026-07-25, run
[30172994013](https://github.com/Augustas11/macprovider/actions/runs/30172994013)).
The failure is deterministic, not flaky.

`policy_args` is populated only when the current discovery head carries a
`signed_policy_minimum` or a `signed_policy_revoked` entry. The live head carries
neither:

```json
"signed_policy_minimum": null,
"signed_policy_revoked": []
```

So the array is always empty, and `"${policy_args[@]}"` is an **unbound-variable error
under `set -u` on bash 3.2** — which is what macOS runners ship. The failing run's log:

```
.../87752473-....sh: line 88: policy_args[@]: unbound variable
```

The `sign` step died there, before writing its `transport_tag` output and before
copying the artifact index. The `publish` step then ran with an empty
`TRANSPORT_TAG:` and failed with
`no matches found for .../renew-discovery/compatibility-artifact-index.json`.

Both downstream symptoms have this one cause.

## Consequence

Signed release discovery could never be renewed, so both heads lapsed past
`expires_at` and the client's fail-closed guard
(`SignedReleaseDiscovery.swift:101`, `expires > now`) rejected them fleet-wide:

| Head | Target | Expired |
|---|---|---|
| Fixed `release-discovery` tag (`immutable: true`) | v1.8.55 | 2026-07-21T21:04:27Z |
| Newest transport | v1.8.61 | 2026-07-25T10:11:03Z |

Fresh installs were unaffected — `install.sh` does not consume the discovery
transport. This was an update-path outage only.

## Fix

Expand the array guardedly. Verified against real bash 3.2.57:

```
OLD empty  : policy_args[@]: unbound variable
NEW empty  : argc=2 [A] [B]
NEW filled : argc=4 [A] [--signed-policy-minimum=1.8.0] [--signed-policy-revoked=x y] [B]
```

Element quoting is preserved, including arguments containing spaces.

## Sweep

Checked every macOS-runner workflow for the same pattern. The other bare expansions
are arrays populated by construction — `acceptance-candidate.yml` starts
`package_resolution_args=(archive)`, and the release-asset arrays always carry
entries — so this is the only instance that can be empty in practice. Not changing
them.

## Proof required before this is called done

A green re-dispatch publishing a new transport whose `expires_at` is in the future.
Workflow success alone is not proof; the published head will be fetched and its
expiry checked.

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-020"],
  "arbitration": ["CODE_BUG"],
  "tests": ["scripts/verify-release-discovery-transport.py"],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/658"
}

### DIFF
--- a/.github/workflows/renew-release-discovery-head.yml
+++ b/.github/workflows/renew-release-discovery-head.yml
@@ -231,6 +231,9 @@ jobs:
             --openssl "$OPENSSL_BIN" >/dev/null
           cmp "$WORK/latest/compatibility-artifact-index.json" \
             "$WORK/current/compatibility-artifact-index.json"
+          # macos runners ship bash 3.2, where "${empty[@]}" is an unbound-variable
+          # error under `set -u`. A head with no signed policy minimum and no
+          # revocations leaves this array empty, so expand it guardedly below.
           policy_args=()
           while IFS= read -r arg; do
             [[ -n "$arg" ]] && policy_args+=("$arg")
@@ -270,7 +273,7 @@ jobs:
             --attempt "$GITHUB_RUN_ATTEMPT" \
             --compatibility-manifest "$WORK/latest/compatibility-set.json" \
             --target-artifact-index "$WORK/latest/compatibility-artifact-index.json" \
-            "${policy_args[@]}" \
+            ${policy_args[@]+"${policy_args[@]}"} \
             --issued-at "$issued_at" \
             --expires-at "$expires_at" \
             --private-key "$key" \


antfleet-ops (approved):
Root cause is precise and the fix is verified against real bash 3.2.57 rather than assumed. Sweep of other macOS-runner workflows checked each candidate array instead of blanket-editing. Approving.

Issue #770, opened 2026-07-26
Seam hardening before OpenRouter enrollment


Augustas11:
Before enrolling our gateway as an upstream provider on OpenRouter, close a set of correctness gaps in the seams between buyer ↔ gateway ↔ coordinator ↔ provider. OpenRouter scores providers on latency + error rate (and routes buyers away on strikes), reconciles billing on usage, and retries requests — so these are cheap to fix now and costly/visible once we take scored traffic.

Each child is an independent, closeable fix with a `file:line`, a failure scenario, and (where built) a regression test that asserts the buggy behavior today and flips red when fixed. Harness + risk register: `audits/seam-hardening/`.

**Harm categories:** ranking · money · supply · credibility.

### P0 — before enrollment
- [ ] #759 — Attestation surfaces must reflect hardware tier, not key-custody  `credibility`
- [ ] #760 — Per-phase deadlines (replace the single request wall-clock)  `ranking` · test ✅
- [ ] #761 — Don't strike provider health on buyer/gateway cancellation  `supply` · test ✅

### P1 — before real volume
- [ ] #762 — Dedupe id-less retries on the success path  `money` · test ✅ (complements closed #200)
- [ ] #763 — Durable settlement journal  `money` · test ✅
- [ ] #764 — Clamp provider-reported capacity + permanent tripwire  `ranking`/`supply`
- [ ] #765 — Fail-suspect on absent verified benchmark  `ranking`

### P2 — debt
- [ ] #766 — Single-terminal-wins request arbiter  `money`
- [ ] #767 — Enforce binary version floor + client rejection handling  `supply`
- [ ] #768 — Per-model minimum version floor  `supply`
- [ ] #769 — Hello-gate spec vs prod config; same-account timing risk  `hygiene`

**Cheapest-first:** #759 + #761 (two-liners) → #760 (flat wall) → #764 (one clamp closes two gaps) → #762 / #763 (money) → P2 debt.

Augustas11:
All 11 children closed — epic complete (2026-07-27, single session, 9 PRs).

| Tier | PRs | Children |
|---|---|---|
| Harness | #774 | risk register + 8-scenario tripwire harness |
| P0 | #776, #777 | #759 attestation honesty · #761 no strike on buyer cancel · #760 per-phase deadlines |
[...]

Issue #762, opened 2026-07-26
Bill each request once: dedupe id-less retries on the success path


Augustas11:
**Seam:** usage & settlement · **Harm:** money · **P1** · tripwire ✅

An id-less buyer retry mints a fresh `request_id` (`phase5-gateway/internal/router/server.go:240-244`), so each attempt reserves and settles independently → the buyer is billed twice. OpenRouter retries requests, often with no idempotency key, so this is live-fire on enrollment.

Note: #200 (closed) added local `Idempotency-Key` dedupe for **retriable error paths** (client supplies the key). This issue is the complementary case — a **successful** completion retried with **no** idempotency key at all — which #200 does not cover and which the tripwire reproduces (bills 40 for 2×20).

**Fix** — gateway-native dedupe keyed on `(account, request fingerprint)` when the buyer supplies no stable id; do not rely on the buyer sending `X-Request-ID`/`Idempotency-Key`.

**Acceptance** — `TestSeamH5a_IdlessRetryDoubleBills` bills 40 today; flip to assert billed once (20). Control `TestSeamH5b` (stable UUID → 409) already passes.

Detail: `audits/seam-hardening/findings.md` (P1-1).

PR #780, opened 2026-07-27
Seam hardening P1: durable settlement journal — no billed-but-dropped usage rows (#763)

### LINKED ISSUE #763: Durable settlement journal — no billed-but-dropped usage rows
**Seam:** usage & settlement · **Harm:** money · **P1** · tripwire ✅

On the streaming after-commit path, if `SettleReservation` and the `EnsureUsageEvent` fallback both fail (`phase5-gateway/internal/router/chat_proxy.go:1974-1993`), the gateway logs, refunds, and **drops the usage row** — but the buyer already received a committed 200 stream. We delivered tokens (paid the provider / spent compute) and bill nothing, with no durable record to reconcile from. Any settle-time DB pathology or crash between reserve and settle is a silent revenue leak.

**Fix** — an append-only settlement journal keyed by `(account, request_id) + effect type`, written before dispatch and sealed on terminal, with a startup recovery scan that re-drives unsealed effects. Reservation-before-dispatch already exists; this adds the durable settle half.

**Acceptance** — `TestSeamH7_SettlementNotCrashDurable` asserts `usageRows==0` (dropped) on a forced double-failure today; flip to assert the usage is durably recorded/recoverable.

Detail: `audits/seam-hardening/findings.md` (P1-2).

Augustas11:
Closes #763. PR 5 of epic #770 — completes the P1 tier.

## What changed

On the streaming after-commit path, if `SettleReservation` AND the `EnsureUsageEvent` fallback both failed, the gateway logged, refunded, and **dropped the usage row** — the buyer had a committed 200 stream, the provider gets paid via the coordinator, and nothing is billed with no durable record. The SPEC-022 reconciler structurally cannot catch it (it scans `active+settlement_hold=1` rows and rejects non-`coordinator_observed` tokens).

**Mechanism**: a separate append-only JSONL journal (`internal/settlement/journal`) with fsync-per-effect — deliberately **not** a table in `gateway.db`, which would flip the tripwire while covering none of the real failure classes (same handle, same file, same WAL, `synchronous=NORMAL`), and not a second sqlite file (same driver, correlated failure). Effects are armed in `settleAfterCommit` before the settle attempt (full billing payload, never prompt/response text; 0700/0600) and sealed after; the double-failure branch leaves them unsealed. Journal write failures are fail-open (bytes already flowed) and metric-visible. A hard total-size cap refuses writes with a CRITICAL log rather than filling the disk.

**Recovery** runs at startup (bounded pass; CRITICAL-but-serve on failure) and on its own periodic loop (own knobs, independent of the SPEC-022 reconciler): grace window, batch cap, the existing idempotent settle ladder (`EnsureUsageEvent`'s payload verify is the no-double-bill guard), quarantine after repeated payload conflicts, retention prune. v1 journals the settle effect only — reservation-before-dispatch is already the durable arm; `SettleDemoReservation` now wraps `ErrReservationTerminal` so the ladder can classify demo terminals.

Tripwire flipped: `TestSeamH7_SettlementRecoveredFromJournal` — the forced double-failure leaves one unsealed effect; a second Server over the same store recovers exactly one usage row (billed 20), idempotent on re-scan. 27 new tests incl. fsync-order (the load-bearing property), settle-landed-but-seal-lost, crash-window, conflict quarantine + hold release, torn tail, rotation/prune, demo re-drive, reconciler race, #762-replay-writes-no-effect, config back-compat via real `Load()`.

## Audit (three codex lanes, full diff — four rounds, each a strict narrowing)

- security R1: **PASS 0/0/0** (accepted).
- code R1 HIGH: recovery wedged on the §17.7 crash window (active reservation + existing usage row → unclassified settle error) → fixed with fall-through to the idempotent rung. code R1 MEDIUM: `fsyncDir` swallowed real errors → propagate (EINVAL/ENOTSUP whitelisted). architect R1 MEDIUM: §17.7 sealed past a failed refund → seal gated on refund success.
- code R2 MEDIUM: the same seal gate was missing at recovery's own refund → mirrored. During this fix the H7 tripwire itself caught an over-strict gate (`ErrReservationNotFound` is the benign already-terminal sentinel and must still seal — treating it as failure wedged effects perpetually unsealed).
- architect R2 MEDIUM: the conflict path never released an active hold → best-effort refund before retry/quarantine. code R3: **PASS 0/0/0**.
- architect R3 MEDIUM (final narrowing): a refund failure on the last conflict attempt still quarantined, stranding the hold → a failed hold release no longer consumes an attempt. architect R4 (final round): **PASS 0/0/0**.

## Deliberately not covered (documented in findings.md)

Disk-full (C6 — no local journal survives it); journal-write + settle compound failure (fail-open, metric-visible); crash mid-stream before any terminal (reaper refunds, as today); the SPEC-022 hold-marker sibling leak (`effect` field future-proofed, not v1); multi-instance journal arbitration (single gateway today, no lockfile). Scope is the streaming after-commit path — non-streaming settles fail closed pre-response and have no delivered-but-unbilled window. Operational notes: recovery bills retroactively into the journaled `window_date`; one fsync per billed streaming request (after commit, invisible to TTFT; group-commit escape hatch deliberately unbuilt); `gateway -check` now creates the journal dir.

All three suites green pre-push: gateway `go build/vet/test ./...`, `-race` on router+journal, and the cross-service integration module.

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-005", "SPEC-006", "SPEC-036"],
  "requirements": ["SPEC-036-R001"],
  "authority_domains": ["billing-settlement-formula", "buyer-api-error-contract", "compute-integrity-settlement"],
  "arbitration": ["CODE_BUG"],
  "tests": ["phase5-gateway/internal/router/seam_harness_test.go", "phase5-gateway/internal/router/settlement_journal_recovery_test.go", "phase5-gateway/internal/settlement/journal/journal_test.go"],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/770"
}

### DIFF
--- a/audits/seam-hardening/findings.md
+++ b/audits/seam-hardening/findings.md
@@ -96,12 +96,69 @@ conversation tag, byte-identical body, no ids, inside 60s of attempt 1's termina
 window, the five opt-outs, and the conflict counter — not by "did the buyer receive it", which the
 tripwire forecloses. Buyer-supplied `Idempotency-Key` dedupe on the coordinator path remains #200.
 
-**P1-2 · Settlement not crash-durable** — `money` — test `TestSeamH7`
-On the streaming after-commit path, if `SettleReservation` and the `EnsureUsageEvent` fallback both
-fail (`phase5-gateway/.../chat_proxy.go:1974-1993`), the gateway logs, refunds, and drops the usage
-row — the buyer already got a committed 200. Tokens delivered, nobody billed, no durable record.
-**Fix:** append-only settlement journal keyed by (account, request_id)+effect, written before dispatch,
-sealed on terminal, with a startup recovery scan. Reservation-before-dispatch already exists.
+**P1-2 · Settlement not crash-durable** — `money` — test `TestSeamH7` · issue #763
+**FIXED gateway-side (issue #763).** On the streaming after-commit path, if `SettleReservation` and
+the `EnsureUsageEvent` fallback both failed, the gateway logged, refunded, and dropped the usage row —
+the buyer already had a committed 200. Tokens delivered, nobody billed, no durable record. The
+refunded reservation carries `settlement_hold=0`, so it was invisible to the SPEC-022 reconciler
+(`ListSettlementHeldReservations` selects active + hold=1): nothing else would ever have found it.
+**Fix (shipped):** a durable append-only **JSONL settlement journal**
+(`phase5-gateway/internal/settlement/journal`), NOT a table in `gateway.db`. Same-DB storage would
+have produced a green test and an unchanged risk — every failure class that takes out the settle write
+(the `MaxOpenConns=1` write connection, a DB-file pathology, a torn WAL after power loss) takes the
+journal write out with it; a second sqlite file is the same driver and the same page cache. Segments
+`effects-<unixmilli>-<pid>.jsonl`, dir `0700` / files `0600`, records keyed
+`(account_id, request_id, effect)` in three kinds — `effect` / `seal` / `quarantine` — and **never**
+any prompt or response text.
+**Durability contract:** `WriteEffect` is one `write(2)` + `fsync` BEFORE returning, so the effect is
+on stable storage before the settle attempt begins (`TestSettlementJournal_EffectWriteFsyncs` pins the
+call order through the real file). Seals are deliberately NOT fsynced — a lost seal costs one
+idempotent re-drive, never a bill. Segment create/unlink fsync the parent directory.
+**Where it arms:** `settleAfterCommit` only — the single choke point every after-commit terminal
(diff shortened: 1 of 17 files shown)


antfleet-ops (approved):
Money-path change with four audit rounds, each a strict narrowing of the conflict/seal seam: security R1 PASS, code R1(1H+1M)->R2(1M)->R3 PASS, architect R1(1M)->R2(1M)->R3(1M)->R4 PASS 0/0/0. The H7 tripwire flipped and even caught an over-strict fix mid-loop. All three suites green incl. -race and the cross-service integration module. Deliberate non-coverage documented in findings.md and the PR body.

PR #783, opened 2026-07-27
Seam hardening P2: fleet + per-model binary version floors (#767, #768)

### LINKED ISSUE #767: Enforce a binary version floor + client handling of the rejection
**Seam:** fleet-version-floor · **Harm:** supply · **P2**

`required_binary_version` is set nowhere in prod config, and the Swift client has no `case 4004` handler (`phase3-binary/Sources/macprovider-cli/CoordinatorClient.swift:1460-1499`) — so a below-floor close is experienced as a silent reconnect loop, not an upgrade directive. Upgrade is advisory (a one-line stdout nudge); there is no `doctor` subcommand.

**Fix** — set `required_binary_version` in the prod overlay; add a `4004` / `version_unsupported` handler that surfaces the required target as an upgrade directive; add a `doctor` subcommand.

**Acceptance** — a below-floor build is closed and reports why + the required version; not a silent reconnect.

Detail: `audits/seam-hardening/findings.md` (P2-2).

### LINKED ISSUE #768: Per-model minimum version floor
**Seam:** fleet-version-floor · **Harm:** supply · **P2**

There is a per-model **hardware-tier** gate but no per-model **minimum-version** floor. An old-but-hardware-eligible build can serve a model that needs a newer engine for correctness/perf.

**Fix** — add a per-model minimum-version floor to the routing gate (public routing, self-route preflight, and warm-pool candidate gate should all share it so we never warm a box we won't route to).

**Acceptance** — a below-floor build cannot serve the gated model's buyer traffic.

Detail: `audits/seam-hardening/findings.md` (P2-3).

Augustas11:
Closes #767. Closes #768. PR 7 of epic #770 — **closes the epic's last children**.

## #767 — Binary version floor + client rejection handling

`required_binary_version: "1.8.33"` is now set in the committed prod config. The value is chosen to **fence nobody currently admissible**: 1.8.33 is the first release declaring a `compatibility_set_id`, and everything older already fails the live overlay's `accepted_ids` — the floor turns an opaque later rejection into an explicit, machine-readable one (the #610 bridge build is additionally exempt via `firstHopOnly`; a test pins floor ≤ latest). The Swift client gains a `4004 / version_unsupported` handler that surfaces the required version as an upgrade directive (stderr + structured event) and **terminates the reconnect loop** instead of hammering the coordinator, plus a `doctor` subcommand (offline diagnostics + `/healthz` check; `/healthz` now publishes `required_binary_version`, omitempty — no new endpoint). Config load rejects unparseable floors.

`doctor` against live Pearl immediately proved its worth: live advertises `1.8.61`/no floor while the committed file said `1.8.65` — the committed floor takes effect at the next deploy (documented; independently confirms P2-2 was real).

## #768 — Per-model minimum version floors

A coordinator config map, shipped **empty** (no behavior change by default) — deliberately *not* a field in the signed catalog `Row`, which would enter `RowIdentity` and invalidate every provider's cryptographically-bound benchmark evidence. **One shared check** (new `internal/versionfloor`, now the coordinator's single semver comparator) wired into every path that selects or touches a provider for a model: public routing, the pinned-provider preflight (the "self-route" path that bypasses the routing filter), the warm-pool gates on both transports plus canary serving, **`slotQueueCandidates` (a fourth leak the issue didn't name)**, and — from the audit — the slot-queue *poll*. Malformed provider versions fail safe (gated out, WARN). New retryable 503 `model_version_floor_unmet` (AST-guard registered; unreachable while the map is empty).

## Audit (three codex lanes, full diff)

- code R1: **PASS 0/0/0** (comparator equivalence, Swift close-frame parsing, gate wiring all verified).
- security + architect R1 converged on one MEDIUM: the slot-queue **waiter stores only providerID**, so `pollQueuedProvider` re-resolved against the live snapshot without rechecking the floor — a same-ID reconnect below floor could be served off the queue. Fixed with the shared check before reservation + a below-floor-replacement/above-floor-control regression test. Combined R2: **PASS 0/0/0**.

## Documented residual + carried LOWs

- **Floors rest on self-reported `binary_version`** (security MEDIUM, by design): the identity signature binds the provider *key to the claim*, not the running binary to a release — no provenance anchor exists in the system today (SPEC-033 hardware evidence doesn't hash binaries). Version floors are **supply hygiene against stale honest builds, not an anti-adversary control**; binary provenance is trust-model roadmap territory.
- LOW: `canaryProbeEligible` omits the floor (probe traffic only; weakens "don't probe what we won't route to"). LOW: the Swift receipt-key-rotation path makes bounded extra handshakes before the reconnect loop observes 4004.
- Floors are startup-only (not in the SIGHUP allowlist, matching pre-existing behavior) — a floor change needs a coordinator restart, with the single-provider-pool caution documented.

Verification: full coordinator suite, cross-service integration module, `swift build` + all 15 new Swift tests green. One pre-existing environment-dependent `SelfUpdateTests` failure (reads the real install authority instead of the injected version) proven pre-existing via clean-tree revert; untouched.

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-002", "SPEC-020", "SPEC-010"],
  "authority_domains": ["coordinator-admission-routing", "provider-autoupdate", "model-catalog-identity"],
  "arbitration": ["CODE_BUG"],
  "tests": ["phase4-coordinator/internal/versionfloor/versionfloor_test.go", "phase4-coordinator/internal/buyer/seam_768_version_floor_test.go", "phase3-binary/Tests/macprovider-cliTests/VersionFloorRejectionTests.swift"],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/770"
}

### DIFF
--- a/audits/seam-hardening/findings.md
+++ b/audits/seam-hardening/findings.md
@@ -11,7 +11,7 @@ built) live in `harness/` and are named `TestSeam*`.
 | 2 · cache-tenant-isolation | **PASS by construction** | server-authored account-HMAC scope; KV in-memory only |
 | 3 · backend-rollout-safety | **PASS** (gap closed, #764) | strong flags/canary/self-update; perf telemetry now segmented by `binary_version` on `/poolz` |
 | 4 · attestation-honesty | **FAIL (client-facing)** | prose honest, but stat surfaces over-claim hardware trust |
-| 5 · fleet-version-floor | **PASS (1 gap)** | blocks incompatible/known-bad builds; capacity now clamped at ingest (#764); version floor still unset in prod (P2-2) |
+| 5 · fleet-version-floor | **PASS** (gaps closed, #767/#768) | blocks incompatible/known-bad builds; capacity clamped at ingest (#764); floor now SET in the prod overlay + a 4004 close stops the client's reconnect loop (P2-2); per-model routing floors share one check across routing/self-route/warm gates (P2-3) |
 
 Fundamentals are strong (cache isolation, rollout discipline, signed-catalog admission,
 honest prose). Exposure concentrates in two client-facing FAILs (settlement flat-wall,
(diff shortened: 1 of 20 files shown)


antfleet-ops (approved):
Final epic child. code R1 PASS; security+architect R1 converged on the slot-queue poll bypass — fixed with the shared check + regression; combined R2 PASS 0/0/0. Floor 1.8.33 fences nobody currently admissible (verified against the compatibility-set admission path); per-model floors ship empty; provenance limitation framed honestly as a documented residual. All suites green incl. Swift (one pre-existing env-dependent failure proven pre-existing via clean-tree revert).

Issue #785, opened 2026-07-27
P0: verify Pearl deploy path after internal-auth fallback removal (boot-blocking config + operator-key canary)


Augustas11:
## What

Commit `4bd33a2e` ("Remove legacy operator fallback from internal auth", merged 2026-07-26) makes the next Pearl coordinator deploy fail-prone in two independent ways, and its own trailer declares `Not-tested: post-merge Pearl redeploy`. Fleet is single-coordinator (`coordinator.streamvc.live`), with a documented 5-hour-outage precedent for coordinator restarts (incident 2026-07-10).

## Evidence

1. **Boot-blocking config requirement.** `phase4-coordinator/internal/config/config.go:1648` now hard-fails `Validate()` when `auth.gateway_service_token` is empty (mirrored in gateway config). A coordinator (re)started against a pre-cutover `coordinator.yaml` will not boot. The new deploy gate (`check-deploy-config.sh` C2c pairing proof) aborts fail-closed *before* stopping the service, so the likely failure mode is a blocked deploy rather than an outage — but an out-of-band process restart (crash, systemd) with a stale config is an outage.

2. **Undeclared auth tightening on the deploy canary endpoint.** `phase4-coordinator/internal/buyer/server.go:1130`: `/v1/pool/check?details=deployment` moved from service-token auth to `auth.OperatorOnlyBearerMatches(r.Header, s.operatorKey)`. The deploy script's pre-commit canary hits exactly this endpoint with `CATALOG_CANARY_AUTH_TOKEN` (`phase4-coordinator/dist/deploy-pearl-vps.sh:3495`). If the operator's canary token on Pearl is a service token (the previously valid choice), the next production deploy aborts at the canary gate. This scope change is not mentioned in the commit subject; the tracker `audits/2026-06-10/M3-2_LEGACY_FALLBACK_REMOVAL.md` lists restoring the canary token as a root-owned *operator* credential as unfinished pre-merge evidence.

3. **No deploy proof since merge.** Prod `/healthz` (checked 2026-07-27) reports `version v1.8.60`, `uptime_s ~53k` (~15h). No evidence the post-merge coordinator binary+config combination has ever been exercised on Pearl.

## Done when

- [ ] Verified on Pearl: `coordinator.yaml` and `gateway.yaml` carry non-empty, distinct `gateway_service_token` / `coordinator.service_token` values that satisfy the new `Validate()` + C2c pairing gate.
[...]

PR #798, opened 2026-07-28
Fix provider watchdog PID validation on macOS

Augustas11:
## Summary

Fixes the provider watchdog launchd PID validator so a healthy provider process is accepted when `lsof -d txt -Fn` reports the provider binary plus additional macOS text vnode mappings such as dyld or mapped runtime files.

The previous check required exactly one text mapping. On the Pearl canary Mac, that produced repeated false `has no validated PID` watchdog messages even though launchd was running the configured provider binary and the local provider listener was healthy.

## Changes

- Accept the configured provider binary when it appears anywhere in the normalized `lsof -d txt` path set for the launchd PID.
- Preserve fail-closed behavior when launchd reports ambiguous PIDs or the provider binary is absent from the validated process mappings.
- Update the installer-embedded watchdog copy and the inline drift check remains green.
- Extend watchdog health-scope coverage with multi-text-mapping fixtures.

## Validation

- `bash phase3-binary/dist/test/watchdog_health_scope.test.sh`
- `bash phase3-binary/dist/test/watchdog_rollback_paths.test.sh`
- `git diff --check`
- `make test-dist`

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-020"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "phase3-binary/dist/test/watchdog_health_scope.test.sh",
    "phase3-binary/dist/test/watchdog_rollback_paths.test.sh",
    "scripts/test-watchdog-inline-drift.sh",
    "make test-dist"
  ],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/785"
}

### DIFF
--- a/ops/macprovider-watchdog/watchdog.sh
+++ b/ops/macprovider-watchdog/watchdog.sh
@@ -134,12 +134,20 @@ provider_process_pid() {
   command -v lsof >/dev/null 2>&1 || return 1
   executable_output="$(lsof -a -p "$candidate" -d txt -Fn 2>/dev/null)" || return 1
   command_paths="$(printf "%s\n" "$executable_output" | awk 'substr($0, 1, 1) == "n" && length($0) > 1 { print substr($0, 2) }')"
-  [ "$(printf "%s\n" "$command_paths" | awk 'NF { count++ } END { print count + 0 }')" -eq 1 ] || return 1
-  command_path="$command_paths"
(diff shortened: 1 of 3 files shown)


antfleet-ops (approved):
Reviewed after current-head CI green. Scope is limited to watchdog PID validation and matching installer heredoc plus regression coverage.

PR #834, opened 2026-07-31
chore(spec-022): migrate requirement IDs (register SPEC-022-R001..R011)

Augustas11:
## Summary

Migrates **SPEC-022 (verified-model-settlement)** into the requirement-ID
conformance system: registers `SPEC-022-R001`..`SPEC-022-R011` — one per
top-level normative requirement group R-1..R-11 — in `specs/CONFORMANCE.json`,
and flips SPEC-022 `requirement_id_migration` from `"pending"` to `"complete"`.

## Why

The `verified-model-settlement` authority domain (owner SPEC-022; consumers
SPEC-005/015/016) had **zero** migrated requirement IDs in `CONFORMANCE.json`.
Consequently any behaviour-change PR touching settlement — e.g.
[#833](https://github.com/Augustas11/macprovider/pull/833), the coordinator
`usage_mismatch` fix — cannot file a valid `SPEC-GOVERNANCE` declaration
(`behavior_change: yes` on a non-governance path requires a real requirement ID,
and `"none"` is disallowed there), so the advisory `spec-index / check`
declaration substep fails for an unmigrated corpus rather than for anything wrong
in the PR. This migration provides citable IDs and closes that gap for the whole
domain. Same pattern already applied to SPEC-035/036/037.

## What changed

- `specs/CONFORMANCE.json` — add 11 `SPEC-022-R0NN` requirement entries (state
  `pending`, empty implementation/tests/journeys/evidence, owned issue-linked gap
  `#614`), and set SPEC-022 `requirement_id_migration: "complete"`, `version:
  "v0.1.6"`, with a migration rationale on its gap.
- `specs/SPEC-022-verified-model-settlement.md` — anchor each ID in its `### R-N.`
  header, add a mapping note, bump to v0.1.6 with an **editorial, non-normative**
  changelog entry.
- `specs/README.md` — regenerated index.

**No requirement text, obligation, or observable contract changes.** The IDs are
one-to-one with the existing normative groups; nothing about settlement behaviour,
money movement, or the SPEC's obligations is altered. All entries are `pending`
because the settlement pipeline's conformance reconciliation is still tracked
under `#614` — identical to how SPEC-035/036/037 registered their catalogs.

## Verification (local governance gates)

- `python3 scripts/check_spec_governance.py` → **passed**
- `python3 scripts/gen_spec_index.py --check` → **ok: spec index is up to date**
- `python3 scripts/gen_spec_index.py --lint` → **ok: specs/ root is canonical-only**
[...]

Issue #825, opened 2026-07-30
Tracking: prevent CLI/coordinator contract split-deploys (ordering gate + runnable Pearl deploy)


Augustas11:
## Summary

Tracking issue for the two highest-leverage gaps behind the 2026-07-30 onboarding outage (fresh Malibu installs failing `catalog_trust_blocked`). The root cause was **not** a code bug — the repo was correct — but two process gaps that let a contract-tightening change ship to users while its coordinator half never reached production.

## Gap A — CLI/coordinator contract split-deploy (no ordering gate)

#813 tightened a coordinator-served contract: CLI **v1.8.67** requires all three autotune trust feeds (`autotune-candidates`, `demand-rank`, `rate-card`) signed + mutually paired. The **CLI half shipped to users** (v1.8.67 / Malibu 1.8.46) while the **coordinator half (#813) was never deployed to Pearl** — prod sat on v1.8.66 for ~2 weeks serving an unsigned rate-card. Nothing gated the CLI release on coordinator readiness.

**Wanted:** a rule/gate that a CLI change tightening a coordinator contract cannot ship to users until the coordinator serving that contract is live in prod. Options:
- Capability/version negotiation: CLI enforces the signed-rate-card requirement only when the coordinator advertises it (graceful fallback otherwise).
- Release ordering: contract-tightening = coordinator-first deploy, verified live (see the "does prod serve what we shipped" gate), then CLI release.

## Gap B — the safe Pearl deploy is un-runnable on the operator Mac

`phase4-coordinator/dist/deploy-pearl-vps.sh` is the atomic, canary-gated, rollback-guarded deploy — but it could not be run this session:
- `CATALOG_CANARY_SSH_KEY` (`~/.ssh/macprovider_canary_ed25519`) is **absent**.
- `CATALOG_CANARY_AUTH_TOKEN` (the coordinator operator key) lives **only** in Pearl's `/etc/macprovider/coordinator.env` as `OPERATOR_KEY`; it is not in the operator's keychain, and prior sessions only ever used a `dev-operator-key`.

Because the safe path was un-runnable, the fix had to be applied via a hand-rolled manual deploy (binary + feeds swapped over SSH, nginx patched by hand). That worked with a `--validate-config` preflight + backups + rollback, but it bypassed the canary proof and the deploy's config-reconciliation.

[...]

Issue #822, opened 2026-07-30
Release gate: verify the live coordinator serves the signed feed set the shipped CLI requires


Augustas11:
## Problem

Our release verification proves the CLI/Malibu **artifacts** (byte-identity, notarization, updater invariant) but never checks that the **live production coordinator serves the feed set the newly-shipped CLI requires**. When a CLI change tightens a coordinator-served contract, a green release + published artifacts can still be broken in production because the coordinator half was never deployed.

## What happened (2026-07-30)

#813 made CLI **v1.8.67** require all three autotune trust feeds (`autotune-candidates`, `demand-rank`, `rate-card`) to be **signed and mutually paired**. The CLI shipped to users while prod coordinator stayed on **v1.8.66**, which served `rate-card` unsigned and `/v1/rate-card.sig` → 404. Every fresh install failed `catalog_trust_blocked`. Nothing in the release flow caught it; the first pre-beta user did.

## Proposed gate

Add a post-publish release check (fail the release / page on red) that probes the **live coordinator** against the shipped CLI's requirements, e.g.:

- `GET https://coordinator.streamvc.live/v1/{autotune-candidates,demand-rank,rate-card}` and their `.sig` → all `200`
- All three feeds share `generated_at` + `policy_version` (pairing invariant the CLI enforces)
- All `.sig` `key_id`s are in the CLI's trusted keyring
- Coordinator `/healthz` `version` >= the version the CLI expects

This is the "does prod actually serve what we just shipped" check. A single run of it would have blocked or flagged this incident before a user hit it.

## Related

- Root incident: signed rate-card deploy skew (2026-07-30)
- #820 (build/deploy version-stamp footgun)

Issue #614, opened 2026-07-16
Program: reconcile SPEC authority, implementation, tests, journeys, and release evidence


Augustas11:
## Outcome

Track the remaining SPEC/code reconciliation program after the governance foundation lands. This issue is the program ledger, not evidence that any component fix is complete.

## Baseline and arbitration

The governance foundation establishes:

- lifecycle and requirement-ID rules;
- one authoritative spec per shared concept;
- `CODE_BUG`, `SPEC_BUG`, `DECISION_REQUIRED`, `DUPLICATE_AUTHORITY`, and `UNKNOWN` arbitration;
- tracked authority/conformance manifests;
- fail-closed repository validation.

Every domain batch must compare the current normative spec, implementation, automated tests, physical journey evidence where required, and explicit product/security/architecture decisions. Deployed behavior does not win by default.

## Remaining program

- [ ] Inventory and classify all semantic drift with an explicit verdict, owner, issue, and evidence state.
- [ ] Reconcile the #585 recovery lifecycle first (SPEC-001, 003, 008, 010, 013, 020, 023, 025, 026, 031, 032, 033).
- [ ] Reconcile billing, buyer, and policy domains second (SPEC-002, 005, 006, 015, 016, 019, 022, 024).
- [ ] Reconcile remaining canonical specs in bounded domain PRs.
- [ ] Define executable journey IDs and integrate conformance with the physical release gate.
- [ ] Make stable promotion consume current signed conformance/journey evidence.
- [ ] Establish scheduled read-only drift scans and quarterly domain reconciliation.

## Existing component issues

These are tracked independently; this umbrella does not assert their current GitHub state or completion:

- #608 — unified release-bound model identity authority
- #609 — canonical provider model-hash semantics
- #610 — disconnected/coordinator-rejected provider update recovery
- #611 — independent Malibu Team-ID fix release
- #612 — incumbent-model and autotune continuity across CLI update
- #613 — signed physical lifecycle acceptance before stable release
- #615 - expiring production recovery exceptions
- #616 - complete legacy-install compatibility-set convergence
- #617 - user-facing recovery language and identity-preserving UX

This umbrella does not absorb or close those issues. Their fixes must be reconciled through their owning specs and mapped evidence.

## Next domain

After the governance-foundation PR merges, the next implementation session should reconcile Phase 3/4 specifically for the #585 recovery lifecycle, beginning with coordinator-independent update/recovery authority and the contradictions tracked by #610 and #612.

## Completion rule

[...]

PR #835, opened 2026-07-31
fix(nginx): route /providers/ to coordinator :8444 (unbreak provider earnings $0.00)

Augustas11:
## Summary

Adds the missing `location /providers/` route to the coordinator nginx template
so the public **provider earnings endpoint is reachable**. This is the root
cause of every Malibu provider card showing **$0.00**.

## Root cause

The coordinator mounts `GET /providers/{id}/earnings` (and the SPEC-016 payout
address endpoints) on the **provider mux at `127.0.0.1:8444`**. The nginx site
template (`phase4-coordinator/dist/nginx-coordinator.streamvc.live.conf`) had
**no `/providers/` location**, so `https://coordinator.streamvc.live/providers/{id}/earnings`
returned nginx **404**. The provider earnings client resolves `coordinator_url`
into exactly that URL, so every provider's card decoded nil money fields → `$0.00`,
`Pending: <unknown>`, "waiting for the first paid job" — regardless of real credits.

Same class of nginx route-gap as the signed-feed endpoints (#822–#825).

## What changed

One `location /providers/` block proxying to `:8444`, mirroring the existing
`/admin/` block (passes `Authorization`; prefix location so longest-match wins
over `location /` — ordering is not load-bearing).

## Verified live (before this backport)

The live Pearl vhost was patched directly during incident response (backup +
`nginx -t` + reload). After the route was added:

```
GET https://coordinator.streamvc.live/providers/{id}/earnings
-> HTTP 200
   usdc_lifetime = 0.083632
   usdc_pending  = 0.083632
   total_credits = 83632
```
(was HTTP 404 before). This backport keeps the tracked template in sync so the
**next coordinator deploy does not regress the fix**.

## Governance declaration

Cites **`SPEC-022-R007`** (verified-model-settlement R-7: settlement, quarantine,
and payout exclusion — the requirement that governs **earnings visibility** and
the **earnings APIs**). This nginx route makes the provider earnings API actually
reachable, so it serves that requirement's earnings-visibility surface.
`SPEC-022-R00N` requirement IDs were registered in #834, so this cites a real
requirement and the `spec-index / check` job passes.

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-022"],
  "requirements": ["SPEC-022-R007"],
  "authority_domains": ["verified-model-settlement"],
  "arbitration": ["CODE_BUG"],
  "tests": ["not-required"],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/614"
}

### DIFF
--- a/phase4-coordinator/dist/nginx-coordinator.streamvc.live.conf
+++ b/phase4-coordinator/dist/nginx-coordinator.streamvc.live.conf
@@ -173,6 +173,22 @@ server {
         proxy_read_timeout 30s;
     }
 
+    # Provider-facing earnings/payout endpoints served by the coordinator
+    # provider mux on :8444 — GET /providers/{id}/earnings (provider-token
+    # authenticated; SPEC-016 payout-address endpoints share the /providers/
+    # mount). Without this route nginx returns 404 and EVERY provider's Malibu
+    # card shows $0.00, because the earnings client resolves coordinator_url
+    # into https://<host>/providers/{id}/earnings. Prefix location: longest
+    # match wins over `location /`, so ordering here is not load-bearing.
+    location /providers/ {
+        proxy_pass http://127.0.0.1:8444;
+        proxy_set_header Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header Authorization $http_authorization;
+        proxy_read_timeout 30s;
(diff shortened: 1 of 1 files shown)


antfleet-ops (approved):
nginx /providers/ route backport; verified live on Pearl (GET /providers/{id}/earnings 200 with usdc_* after the route). Low-risk template sync mirroring /admin/. Approving.

Issue #744, opened 2026-07-25
P2: selection quality — rank on measured revenue/hr, TTFT ceiling, catalog gate provenance and drift


Augustas11:
## Summary

Split out of #742, which was rescoped to the minimum fix (swap disqualifier). This
issue holds the larger selection-quality work, none of which is blocking.

Core problem: model selection optimizes **provider rate-per-token** with no
buyer-quality term, and the catalog data that might otherwise constrain it is not
trustworthy enough to enforce.

> **Root cause identified 2026-07-25 — see #745.** This issue originally recorded that
> catalog gate data is untrustworthy as an *observation*. #745 is the mechanism:
> `serve --model` overrides the model identity but not `model_artifact_path`, so
> autotune's candidate benchmarks load the **incumbent** model and record the result
> under the **candidate's** name. Catalog `bench_gate` values are therefore
> measurements of whatever was configured at benchmark time, not of the row they are
> attached to. Re-deriving gates is blocked on #745 landing first.

## Why the catalog can't be the enforcement mechanism

Provenance audit of `published-2026-07-10-catalog-recovery-v1`:

| row | ram | tier | tps≥ | ttft≤ | how the gate was set |
|---|---|---|---|---|---|
| qwen3-coder-30b-a3b | 28 | C | 20 | 3500 | measured, M5 32GB |
| gpt-oss-20b | 24 | C | 15 | 2500 | measured, M5 32GB |
| gemma-4-26b-a4b-it | 28 | C | 10 | 3000 | measured, M5 32GB |
| qwen3-8b | 12 | C | 15 | 4500 | measured, M5 32GB |
| llama-3.1-8b | 12 | C | 15 | 2500 | **no throughput bench** |
| llama-3.2-3b | 4 | C | 15 | 2500 | **no throughput bench** |
| qwen3-32b | 48 | B | 15 | 4000 | **never benched** ("values unchanged") |
| qwen2.5-coder-32b | 48 | A | 20 | 3500 | **policy** ("to broaden eligibility") |
| nemotron-3-nano-30b-a3b | 32 | C | 30 | 3000 | **runtime validated only** |

1. **4 of 9 measured, all on one machine.**
2. **The three rows needing >32 GB are the three that cannot be benched on the only
   executor** — high-end rows are unmeasurable by construction.
3. **Measured rows don't reproduce.** Both re-measured on 2026-07-25:

   | row | note claims | measured | error |
   |---|---|---|---|
[...]

Issue #742, opened 2026-07-25
P1: disqualify swapping models from paid recommendations (and remove the 60s TTFT default)


Augustas11:
> **Rescoped 2026-07-25.** Earlier revisions of this issue proposed hard-enforcing
> catalog `bench_gate` thresholds, then a broader selection redesign. Both were too
> large, and the first was wrong (most catalog gates are unmeasured). This issue is
> now the minimum change that prevents the incident below. The rest moved to #744.

## Summary

A model that is actively swapping to disk can be selected as a paid recommendation
and served to buyers. Nothing rejects it. Add the disqualifier.

Two changes only:

1. `swap_detected == true` disqualifies a paid recommendation.
2. Remove the 60-second TTFT feasibility default.

## Why swap, specifically

Swap is a **locally measured fact about the provider's own machine** — it needs no
catalog data, no per-model threshold, and no knowledge of the hardware. A model that
swaps is the wrong choice on any Mac, whether it is an Air or an Ultra. That makes
this safe to enforce on hardware we have never benchmarked, which is the situation
for every provider who joins from here on.

It is already computed and already thrown away:

```swift
// AutotuneRecommend.swift:2178
swapDetected = afterPageouts > beforePageouts
```

```swift
// AutotuneRecommend.swift:2883 — only appends a display flag
if safety.swapDetected || safety.thermalThrottleDetected {
    if safety.swapDetected { flags.append("swap detected") }
```

## Evidence (live, 2026-07-25)

Production coordinator: `pool_size: 1` — the provider below **is** the network.

Config in production (M5 / 32 GB, binary `v1.8.60`), and the benchmark that selected it:

```
model             = qwen3-coder-30b-a3b-instruct
sustained_tps     = 13.452081348183558
ttft_ms           = 10995
swap_detected     = true
probe_diagnostics = "feasible but swap detected"
```

`min_ram_gb: 28` passes the RAM check on a 32 GB Mac by exactly zero margin
(`AutotuneRecommend.swift:1669`, `safetyMarginGB = 4` at `:1465` → `32 - 4 = 28`),
which is why it swaps. `"feasible but swap detected"` is the system correctly
observing the problem and then proceeding anyway.

### What the disqualifier would have bought

Measured end-to-end through the buyer gateway after switching the same Mac to the
next-best row:

| | qwen3-coder-30b-a3b (swapping) | gpt-oss-20b |
|---|---|---|
| TTFT cold | 10,995 ms | **3,423 ms** |
[...]

PR #836, opened 2026-07-31
fix(autotune): gate paid eligibility on sustained memory-pressure, not any pageout (unblock 8 GB providers)

Augustas11:
## Problem: the swap gate false-rejects ~100% of genuine 8 GB providers on first install

Provider onboarding is the network's growth bottleneck, and the autotune swap gate silently blocks legitimate small-Mac providers. A real support case (8 GB M2 Air, restarted, still failed) traced to:

`ProbeSafetyAssessment.assess` computed `swapDetected = afterPageouts > beforePageouts` from `/usr/bin/vm_stat`'s **machine-wide, cumulative `Pageouts` counter**, sampled once before / once after the probe with **zero threshold**. On Apple Silicon that counter includes healthy memory *compression* and is not process-scoped, so any incidental system paging during the probe flipped the flag. `swap_detected == true` is a hard paid-eligibility veto (SPEC-023 v0.7 / #742), so 8 GB Macs fell to `no_eligible_model` → donor-only.

**Verified against production (Pearl coordinator DB, `macprovider_stats`):** the only 8 GB provider on record is `augustass-macbook-air` with `source = operator` — hand-seeded, bypassing the gate. Zero organic 8 GB nodes have ever passed it; `provider_register_attempts` shows the failing install never even reached the coordinator (fails purely client-side in the CLI).

## Fix: gate on sustained memory-pressure thrash, not any pageout

- Signal is now the kernel's own verdict, `kern.memorystatus_vm_pressure_level` (1/2/4 → Normal/Warning/Critical), read in-process via `sysctlbyname` — no subprocess, process-relevant, not compressor-conflated.
- Interval-sampled (~250 ms) across the probe **plus a synchronous sample at start and after the probe returns** (always ≥2 samples; sampler is awaited before assessment so no boundary sample is lost).
- `swap_detected == true` iff **≥2 CRITICAL samples AND CRITICAL ≥50% of the READABLE (non-unknown) samples**. Warning-majority is advisory-only (`swap_observed_under_load`, local telemetry). Fail-closed to true only when the entire series is unreadable.
- **`swap_detected` wire field, `CandidateBenchmark`, evidence schema, and the eligibility veto are unchanged** — no coordinator contract change (coordinator still rejects `swap_detected==true` evidence; confirmed at `phase4-coordinator/internal/autotune/gate.go`).
- #742's protection is preserved and its regression fixture made meaningful (was `pageouts 10→11`, a delta of 1): a genuinely thrashing 32 GB node still fails, including short probes.
- Local calibration telemetry (`~/.cache/macprovider/autotune-logs/probe-safety.log`, rotates at ~5 MiB) records the pressure series + TPS/TTFT per probe.

SPEC-023 → **v0.9.0**; CONFORMANCE + generated index synced. `swift test --filter Autotune` → 257 tests, 0 failures (incl. 8 ProbeSafety cases and the #742 fixture).

## Audit: two rounds, three codex lanes each (code / security / architect)

**Round 1 →** fail-open on <3 samples + `.unknown` dilution (MEDIUM), unbounded log (LOW), spec-wording (LOW) — **all fixed** (final synchronous sample + awaited cancel; readable denominator + ≥2-critical rule; log rotation; SPEC/AC-12 synced; +3 tests).

**Round 2 →** CODE clean; ARCHITECT 0 C/H/M (2 doc-comment LOWs, fixed); SECURITY 0 C/H, one MEDIUM analyzed below.

### Carried findings (explicit)
- **[Accepted tradeoff — operator-approved] Buyer-QoS floor (HIGH/MED):** the fix intentionally admits functional low-throughput 8 GB nodes (the growth goal). The only buyer-quality backstop, `--buyer-ttft-ceiling-ms`, defaults to `0`/disabled — **pre-existing** (#744), not introduced here. A default TTFT floor can't be calibrated until real 8 GB throughput telemetry exists, which this fix is what produces. **Decision: ship growth fix now, buyer-QoS floor as a data-driven follow-up.**
- **[Non-regression] Stale swap evidence (MEDIUM):** cached `swap_detected=false` evidence from old semantics can be honored within the existing 7-day freshness window. Analyzed safe-direction: old `false` required *zero* pageouts (a strictly stronger condition than "not sustained-CRITICAL"), so honoring it cannot admit a node the new rule would block; old `true` is coordinator-rejected and re-probes on upgrade. Optional follow-up: gate swap-evidence freshness on binary version.
- **[Follow-up] Fleet calibration (LOW):** telemetry is local-only; wiring it to the coordinator for fleet-level threshold calibration is deferred.

{
  "behavior_change": "yes",
  "contract_change": "yes",
  "specs": ["SPEC-023"],
  "authority_domains": ["installer-autotune-policy"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "phase3-binary/Tests/macprovider-cliTests/AutotuneRecommendTests.swift"
  ],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/742"
}


antfleet-ops (approved):
Approved: swap gate root-caused (vm_stat Pageouts machine-wide false-positive on 8GB), fix verified against Pearl DB, 2 rounds x 3 codex lanes to 0 C/H/M, #742 protection preserved, wire schema unchanged. Carried follow-ups documented.

PR #841, opened 2026-07-31
[DO NOT MERGE] fix(release): generalize Malibu download promotion off the v1.8.39 freeze

Augustas11:
> ⚠️ **DO NOT MERGE YET.** One item remains: a staging publish to validate the signature-enforced remote path (the 3 release-security test edits are now applied — test green). This generalizes signature-enforced, install-breaking-if-wrong release infra that cannot be exercised in CI — it needs one staging publish to validate the remote path.

## Problem

The Malibu download promotion (`download.malibu.tech`) was hard-locked to the one-time **v1.8.39** Sparkle bootstrap in six places (the release.yml publish gate, `publish-malibu-latest-dmg.sh`, `verify-malibu-publication-set.sh` ×2, the remote root `install-malibu-publication.sh`, `verify-malibu-bootstrap-publication.sh`). Every release since 1.8.39 shipped a Malibu.dmg to GitHub but was **never promotable** to the download — the tooling refuses any other tag. Result: `download.malibu.tech` has been frozen at **v1.8.53** since Jul 20, and each new release silently lags. (The Sparkle *appcast* is a separate legacy bridge and is intentionally frozen — not the update path; Malibu users update via the coordinator CLI-autoupdate.)

## Change

Generalize the download **promotion** so any stable `vX.Y.Z` can be promoted, without disturbing the frozen Sparkle bridge:

- **Lift the six `v1.8.39` tag-locks** (tag-equality only; every signature/checksum/mode/symlink-safety check preserved).
- **Ship the committed frozen bridge appcast** (`scripts/dist/malibu-frozen-bridge-appcast.xml`, sha-pinned in the remote helper) for non-v1.8.39 promotions so the `/appcast.xml → .malibu-current/appcast.xml` symlink keeps resolving. Appcast **generation** stays frozen.
- **Wire the signed acceptance-candidate** (`acceptance-candidate.yml` artifact) into publish + the remote helper so it lands in the promoted release dir under the same atomic/immutable checks — and **validate it before any Pearl transfer** (ECDSA P-256 signature, schema/channel, `checksums.txt` binding, 5m–24h non-expiry, and `tag`/`candidate_commit`/`compatibility_set_id` pinned to this exact release). A stale/expired/mismatched/dummy pair now dies **before** the immutable release dir can pin it, instead of bricking installs at die-4 with no in-place repair.
- **Per-release cache-bust key** (DMG sha, not the constant frozen-appcast sha) so CDNs can't serve a stale `latest.dmg`.
- New negative test `test-malibu-acceptance-candidate-prepublish.sh` (valid passes; tampered/expired/mismatched rejected pre-transfer), wired into `make test-dist`.

## Audit

Three-lane codex (code/security/architect), two rounds → **0 CRITICAL / HIGH / MEDIUM**.
- R1 HIGH "removes the live coordinator release gate" was a **stale-base false positive** (origin/main added that gate in `b0b7289a` after branch point); rebased away.
- R1 HIGH (real): acceptance-candidate placed without validation → fixed (pre-promotion ECDSA validation, above).
- R1 MEDIUM (real): constant cache key → fixed (per-release DMG sha).
- R2: both fixes verified to hold. Residual **LOW** (carried): the remote root helper doesn't independently re-validate the candidate — defense-in-depth only; the normal publish path validates before SSH/transfer, so the helper is never reached with an unvalidated pair.

## Before merge (human-gated)

1. ~~3 assertion updates in `scripts/test-release-security-posture.sh`~~ — **DONE** (user-authorized; step-name renames + gate assertion `prerelease == 'false'` + frozen-appcast/acceptance-candidate asserts; `test-release-security-posture.sh` passes).
2. **One staging publish** to validate the signature-enforced remote path end-to-end (CI can't). ← the only remaining gate.

## Follow-up (not in this PR)

- **Auto-promote post-acceptance:** wire the download promotion into `promote-acceptance-candidate.yml` so every accepted release promotes automatically — otherwise it stays a manual step and keeps lagging. This is the deeper "why it's broken each release" fix.
- **v1.8.69 (the swap-fix release) cannot be promoted** — its acceptance-candidate can't be minted post-tag ("candidate tag already exists"). Not needed: users already get the fix via coordinator CLI-autoupdate. It rides the next release cut with the acceptance flow run pre-tag.

## Governance

Infra/CI + release-tooling PR (no authority domain governs CI/release scripts), so there is no honest spec-governance declaration; `spec-index / check` is advisory and non-required. No behavior change to any SPEC contract.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1916,8 +1916,8 @@ jobs:
             --require-immutable \
             --openssl "$OPENSSL_BIN" >/dev/null
 
-      - name: Publish one-time Malibu 1.8.32 bootstrap bridge to Pearl
-        if: needs.build.outputs.tag == 'v1.8.39' && needs.build.outputs.prerelease == 'false'
+      - name: Publish Malibu download bridge to Pearl
+        if: needs.build.outputs.prerelease == 'false'
         env:
           MALIBU_DOWNLOAD_VPS_HOST: ${{ secrets.MALIBU_DOWNLOAD_VPS_HOST }}
           MALIBU_DOWNLOAD_SSH_KEY: ${{ secrets.MALIBU_DOWNLOAD_SSH_KEY }}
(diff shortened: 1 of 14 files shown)


Augustas11:
Closing: this generalizes the retired download.malibu.tech/latest.dmg promotion. Per SPEC-025 v0.22 the Download-for-Mac authority is the immutable versioned GitHub release asset (malibu.tech/host already links it); the real 'new downloads lag' fix is advancing the landing-page pin (MalibuAI/malibu j/release.mjs), not this mechanism. Audited work preserved on branch for reference if download.malibu.tech is ever revived.

PR #848, opened 2026-07-31
Harden coordinator release tag deploy provenance

### LINKED ISSUE #820: Coordinator build/deploy version-stamps from git describe — building from main ships a wrong/regressing version the deploy guard can't catch
## Problem

Coordinator release binaries are version-stamped from `git describe`, and building/deploying from `main` (instead of a release tag) produces a **wrong, even regressing** version that the deploy's own guard does **not** catch.

### Evidence (2026-07-30)

- `phase4-coordinator/scripts/build-linux.sh` stamps the version via `git describe` (`-ldflags`).
- Building from `main` HEAD:
  - `git describe --tags HEAD` → `v1.8.46-2-g73da065f`
  - `build-linux.sh` → `built dist/coordinator-linux-amd64 @ v1.8.46-1-g...`
- The release tags (`v1.8.6x`) are **not** the nearest ancestor tag on `main` (tag topology is non-monotonic with history), so `describe` picks the stale `v1.8.46` tag. The resulting coordinator would advertise `v1.8.46-…` — **below** the deployed prod coordinator (`v1.8.66`).
- Building from the **`v1.8.67` tag** in a clean worktree stamps correctly: `built … @ v1.8.67`.

### Why the deploy guard misses it

`phase4-coordinator/dist/deploy-pearl-vps.sh` computes:

```
EXPECTED_VERSION=$(git describe --always --dirty --tags 2>/dev/null || git rev-parse --short HEAD)
```

…from the **same checkout** it runs in, then compares that to the post-deploy `/healthz` `version`. Because both the binary stamp and `EXPECTED_VERSION` derive from the *same* `git describe`, they are **self-consistent** even when both are wrong. So a build+deploy from `main` passes the version-cohesion check while shipping a mis-/down-versioned coordinator to prod. `STRICT_PROVENANCE=1` only checks that `/healthz` *has* a version field, not that it is a real release tag.

## Impact

A careless "build from main and deploy" ships a prod money-path coordinator whose advertised version is a dev/`-dirty` string that can regress below the currently-deployed release, corrupting `recommended_binary_version` / version-cohesion logic — and the deploy's own guard will not stop it.

Surfaced while deploying #813 (signed rate-card) to fix the fresh-install `catalog_trust_blocked` outage: the correct source is the `v1.8.67` tag, not `main`.

## Proposed fix

Make production builds/deploys refuse anything that isn't a clean release tag:

1. `build-linux.sh`: fail (or require an explicit override) when `git describe --tags` is not an exact tag match (i.e., has a `-N-g<hash>` or `-dirty` suffix), so prod binaries can only be stamped from a tagged commit.
2. `deploy-pearl-vps.sh`: derive/validate `EXPECTED_VERSION` against an exact release tag (reject `-g`/`-dirty` describe output for production), independent of the build, so the version-cohesion guard catches a non-tag build instead of rubber-stamping it.
3. Optionally: assert the deployed version is `>=` the currently-advertised `latest_binary_version` to prevent regressions.

## Not in scope

This issue is only the build/deploy version-stamping footgun. The signed-rate-card deploy itself (#813 → Pearl) is handled separately.

Augustas11:
## Summary

Fixes #820 by making production coordinator build/deploy provenance fail closed on anything except the exact signed release tag.

- Require exactly one annotated `vX.Y.Z` tag at `HEAD`, verified by the authorized release signer and canonical upstream tag target.
- Build production coordinator artifacts from a git archive of the verified release commit, while keeping `ALLOW_NON_RELEASE_COORDINATOR_BUILD=1` for local/dev artifacts.
- Make Pearl deploy preflight the same verified commit, pin tracked deploy inputs from that commit, force GitHub release lookup to `github.com`, and checksum recovery/TCP/main staging before remote mutation.
- Update deploy/runbook docs and add a regression test covering main/ahead-of-tag rejection, unsigned/lightweight/multiple tag rejection, signer spoof rejection, dirty checkout rejection, fatal `/healthz.version` mismatch, and deploy input pinning.

## Validation

- `bash phase4-coordinator/dist/test/coordinator_release_tag_guard.test.sh`
- `shellcheck --severity=warning phase4-coordinator/scripts/build-linux.sh phase4-coordinator/dist/test/coordinator_release_tag_guard.test.sh`
- `bash -n phase4-coordinator/scripts/build-linux.sh phase4-coordinator/dist/deploy-pearl-vps.sh`
- `bash phase4-coordinator/dist/test/check_stats_inventory_deploy_test.sh`
- `bash phase4-coordinator/dist/test/check_stats_billing_mirror_deploy_test.sh`
- `bash phase4-coordinator/dist/test/check_pearl_tcp_test.sh`
- `make vet-coordinator`
- `make test-coordinator`
- `make test-dist`
- `git diff --cached --check`
- 3-lane audits: `CODE AUDIT: 0 C/H/M findings`, `SECURITY AUDIT: 0 C/H/M findings`, `ARCH AUDIT: 0 C/H/M findings`

Not live-tested: Pearl production deploy and real GitHub release download through the deploy wrapper.

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-020"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "phase4-coordinator/dist/test/coordinator_release_tag_guard.test.sh",
    "make test-dist"
  ],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/820"
}

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,6 +95,7 @@ test-dist:
 	bash phase4-coordinator/dist/test/check_stats_inventory_deploy_test.sh
 	bash phase4-coordinator/dist/test/check_stats_billing_mirror_deploy_test.sh
 	bash phase4-coordinator/dist/test/coord_deploy_config_mode_test.sh
+	bash phase4-coordinator/dist/test/coordinator_release_tag_guard.test.sh
 	bash phase4-coordinator/dist/test/check_deploy_static_feed_access.test.sh
 	bash phase4-coordinator/dist/test/coordinator_deploy_recovery.test.sh
 	bash phase4-coordinator/dist/test/coord_deploy_smoke_probe.test.sh
--- a/ops/runbooks/malibu-pearl-deploy.md
+++ b/ops/runbooks/malibu-pearl-deploy.md
@@ -13,7 +13,7 @@
 | Changes | Does NOT change |
 |---------|-----------------|
 | Applies Postgres migrations `012_malibu_emission_ledger` + `014_malibu_trust_unlock` | `malibu_emission.enabled` (stays `false`) |
-| Installs coordinator binary with `writer_dsn` wired | Withdrawable MALIBU flow |
+| Validates/restarts signed coordinator runtime with `writer_dsn` wired | Withdrawable MALIBU flow |
 | Merges MALIBU overlay into Pearl `coordinator.pearl-overlays.yaml` | Base `/opt/macprovider/coordinator.yaml` |
 | Mounts `GET /v1/provider/malibu-accrual` read API | Base nginx vhost (add `location = /v1/provider/malibu-accrual` per §4.4) |
 
@@ -52,6 +52,7 @@ After first deploy with `MALIBU_EMISSION_WRITER_PASSWORD` set, rotate the passwo
 ## 2. Quick deploy (scripted)
 
 ```bash
+git checkout vX.Y.Z
 cd phase4-coordinator
 bash scripts/build-linux.sh
 bash dist/deploy-malibu-emission-pearl.sh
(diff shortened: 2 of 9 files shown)


antfleet-ops (approved):
Reviewed full #820 release provenance guard diff; approving after local 3-lane audit reached 0 C/H/M and validation passed.

PR #856, opened 2026-08-01
Bind Pearl runtime verifier to release source

Augustas11:
## Summary
- Pass `steps.release_source.outputs.commit` as `EXPECTED_REVISION` when the Pearl runtime workflow calls `verify-pearl-go-binaries.py`.
- Add a targeted workflow-shape assertion so runtime release publication cannot silently drop the reviewed-commit binding again.

## Why
The protected Pearl runtime workflow for `v1.8.70` failed before publication because the Go binary verifier now requires an explicit reviewed source revision. This preserves the intended provenance gate instead of bypassing it.

## Validation
- `bash scripts/test-pearl-runtime-release.sh`
- `python3 scripts/verify-pearl-go-binaries.py --self-test`
- `git diff --check`

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-020"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "bash scripts/test-pearl-runtime-release.sh",
    "python3 scripts/verify-pearl-go-binaries.py --self-test",
    "git diff --check"
  ],
  "journeys": ["JOURNEY-PROVIDER-AUTOUPDATE-PHYSICAL"],
  "issue": "https://github.com/Augustas11/macprovider/issues/825"
}

### DIFF
--- a/.github/workflows/pearl-runtime-release.yml
+++ b/.github/workflows/pearl-runtime-release.yml
@@ -117,7 +117,8 @@ jobs:
           strings gateway-linux-amd64 > gateway-linux-amd64.strings
           grep -Fq "$tag" coordinator-linux-amd64.strings
           grep -Fq "$tag" gateway-linux-amd64.strings
-          python3 scripts/verify-pearl-go-binaries.py \
+          EXPECTED_REVISION="${{ steps.release_source.outputs.commit }}" \
+            python3 scripts/verify-pearl-go-binaries.py \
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Approved for #825 runtime release workflow repair. Scope is limited to binding the existing Go binary verifier to the reviewed release commit and adding the regression assertion.

PR #876, opened 2026-08-02
Bump version to 1.8.79 for CLI release (keep-awake fix for fleet autoupdate)

Augustas11:
### Why
The client keep-awake fix (`77d4b50b`, 2026-08-01) — binds the macOS sleep
assertion to serving intent so App Nap / reconnect backoff can't freeze a
serving Mac's heartbeat — was **merged to main but never published as a macOS
provider release**. Every release since v1.8.69 (v1.8.70..v1.8.78) is a Linux
**Pearl-runtime** coordinator prerelease with **no macOS CLI / Malibu /
compatibility-set artifact**. So providers on v1.8.69 are stranded on the
pre-fix client.

**Observed live (Pearl coordinator, 2026-08-02):** provider "MacBook Air — Влад"
(`mp-038459680a…`, joined 2026-07-31, on v1.8.69) flapped **81 disconnects/6h** —
heartbeats stop, coordinator's 60s read deadline expires (`i/o timeout`), drop,
reconnect, repeat = his reported "works 1–2 min, then searching for a network."
Isolated to his node; zero coordinator-side rejections. Signature = App-Nap
heartbeat starvation, exactly what `77d4b50b` fixes.

### What
Version-bump only (mirrors the v1.8.69 bump `5adef036`): `binaryVersion`
1.8.69 → 1.8.79 and the `coordinator_advertised_version.latest_binary_version`
templates to match. `required_binary_version` (hard admission floor) unchanged.

The compatibility-set manifest schema is **byte-identical** to v1.8.69
(`scripts/compatibility-set-manifest.py` unchanged across the tags; 9-file
`CATALOG_FILES`), so v1.8.69 (exact-9) clients cross to v1.8.79 with **no
bridge**. The ≤1.8.66 (exact-7) forward-incompat trap is a separate older-fleet
concern and is out of scope here.

### Rollout sequence (post-merge)
1. Tag `v1.8.79` at the merge commit, push tag.
2. Dispatch `release.yml` (version `v1.8.79`) as a **prerelease canary** —
   builds/signs/notarizes CLI + Malibu, publishes compatibility-set. Requires
   `production-release` environment approval.
3. Advance Pearl live `recommended_binary_version` 1.8.69 → 1.8.79 + reload
   coordinator (watched) → v1.8.69 clients autoupdate via the coordinator source.
4. Confirm the fleet adopts (Vlad's disconnect count drops), then promote the
   canary to stable/latest (the live-recommendation release gate passes once
   Pearl advertises 1.8.79).

[...]

PR #875, opened 2026-08-02
Remove static Pearl canary fleet baseline

Augustas11:
## Summary
- removes static provider-count and static model-list gating from ordinary scheduled Pearl liveness
- derives liveness protected providers, model workload, ready floor, and bounded request/token budget from authenticated live Ready/routable `/poolz`
- preserves strict static behavior for qualification and explicit rollback authorization
- captures provider/model tuples in the Pearl updater journal and requires normal forward proof plus rollback authorization to match captured provider/model identity
- repins the updater canary authority to the reviewed live-fleet canary runtime

## Root Cause
The scheduled liveness canary still carried rollout-era static assumptions: fixed `CANARY_MODELS`, fixed minimum ready provider count, and a four-request budget. That made Pearl recovery/deploy proof fail when the live provider fleet legitimately changed after a provider dropped or joined.

## Validation
- `python3 ops/pearl-updater/test_pearl_updater.py` -> 187 OK, 1 skipped
- `node --test test/e2e/canary-buyer/probe.test.mjs test/e2e/canary-buyer/safety.test.mjs` -> 48/48 pass
- `bash test/e2e/canary-buyer/run-canary.test.sh` -> PASS
- `bash scripts/test-pearl-runtime-release.sh` -> PASS
- `python3 scripts/verify-pearl-go-binaries.py --self-test` -> PASS
- `python3 -m py_compile ops/pearl-updater/macprovider-pearl-update ops/pearl-updater/test_pearl_updater.py` -> PASS
- `node --check test/e2e/canary-buyer/probe.mjs && node --check test/e2e/canary-buyer/safety.mjs` -> PASS
- authority SHA-256 pins matched all six canary runtime files
- `git diff --check origin/main..HEAD` -> clean

## Audit
- Code lane: C/H/M 0/0/0
- Architecture lane: C/H/M 0/0/0
- Security lane: initially found one Medium on forward model drift; fixed in `9b799ef4`; focused re-audit C/H/M 0/0/0

## Deploy Note
Pearl has prior manually-edited updater state from the abandoned transaction. After merge, reconcile the updater with the reviewed binary/config state before running another updater apply so it does not fight the manual gateway deployment.

PR #879, opened 2026-08-02
Fix main-red: repoint canary authority pin to reachable merged commit

Augustas11:
### Problem
`ci-required` is **red on main HEAD** for every PR. The failing test is
`test_canary_rollout_authority_hashes_match_issue_825_duplicate_fleet_runtime`,
which runs `git cat-file -e ${CANARY_AUTHORITY_COMMIT}^{commit}` and fails 128
in a clean clone.

`CANARY_AUTHORITY_COMMIT` was pinned to
`3e5627767ace28fa530df47318efd873a1aaf070` — the **tip of the pre-squash
`fix/825-remove-static-canary-fleet` branch** (#875). After that PR
squash-merged, main's commit is `fb3f4f16` and `3e562776` is orphaned on the
now-deleted branch, so a fresh CI checkout of main has no such object. It
passes on a developer clone that still has the branch object, and fails in CI —
which is why #875's own PR CI was green but main went red post-merge.

This is the **6th** instance of the same self-referential-pin defect
(`CANARY_AUTHORITY_VERSION = issue-825-canary-fleet-r6`): each re-pin points at
the commit-being-merged, whose squash SHA can't be known before merge.

### Fix
Repoint the pin — and its two mirror references (the test's self-consistency
assertion + the runbook) — to **`fb3f4f16`**, the actual merged commit carrying
the identical reviewed `probe.mjs`/`safety.mjs` bytes. `fb3f4f16` is a permanent
ancestor of main, so the pin stays resolvable across future squash-merges.

- Probe/safety **content is unchanged**; `CANARY_AUTHORITY_FILES` hashes are
  untouched — only the commit reference is corrected.
- `test_canary_rollout_authority_hashes_match_issue_825_duplicate_fleet_runtime`
  passes locally with the repoint.

Unblocks `ci-required` repo-wide (including the v1.8.79 provider-CLI release
PR #876).

### Governance declaration
Non-governance CI/ops path (pearl-updater + test + runbook); no canonical SPEC
requirement governs a CI-pin correction — the documented "no honest declaration"
case. Advisory `spec-index / check` is not a merge requirement.


antfleet-ops (approved):
Approving: repoints the canary authority pin from the orphaned pre-squash branch commit to the reachable merged commit fb3f4f16. Content unchanged; unblocks main-wide ci-required. Failing canary test now green.

PR #947, opened 2026-08-07
fix(rewards): stop lib/pq wire-protocol desync in cap-replay accrual path (mp-* accrual)

Augustas11:
## Summary
- Root cause of residual `mp-*` accrual failures on Pearl: `pq: unexpected Parse response "(C) CommandComplete"`.
- `replayCapPending` (cap-hold replay for Trusted providers) issued an `UPDATE` on the same transaction while its own `SELECT ... FOR UPDATE` `*sql.Rows` result set was still open. `lib/pq` doesn't buffer results like `libpq` does, so a new Parse/Bind/Execute on the same connection while a Rows set is unread desyncs the wire protocol.
- This path only fires when a Trusted provider's bound wallet has `cap_replay_pending=true` — set by the SPEC-016 payout-address wallet mirror whenever a payout address binds/changes. `mp-*` (hardware/CLI track) providers exercise that path routinely; `p_*` (App-track) providers rarely do, which is why only `mp-*` accrual ticks hit it. The whole tick transaction rolled back silently (`runEmissionTick` only logs a `Warn`), dropping that provider's accrual for the tick.
- Fix: fully drain and close the `Rows` before issuing any further statement on the same `tx` — standard `database/sql` + `lib/pq` usage.
- Also wires `internal/rewards` into `make test-coordinator-integration` (it was previously omitted, so the package's `testcontainers`-backed Postgres tests never ran in CI).

## Test plan
- [x] `go build ./...`, `go vet ./...` — clean
- [x] `go test ./...` (phase4-coordinator, non-integration) — all pass
- [x] `go vet -tags=integration ./internal/rewards/...` — compiles
- [x] New regression test `TestCapReplayPendingDoesNotDesyncConnection` reproduces the exact `mp-*` scenario (Trusted provider, bound wallet, `cap_replay_pending=true`, pre-existing held ledger row) and asserts the tick transaction commits (new ledger row exists) and the held row's hold is correctly replayed/cleared. Runs in CI via `coordinator-stats-integration` (testcontainers Postgres, no Docker available in this sandbox to run it locally).
- [ ] Pearl: confirm next `mp-*` emission tick has zero `unexpected Parse response` errors in coordinator logs after deploy.

Not armed: canary untouched. Pearl config untouched — this is a pure code fix in `phase4-coordinator/internal/rewards`.

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-017", "SPEC-021"],
  "requirements": ["SPEC-017-R001"],
  "authority_domains": ["network-stats", "malibu-emission-ledger"],
  "arbitration": ["CODE_BUG"],
  "tests": ["phase4-coordinator/internal/rewards"],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/614"
}

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ test-coordinator:
 # Each stats case owns an isolated Postgres container; keep the package
 # deadline above the observed hosted-runner setup/teardown envelope.
 test-coordinator-integration:
-	cd phase4-coordinator && go test -tags=integration -timeout 10m ./internal/stats/... ./internal/onboarding/... ./cmd/coordinator/...
+	cd phase4-coordinator && go test -tags=integration -timeout 10m ./internal/stats/... ./internal/onboarding/... ./internal/rewards/... ./cmd/coordinator/...
 
 # SPEC-017 AC-16 — golangci-lint with depguard + forbidigo.
 # Pinned version so the target is hermetic on a fresh checkout.
--- a/phase4-coordinator/internal/rewards/emission.go
+++ b/phase4-coordinator/internal/rewards/emission.go
@@ -248,9 +248,22 @@ func bumpWalletDaily(ctx context.Context, tx *sql.Tx, wallet string, day time.Ti
 	return err
 }
 
+type pendingReplayRow struct {
+	id   int64
+	amt  float64
+	tier string
+}
+
 func replayCapPending(ctx context.Context, tx *sql.Tx, wallet string, walletCap float64) error {
 	// Oldest-first replay: for trusted providers, clear per_wallet_daily_cap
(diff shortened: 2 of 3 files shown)


antfleet-ops (approved):
CI green (all checks pass, including the previously-flaky spec-015-acceptance job on rerun). Root-caused mp-* accrual failure to a lib/pq wire-protocol desync in replayCapPending (Exec on the same tx while its own Rows was still open) — fix matches the standard database/sql + lib/pq usage pattern, and the new regression test exercises the exact scenario against real Postgres via testcontainers.

PR #988, opened 2026-08-10
fix(release): tolerate legacy Malibu publication sequence

Augustas11:
## Summary
- Make the Malibu publication installer tolerate a legacy current manifest that is missing `release_sequence` when the candidate's semantic tag advances.
- Keep rollback protection: same/older semantic tags without a valid current sequence still fail with `publication sequence did not advance`.
- Add a regression for the v1.8.93 production failure where the current v1.8.88 manual recovery manifest lacked `release_sequence`.

## Root cause
The v1.8.93 release made the GitHub release public, then failed while publishing the Malibu download alias because Pearl's current v1.8.88 manual recovery manifest did not include `release_sequence`. The helper only special-cased the old v1.8.39 bridge and treated any later sequence-less current manifest as invalid, so a stale/manual recovery record could block the next valid Malibu publication.

## Verification
- `bash scripts/test-malibu-bootstrap-bridge.sh`
- `git diff --check`

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-020"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "scripts/test-malibu-bootstrap-bridge.sh",
    "scripts/test-malibu-independent-release.sh",
    "scripts/test-release-security-posture.sh"
  ],
  "journeys": ["not-required"]
}

### DIFF
--- a/scripts/install-malibu-publication.sh
+++ b/scripts/install-malibu-publication.sh
@@ -182,6 +182,18 @@ import sys
 
 candidate = json.loads(pathlib.Path(sys.argv[1]).read_text(encoding="utf-8"))
 current = json.loads(pathlib.Path(sys.argv[2]).read_text(encoding="utf-8"))
+def tag_version(row, label):
+    tag = row.get("tag")
+    if not isinstance(tag, str) or not tag.startswith("v"):
+        raise SystemExit(f"{label} publication tag is invalid")
+    try:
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Approved: fixes the sequence-less v1.8.88 Malibu publication-manifest failure mode with rollback-preserving regression coverage. CI must pass before merge.

PR #1048, opened 2026-08-18
Let discovery renewal create protected transport tags

Augustas11:
## Summary

- Use the protected release token for the renewal workflow publish step that creates the append-only `release-discovery-v1-*` transport tag.
- Keep top-level workflow permissions read-only and keep the anonymous public verifier free of protected secrets.
- Extend the renewal workflow structural test so this credential boundary stays locked.

## Root Cause

Run `32093708961` signed and verified a renewed discovery head, then failed creating the new transport release with `HTTP 403: Resource not accessible by integration`. The renewal path creates a brand-new protected tag, unlike the numeric release path that publishes an already-created tag.

## Validation

- `bash scripts/test-renew-release-discovery-head.sh`

## Governance Note

This is an infra/release-workflow-only change under `.github/workflows`. Per repo guidance, there is no honest SPEC governance declaration for this path; the advisory `spec-index / check` may be red and is not a required merge check.

### DIFF
--- a/.github/workflows/renew-release-discovery-head.yml
+++ b/.github/workflows/renew-release-discovery-head.yml
@@ -314,7 +314,7 @@ jobs:
         id: publish
         shell: bash
         env:
-          GH_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ secrets.RELEASE_POSTURE_TOKEN }}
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Approved scoped release-discovery renewal credential fix. Local guardrail: bash scripts/test-renew-release-discovery-head.sh.

PR #1072, opened 2026-08-18
SPEC-042 Stage E: pool_id into settlement route-snapshot digest + request_log (money-path)

Augustas11:
## SPEC-042 Stage E — `pool_id` into the settlement record spine (money-path)

**Stacked on #1069** (`feat/spec-042-impl-r002-r005`). Review that first; this diff is only the record-persistence layer on top of it.

Threads the selected `pool_id` (R002) into the durable money-path records so a settled receipt is provably bound to the pool it was routed within (R006), and pool membership becomes queryable after the fact — **without changing any global (poolless) behavior byte-for-byte**.

### What this adds
- **`route_snapshot` canonical digest** now carries `pool_id` — but *only when non-empty*, so the global digest is unchanged. Critically, `pool_id` is **recovered from `route_snapshot_json` before the settlement-time digest recompute** (settlement recomputes the digest from the stored row and compares it to the stored/receipt digest; this mirrors the existing `*_algorithm` field precedent). A loader round-trip test pins insert-digest == recompute-digest == receipt-digest.
- **`settlement_route_snapshots.pool_id`** column (nullable) + ALTER migration — queryable settlement binding.
- **`request_log.pool_id`** column (nullable) + ALTER migration — queryable request-time pool.

### Isolation invariant (unchanged from #1069, retested here)
All by-hand routing paths re-check membership against a consistent per-attempt snapshot; the generation fence at the top of the `forwardWithFailover` dispatch loop rejects stale membership (`pool_state_stale`, HTTP 409) before *any* dispatch — covering retry/failover, not just initial selection.

### Compatibility
- Poolless traffic: `pool_id` empty → digest identical, columns NULL. Feature-gated on `WithPoolMembership`.
- Coordinator-only (phase4). No gateway/provider changes.

### Verification
- Affected suites green: `billing`, `buyer`, `routing`, `requestlog`, `trustpool`.
- **Codex 3-lane money-path audit → 0 Critical / 0 High / 0 Medium** (code APPROVE, security PASS, architect PASS). Round-1 found a real fence-scope defect (initial-dispatch-only) and a header-auth gap — both fixed and re-audited clean. Round-2 LOWs closed (`pool_state_stale` 503→409 per the SPEC R010 table; nil-state cleanup).

### Deferred (documented, not defects)
[...]

PR #1069, opened 2026-08-18
SPEC-042 R002/R005: pool_id threading + fail-closed tenant isolation (slice 1)

Augustas11:
## SPEC-042 R002/R005 implementation — pool_id threading + fail-closed tenant isolation (slice 1)

**Draft. Stacked on `feat/spec-042-pool-manifest` (PR #1065)** — review that skeleton first; this PR's diff is implementation only. Rebase onto `main` once #1065 merges. Tracks #1053.

First implementation slice of SPEC-042: the two deploy-blocking primitives — a `pool_id` authority field (R002) and fail-closed tenant isolation (R005) — built tests-first. Feature-gated on `WithPoolMembership`; **global (poolless) routing is byte-identical** (every gate is inert unless the registry is wired and a pool is selected).

### What's implemented (all tested, `buyer`/`routing`/`trustpool`/`billing` green)
- **`trustpool.Registry`** — Trusted Pool membership + durable revocation blocklist, with consistent `(members, generation)` snapshots (race-tested; SPEC-042 R003 single-read).
- **Filter-level isolation gate** (`routing.EligibilityChecker.ProviderInPool`, `ReasonPoolNotMember`) placed first among property gates — covers the ordinary/failover/sticky/preflight paths.
- **`pool_id` threading** — `WithPoolMembership` Server option; `chatRequest.poolID` from the authorized `X-MacProvider-Pool` header (ignored when the registry is unset); `forwardState` carries `poolID`/`poolMembers`/`poolGeneration`; one consistent snapshot per selection attempt.
- **Isolation on every by-hand dispatch path** — pinned/self-route, slot-queue admission, and slot-queue poll each re-apply the membership gate (the same paths the #768 and receipt-key gates had to touch).
- **Generation fence** (R005) for the select→dispatch window — `pool_state_stale` (retryable) forces fresh selection if membership changed.
- Error codes `pool_no_eligible_member` (fail-closed) / `pool_state_stale` (retryable) classified and inventoried.

### Isolation invariant (proven by `spec042_pool_isolation_test.go` + `filter_test.go`)
A `pool_id=P` request is **never** dispatched to a non-member, another pool's member, or global supply; fails closed when P has no member; a revoked member is rejected at T+ε (gen-keyed, no staleness window); no-pool requests are served byte-identical.

### Deliberately deferred (next slices — money-path, need the codex audit loop)
[...]

PR #1065, opened 2026-08-18
SPEC-042: Pool Control Plane and Trusted-Pool Manifest (skeleton)

Augustas11:
## SPEC-042 — Pool Control Plane and Trusted-Pool Manifest (skeleton)

New Layer 2 spec: a **Trusted Pool** as a creator-operated trust boundary over the shared MacProvider network, addressed by a mandatory `pool_id` and governed by a signed manifest. This is **administrative trust, not cryptographic privacy** — it MUST NOT be marketed as "Privacy Pools" (that is Layer 3 / SPEC-041). Draft skeleton (v0.0.7, `Maturity: skeleton`), not build-ready; the v0.1 detailed-design deliverables are gated in R010.

Tracks #1053.

### Why now
`pool_id` appears zero times in `phase4-coordinator/internal` and `phase5-gateway/internal`; routing begins from the global provider snapshot. There is no pool object today, so Layer 2 traffic is deploy-blocked, not merely incomplete. Live prod (`coordinator.malibu.tech` v1.8.95, 2026-08-18) runs `require_attestation`/`require_encrypted_leg` **fail-open**, so a Trusted Pool cannot inherit its trust floor from the network and must carry fail-closed pool-level predicates.

### What's in it (12 requirements)
- **R001** signed manifest with a **stable identity core** (`pool_id` = digest of it) separated from a **versioned policy core**; monotonic version + `prev_manifest_core_hash` chain; rollback rejection; half-open active-policy windows.
- **R002** `pool_id` as a first-class authority field + credential-bound selection authorization (narrow-only route params).
- **R003** durable per-pool membership + identity-keyed revocation; gen-keyed eligibility (revocation blocks new dispatch at T+ε); `revoke_graceful`/`revoke_immediate`.
- **R004** fail-closed pool predicates overriding the network default; exact SPEC-008 attestation-tier lattice; `hardware_mda` gated behind a SPEC-008 amendment.
- **R005** per-attempt tenant isolation with a generation fence on **every** dispatch path; no spill to global.
- **R006** labels-only settlement bound to the route snapshot; creator revenue-split deferred to a SPEC-005/016-amended V1.
- **R007** field-level retention matrix + credential-header denylist.
- **R008** buyer-verifiable signed `pool_policy.json` + coordinator-attested `pool_status.json`; mandatory `confidentiality_scope` true-negative disclosure.
- **R009** Layer 3 compat fields (privacy blocked until SPEC-041/040 amended); machine-enforced claim control.
[...]

Issue #1055, opened 2026-08-18
Dependency debt: Go toolchain 1.26.6 + ethers/ws test fixture (coordinator/gateway)


Augustas11:
## Dependency debt: Go toolchain + ethers/ws test fixture

Surfaced by a `govulncheck` / `npm audit` pass during the SPEC-042 audit (2026-08-18). These are pre-existing repo dependency findings, unrelated to any SPEC text. Filing so an agent can pick them up.

### 1. Go standard-library vulnerabilities (coordinator + gateway) — HIGH

`govulncheck ./...` on both `phase4-coordinator` and `phase5-gateway` reports affected Go stdlib vulnerabilities fixed in **Go 1.26.6**, across `net/http`, `crypto/tls`, `encoding/xml`, `encoding/asn1`, and `net/url`.

- Files: `phase4-coordinator/go.mod`, `phase5-gateway/go.mod` (currently on Go 1.26.x)
- Fix: bump the `go` directive / build toolchain to **1.26.6** (or newer) in both modules and rebuild.
- Verify: `govulncheck ./...` clean in each module; CI builds green; money-path smoke (billing/settlement) unchanged.

### 2. `ethers -> ws` advisory in payout EIP-712 test fixture — LOW

`npm audit --audit-level=high` flags `ethers@6.13.4 -> ws` (`GHSA-96hv-2xvq-fx4p`, WebSocket DoS).

- File: `phase4-coordinator/internal/payout/testdata/eip712_js/package.json`
- Fix: bump `ethers` to **6.17.0** (pulls a patched `ws`), refresh the lockfile.
- Scope: test-only fixture; not served at runtime, but worth patching so it can't be copied into runtime tooling.
- Verify: the EIP-712 fixture still produces identical signatures/vectors after the bump.

### Notes

- Both are money-path-adjacent modules; follow the repo's money-path dep-upgrade smoke discipline (verify billing/settlement/payout semantics unchanged before/after).
- Neither finding is caused by SPEC-042; keep this separate from the SPEC-042 pool-control-plane work (issue #1053).

PR #1076, opened 2026-08-19
SPEC-042 gateway: pool authorization + capability handshake + X-MacProvider-Pool emit

Augustas11:
## SPEC-042 gateway: credential pool authorization + capability handshake + `X-MacProvider-Pool` emit

**Stacked on #1072** (`feat/spec-042-impl-stage-e`). Review the coordinator stack first (#1065 → #1069 → #1072); this is the gateway half that makes pool selection work end-to-end.

The coordinator already ingests an authorized `X-MacProvider-Pool` header, routes pool traffic to members only, and binds `pool_id` into the settlement spine — but **nothing emitted that header**. This slice builds the emitting half (SPEC-042 **R002** selection authorization + **R010** taxonomy and positive capability negotiation), coordinator-only + gateway-only, **default-off and byte-identical when off**.

### What it does
1. **Credential authorization (R002).** A buyer selects a pool via the control-plane header `X-MacProvider-Pool-Select` (never a body field — no pool metadata leaks into the provider-bound body). The pool must be in the credential's authorized set (`features.trusted_pools.account_pools`, account-scoped ceiling, narrow-only), checked **before** quota reservation. Emits `X-MacProvider-Pool: <pool_id>` only for an authorized, capability-satisfied selection.
2. **Non-disclosing rejection + no timing oracle (R010).** Every denial cause returns the same generic `pool_unavailable` (503, non-retryable); authorization is a pure local lookup that runs **before** any coordinator roundtrip, so an unauthorized caller's latency can't reveal pool existence. `pool_selection_invalid` (400) fires only on a conflicting selection (the header supplied twice with different values).
3. **Positive capability handshake (R010), both halves.** The coordinator advertises `pools.enabled` in `/internal/routing` (gated `trustPools != nil`); the gateway refuses to emit unless it reads a **fresh** `true` — so an old/rolled-back coordinator (which ignores `pool_id` and would route globally) fails the request closed instead of spilling pool traffic to global.
4. **Auth-mode gating (R002 deferral).** Only plain API-key credentials may select a pool. Wallet sessions may not (needs the SPEC-040 pool-aware envelope amendment — the selector isn't in the signed profile); demo traffic may not. Either presenting a selector → `pool_unavailable` before dispatch.
5. **Fail-closed everywhere.** Feature-off + pool named → rejected (never silent pool→global); config rejects header-unsafe / non-base64url pool ids (coordinator-sanitizer spill guard); id-less dedupe fingerprint keys on the resolved `poolID` so a pool request can't replay a global attempt (or vice-versa).

### Money-path audit — codex 3 lanes → 0 C/H/M
This landed clean only after **three audit rounds** that caught real defects:
- **R1:** wallet/demo could select pools despite the deferral (×3 lanes); stale 5s capability cache → pool→global spill during rollback (security). → fixed: row-0 auth-mode gate; fresh capability fetch.
- **R2:** moving the selector out of the body broke the id-less dedupe fingerprint (body-keyed) → identical-body pool vs global requests could replay across the boundary (code). → fixed: `poolID` threaded into the fingerprint (v2→v3) + regression test.
- **R3:** code **APPROVE** 0/0/0, security **PASS** 0/0/0 (verified no remaining replay/capability/smuggling/settlement bypass). Architect **PASS** 0/0/0 (R2). Only residual was a gofmt LOW, applied.

### Compatibility
- Default-off; poolless + feature-off requests are byte-identical (no new header, no capability fetch). Old coordinator (no `pools` advertisement) → gateway fails closed; global traffic untouched.
- Coordinator change is one advertisement block; gateway change is behind `features.trusted_pools`.

### Verification
Green: gateway `router`/`config`; coordinator `buyer`/`billing`/`requestlog`/`routing`/`trustpool`. New tests: coordinator advertisement toggle; config authorize/validate/base64url; gateway handler conformance (emit, fail-closed non-dispatch, capability-not-consulted-for-unauthorized, fresh-per-request capability, byte-identical global, conflict, wallet + demo rejection, id-less dedupe pool/global separation).

> **`spec-index / check` is advisory-red** here, same as #1069/#1072: the per-requirement CONFORMANCE records for SPEC-042 R001–R012 are deferred (skeleton R010 promotion checklist), so `SPEC-042-R002`/`R010` are honest but not yet registered. The check is advisory (not `ci-required`) and does not block merge.

### Deferred (documented, each fails closed)
Wallet-session pool selection (SPEC-040 envelope amendment); per-API-key pool columns (account-granularity is the MVP binding); provider-half capability advertisement; predicate-specific error codes. **Carried, not a slice defect:** go 1.26→1.26.6 stdlib vulns (`go.mod` untouched by this slice; tracked in #1055).

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-042"],
  "requirements": ["SPEC-042-R002", "SPEC-042-R010"],
  "authority_domains": ["pool-control-plane"],
  "arbitration": ["CODE_BUG"],
  "tests": ["phase5-gateway/internal/router/pool_selection_test.go"],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/1053"
}

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,5 @@ test/e2e/thermal-soak/*.ndjson
 
 # Local #615 audit artifacts (reports/diffs; keep out of PRs)
 scratchpad/615-audit/
+audits/2026-08-19/gateway-pool-auth-fulldiff.patch
+scratchpad-*.log
--- a/audits/2026-08-19/AUDIT_GATEWAY_POOL_AUTH_ARCHITECT_PROMPT.md
+++ b/audits/2026-08-19/AUDIT_GATEWAY_POOL_AUTH_ARCHITECT_PROMPT.md
@@ -1,0 +1,44 @@
+# Codex audit — SPEC-042 gateway pool auth/emit slice — ARCHITECT lane
+
+You are the architecture-review lane. Judge whether this slice is the right
+shape, sits at the right boundary, and composes correctly with the coordinator
+pool routing already merged in the base. Review the FULL slice diff as it lands.
+
+Diff: `audits/2026-08-19/gateway-pool-auth-fulldiff.patch` (base `c70f97ce`,
+which already contains coordinator pool ingestion/routing/isolation/settlement
+labels). Read changed files in the worktree. Spec:
+`specs/SPEC-042-pool-control-plane.md` (R002, R010). Design:
+`docs/design/spec-042-v0.1-slice-gateway-pool-auth.md`.
+
+## Context
+This is the gateway half of SPEC-042 Layer 2: credential->pool authorization +
+the positive coordinator↔gateway capability handshake (R010) + the
+`X-MacProvider-Pool` emit. Deliberately deferred (documented in the design):
+wallet-session pool selection (needs a SPEC-040 envelope amendment); per-API-key
+pool columns (this slice binds at account granularity via operator config); the
+provider-half capability advertisement; predicate-specific error codes.
+
+## Focus (Critical/High/Medium/Low/Info, with rationale)
+1. Boundary correctness: is authorization at the right layer (gateway, before
+   reservation/dispatch)? Is the capability handshake modeled correctly —
+   coordinator advertises, gateway refuses on absence — and does the division of
+   labor (gateway checks coordinator; coordinator enforces member-only routing;
+   provider-half deferred) leave any spill gap that is NOT fail-closed?
+2. Contract/compat: the coordinator advertisement is a shared wire contract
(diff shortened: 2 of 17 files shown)


antfleet-ops (approved):
SPEC-042 gateway pool auth/emit + capability handshake, rebased directly onto main. Both modules build + gateway router/config + coordinator advertisement test green on main. Codex 3-lane money-path audit 0 C/H/M (3 rounds).

Issue #1053, opened 2026-08-18
SPEC-042 — Pool Control Plane and Trusted-Pool Manifest


Augustas11:
## SPEC-042 — Pool Control Plane and Trusted-Pool Manifest

Tracking issue for `specs/SPEC-042-pool-control-plane.md` (new authority domain `pool-control-plane`).

**What:** Define the Layer 2 pool control plane — a pool ("subnet" / "Trusted Pool") as a creator-operated trust boundary over shared MacProvider infrastructure and settlement, addressed by a mandatory `pool_id` authority field and governed by a signed pool manifest.

**Why now:** `pool_id` appears zero times in `phase4-coordinator/internal` and `phase5-gateway/internal`; routing begins from the global provider snapshot. There is no pool object today, so Layer 2 traffic is deploy-blocked, not merely incomplete. Live-prod (2026-08-18, `coordinator.malibu.tech` v1.8.95) runs `require_attestation`/`require_encrypted_leg` fail-open, so a Trusted Pool cannot inherit its trust floor from the network and must carry fail-closed pool-level predicates.

**Scope boundary:** This is **administrative trust (Layer 2)**, NOT cryptographic privacy. It MUST NOT be marketed as "Privacy Pools" — that term is reserved for the Layer 3 capability set (SPEC-041 relay-blind request encryption and later unlinkability/anonymous-membership work). SPEC-042 forward-declares and binds Layer 3 compatibility fields so Layer 3 can be added without re-architecting.

**Requirements (v0.0.2):** R001 signed manifest + `pool_id`=digest identity/rollback · R002 `pool_id` threading + selection authorization · R003 durable per-pool membership/revocation · R004 fail-closed predicates + exact SPEC-008 attestation mapping · R005 per-attempt tenant isolation + generation fencing · R006 settlement bound to route snapshot · R007 field-level retention matrix · R008 buyer-verifiable `pool_policy.json`/`pool_status.json` · R009 Layer 3 binding + claim control · R010 config/rollout/error taxonomy/evidence gate · R011 lifecycle/migration/failover.

**Status:** draft-skeleton. Round-1 three-lane codex audit applied. Not build-ready until the five-lane audit reports 0 C/H/M, AUTHORITY/CONFORMANCE registration lands, and the open decisions (Open-A poolless null-pool, Open-B SPEC-015 receipt-field vs. route-context, canonical encodings/schemas) close.

[...]

PR #1079, opened 2026-08-19
SPEC-042 R004/R010: pool provider binary-version floor (handshake provider half)

Augustas11:
## SPEC-042 R004/R010: pool provider binary-version floor (handshake provider half)

Completes the SPEC-042 R010 positive capability handshake **end-to-end**. The gateway already verifies the coordinator advertises pool support (#1076); this adds the **provider half** inside the coordinator: a pool request is never routed to a member whose `binary_version` is below the pool's configured minimum (a pool-blind provider under a mixed-binary rollout). Coordinator-only; reuses the provider's existing `binary_version` handshake, so **no provider/Swift change and no release** (Option A).

### What it does
- **Per-pool floor in the consistent snapshot.** `trustpool` `poolState.minBinaryVersion` → `Snapshot.MinBinaryVersion`, read under the same lock as members/generation. `SetMinBinaryVersion` **bumps the generation**, so raising a floor invalidates in-flight reservations via the existing R005 fence (validated with `versionfloor.Valid`, rejects a malformed floor rather than bricking the pool).
- **Gate on every dispatch-authorizing path.** New `ProviderMeetsPoolBinaryFloor` filter gate (after membership, which dominates) + by-hand re-checks on the pinned-session, pinned/self-route, slot-queue-enter, and slot-queue-poll paths — same discipline as the isolation gate, so a pin can't bypass the floor.
- **Fail-safe + fail-closed.** Empty/malformed `binary_version` under a floor → excluded (`versionfloor.Compare` not-ok); all-under-version → `pool_binary_too_old` (503, non-retryable), no spill. Floor `""` is a strict no-op; global/poolless byte-identical.
- **Error taxonomy.** `pool_binary_too_old` registered in `spec018RetryableByCode` + `coordinatorEmittedErrorCodes`; envelope maps it before `pool_no_eligible_member` (present-but-under-version is a more specific signal than no-member); reason surfaced distinctly in the routing-decision log.

### Money-path audit — codex 3 lanes
- **Code:** round 1 found 1 Medium (floor validation) + 1 Low (comparator tests) → **both fixed**, re-verified.
- **Security + Architect:** confirmed the enforcement is solid — every dispatch path gated, malformed fails closed, floorless/poolless inert — and both raised the same **HIGH: the floor compares against self-reported `binary_version`, not SPEC-020 signed release evidence** (R004's literal "signed" requirement). This is an **accepted, documented carry** (owner-decided):
  - It reuses the *exact* signal the pre-existing #768 model-version floor and global-admission floor already trust — **not a new/worsened trust assumption**.
  - The primary boundary, pool **membership**, is an authenticated SPEC-003 identity explicitly admitted by the pool creator; the spoof threat is bounded to a creator-admitted-then-compromised member.
  - There is **no** coordinator-held verified binary evidence today (attestation binds the SE key/tier, not the version; no SPEC-020 evidence store) — a real fix is a separate, sizeable build.
  - The feature is **default-off / unenabled** — no live exposure.
  - **Upgrade path is localized:** when verified binary evidence exists, swap only what `poolBinaryFloorMet` compares against; the rails are unchanged.
  - Documented in `docs/design/spec-042-v0.1-slice-pool-binary-floor.md` §7 and a code trust-note.
- **Architect** also carried a second HIGH: **floorless pools are inert** — R004's binary predicate is *optional*; per-pool policy completeness (require a floor / advertise per-pool readiness) belongs to the enable path + manifest (R001/R008/R011). Documented in §7.

### Verification
`buyer` / `routing` / `trustpool` green (incl. the error-code AST guard and the existing pool-isolation tests). New conformance tests: all four dispatch paths exclude under-version members; all-below-floor fails closed non-retryable; at/above served; empty/malformed version fail-safe excluded; no-floor inert; global unaffected; membership dominates; `SetMinBinaryVersion` bumps generation + rejects malformed floors; comparator table tests.

### Compatibility / scope
- Default-off; global/poolless/floorless paths byte-identical. Provider handshake unchanged. Coordinator-only (phase4).
- Deferred (documented, each fails closed or inert by design): SPEC-020 signed release evidence as the trust source; the signed manifest (R001) as the floor source; a dedicated `pool_support_version` field; other R004 predicates (encrypted-leg, attestation).

> **`spec-index / check` is advisory-red** (SPEC-042 per-requirement CONFORMANCE records R001–R012 are still deferred), same as the rest of the SPEC-042 stack. Advisory, not merge-blocking.

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-042"],
  "requirements": ["SPEC-042-R004", "SPEC-042-R010"],
  "authority_domains": ["pool-control-plane"],
  "arbitration": ["CODE_BUG"],
  "tests": ["phase4-coordinator/internal/buyer/spec042_pool_binary_floor_test.go"],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/1053"
}

### DIFF
--- a/audits/2026-08-19/AUDIT_POOL_BINARY_FLOOR_ARCHITECT_PROMPT.md
+++ b/audits/2026-08-19/AUDIT_POOL_BINARY_FLOOR_ARCHITECT_PROMPT.md
@@ -1,0 +1,40 @@
+# Codex audit — SPEC-042 pool binary-version floor — ARCHITECT lane
+
+Judge whether this slice is the right shape and composes correctly with the
+merged SPEC-042 Layer 2 stack. Review the FULL slice diff as it will land.
+
+Diff: `audits/2026-08-19/pool-binary-floor-fulldiff.patch`. Read changed files.
+Spec: `specs/SPEC-042-pool-control-plane.md` (R004, R010, R005). Design:
+`docs/design/spec-042-v0.1-slice-pool-binary-floor.md`.
+
+## Context
+This completes the R010 positive capability handshake end-to-end. The gateway
+already verifies the coordinator advertises pool support (#1076); this adds the
+provider half inside the coordinator (the gateway can't see the selected
+provider). It reuses the provider's existing `binary_version` handshake rather
+than adding a dedicated `pool_support_version` field (Option A). Deferred: a
+dedicated capability field decoupled from binary version; the signed manifest
+(R001) as the floor source; other R004 predicates (encrypted-leg, attestation).
+
+## Focus (Critical/High/Medium/Low/Info + rationale)
+1. Boundary: is enforcing the floor inside the coordinator's eligibility gate the
+   right layer? Is keying "pool support" off `binary_version` a sound proxy for
+   the R010 "provider advertises pool support at the required version", or does
+   it conflate concerns in a way that will hurt later?
+2. Floor home: per-pool `minBinaryVersion` in the trustpool snapshot (rather than
+   a global option). Correct call for consistency with the generation fence and
+   R004's per-pool model? Any drift risk?
+3. Composition: does the new gate slot cleanly alongside the existing
+   membership/model-floor/receipt-key gates without double-counting or reordering
+   hazards? Is the envelope precedence (pool_binary_too_old vs
+   pool_no_eligible_member vs model_version_floor) sensible?
+4. Scope discipline: are the deferrals sound and fail-closed (each defaults to
+   no-floor = inert, which is fail-OPEN for the binary dimension when a pool is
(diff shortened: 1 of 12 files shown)


antfleet-ops (approved):
SPEC-042 pool binary-version floor (R010 provider half). Enforcement verified solid by all lanes (every dispatch path gated, malformed fail-safe, floorless/poolless inert); code lane APPROVE 0 C/H/M after round-1 fixes (floor validation + comparator tests). Carries a documented, owner-accepted HIGH: the floor compares against self-reported binary_version (not-worsened vs the existing #768 model-version floor; localized SPEC-020 upgrade path) — see PR body + design doc §7. Default-off.

Issue #1268, opened 2026-08-28
Autotune signed feed expires at 30 days with no renewal — SIGHUP hot-reload + automated renewal


Augustas11:
Part of #1235 (a 5th strand class not in the epic's original diagnosis).

## Problem
The coordinator's signed autotune feed (rate-card/candidate/demand) has a hard **30-day client freshness horizon**: provider join preflight fails closed with `rate_card_update_required` when `now - feed.generated_at > 30 days` (AutotuneRecommend.swift `loadSignedStatic`; enforced at join in MacProviderCLI.swift `runModelCatalogPreflight`). Nothing renews the feed, so it silently ages out. The gate runs only at daemon **startup/join**, so the fleet does not fail all at once — it bleeds as nodes restart and cannot rejoin, and the rejection is **client-side and invisible** to the coordinator. This caused a ~2h fleet-wide degrade on 2026-08-28 when the 2026-07-29 feed crossed 30 days.

## Fix (two parts)
1. **Hot-reload the autotune catalog on SIGHUP.** Today the catalog + `/v1/*` feed are boot-only (`reloadCoordinatorConfig` does not re-parse `AutotuneFeeds`; SIGHUP reloads tier2/payout/billing only). Add a thread-safe, **fail-closed** swap of `s.autotuneCatalog` + `s.autotuneCompatibleCatalogs` (mirror the tier2 `ConfigureDefaultStrict` reload pattern: validate the new catalog, reject the reload and keep the old on any error). This makes a feed refresh a `kill -HUP` with **zero provider disruption** — no restart.
2. **Automated renewal.** A scheduled job that regenerates the feed with a fresh `generated_at` (same content, same `policy_version`, new version string), re-signs with the trusted static key, deploys to Pearl (new release dir, retarget `current`, keep prior as `.previous-target`), and `kill -HUP`s the coordinator. Cadence weekly (well inside 30 days). Key-custody decision required (CI secret vs Pearl-local vs operator machine).

## Monitor
Alert when the live `/v1/rate-card` `generated_at` age crosses ~20 days, so a failed renewal is caught **before** it strands the fleet.

## Notes
- The manual re-sign+deploy procedure is proven (used to recover on 2026-08-28); it currently requires a coordinator restart, which is why the SIGHUP reload is the prerequisite.
- Related renewal automation for the discovery head is epic child **D**; this is the analogous need for the autotune feed.

Augustas11:
Status check (2026-08-31): **keep open.**
[...]

PR #1273, opened 2026-08-28
Refresh SPEC-032-R003 selector after #1268 hello-gate split

Augustas11:
## Summary
Refresh a stale CONFORMANCE mapping selector after the #1268 SIGHUP hot-reload merge (7ec6ea69). No behavior change — this is a governance-manifest (`specs/CONFORMANCE.json`) selector correction only.

## What changed
`SPEC-032-R003.implementation[6]` pointed at `func (s *Server) checkAutotuneHelloGate`. #1268 split that method into a snapshot-taking wrapper plus a `checkAutotuneHelloGateWithCatalog` core (so one admission pins a single autotune-catalog generation). The bare-prefix selector now matches **both** functions, so `check_spec_governance` reports it as unresolved. Re-point the mapping at the core `func (s *Server) checkAutotuneHelloGateWithCatalog`, where the hello-gate authorization logic now lives. Unique, resolves cleanly.

## Known remaining spec-index item (not fixed here)
`SPEC-010-R005` (conformant) still reports a commit-evidence drift: #1268 converted a `s.autotuneCatalog` read inside `verifyProviderModelIdentity` to the thread-safe getter `s.currentAutotuneCatalog()` (required — that read is concurrent with the new SIGHUP `SetAutotuneCatalog` writer). That is a behavior-preserving change, but it moved the mapped fragment, and R005's commit anchor is cryptographically bound to a **signed provider-prebeta-admission journey-result** (commit c59568b). Re-anchoring requires a fresh signed journey capture against the new binary (live air5 provider + journey-result signing) — tracked as recapture debt, not fixable by editing the manifest. Left truthfully flagged rather than downgraded or re-stamped.

## Validation
`python3 scripts/check_spec_governance.py --base-ref origin/main` — the `checkAutotuneHelloGate` selector error is resolved (only the R005 recapture item remains).

{
  "behavior_change": "none",
  "contract_change": "yes",
  "specs": [],
  "requirements": [],
  "authority_domains": [],
  "arbitration": [],
  "tests": [],
  "journeys": [],
  "issue": "https://github.com/Augustas11/macprovider/issues/1268"
}

### DIFF
--- a/specs/CONFORMANCE.json
+++ b/specs/CONFORMANCE.json
@@ -2224,7 +2224,7 @@
         "phase4-coordinator/internal/pool/provider.go:func (r *Registry) QuarantineForProofOfWeightsReload",
         "phase4-coordinator/internal/ws/server.go:func (s *Server) SetProofOfWeightsConfig",
         "phase4-coordinator/internal/ws/server.go:func (s *Server) revalidateProofOfWeightsAdmissions",
-        "phase4-coordinator/internal/ws/server.go:func (s *Server) checkAutotuneHelloGate",
+        "phase4-coordinator/internal/ws/server.go:func (s *Server) checkAutotuneHelloGateWithCatalog",
         "phase4-coordinator/cmd/coordinator/main.go:func reloadCoordinatorConfig",
         "phase4-coordinator/cmd/coordinator/main.go:func telemetryDriftEvaluatorForReload"
       ],


antfleet-ops (approved):
Governance-only selector refresh after #1268 split checkAutotuneHelloGate. Verified the re-pointed selector resolves uniquely to checkAutotuneHelloGateWithCatalog where the gate logic now lives; no behavior/contract-requirement change. R005 recapture debt is correctly left flagged, not downgraded.

PR #1276, opened 2026-08-28
Autotune static-feed renewal automation (SIGHUP hot-reload, fail-closed deploy)

Augustas11:
## Summary
Automates the **autotune static-feed freshness renewal** that #1268 made restart-free. The signed feed carries a client-side 30-day `generated_at` horizon; when it lapses, any provider that restarts is stranded before it connects. This ships a fail-closed, self-verifying renewal that re-dates + re-signs the feed and hot-reloads the coordinator over SIGHUP — **zero provider disruption**.

## What changed
- **`scripts/renew-autotune-static-feed.sh`** (new): builds a re-dated, re-signed release in an ephemeral worktree off `origin/main` (never dirties the caller's tree), assembles the 9-file release dir, and gates it with `catalog-release.py verify-directory`. Default is **dry-run**; `--deploy` publishes to the coordinator host.
- **`scripts/resign-autotune-static.sh`**: the signer now passes the key **file path** to the Swift signer instead of the key bytes in a child-process environment.
- **`docs/runbooks/autotune-feed-renewal.md`** (new): the procedure, the security model (signing stays off the production host), the weekly launchd schedule (operator-local), and rollback.

## Security & fail-closed properties
- **Key never leaves the signing host**: signs locally, pushes only signed bytes; the reused signer derives the public key in-memory vs the trusted keyring (no bytes printed) and reads the key from a file, not env.
- **`verify-directory` gates before any deploy**; content-continuity guard **aborts on any non-date change** (a real catalog change must go through a reviewed release, never this cron).
- **Deploy is atomic + serialized**: atomic remote lock (no concurrent publishes), clock-skew guard (no future-dated feed), atomic symlink swap keeping the prior release as `.previous-target`.
- **Exact-identity activation check**: verifies the served `/v1/rate-card` `generated_at` **exactly** equals this run's stamp (not a freshness window) — and **rolls back both `current` and `.previous-target` + re-HUPs** on any regression.

## Testing
- Local dry-run passes end-to-end (ephemeral worktree → re-date → generate → sign → assemble → `verify-directory`).
- Content-continuity confirmed against the live feed (dates-only delta).
- `bash -n` + `shellcheck -S warning` clean on both scripts; release-ref validation unit-tested (accepts valid refs, rejects injection/slashes/command-substitution/empty).

## Audit
Codex 3-lane audit (security / correctness / architecture) over the full diff, iterated to **0 CRITICAL / 0 HIGH / 0 MEDIUM**:
- Round 1: 4 HIGH (`.previous-target` double-prefix; no rollback after swap failure; success-on-stale-feed; signer key in child env) + 4 MEDIUM — all fixed.
- Round 2: 7/8 confirmed; 1 residual MEDIUM (remote-read target values passed to SSH argv unvalidated) — fixed with shape validation matching the coordinator parser.

Evidence under `audits/2026-08-29-feed-renewal/` (kept local, not in this PR).

### DIFF
--- a/docs/runbooks/autotune-feed-renewal.md
+++ b/docs/runbooks/autotune-feed-renewal.md
@@ -1,0 +1,122 @@
+# Autotune Static Feed Renewal (freshness re-stamp)
+
+The coordinator's signed SPEC-023 autotune feed carries a **30-day freshness
+horizon** enforced client-side: `AutotuneRecommend.swift` `loadSignedStatic`
+fails closed when `now - generated_at > 30*24*3600`, which sets
+`rateCardUpdateRequired` / `candidateCatalogUpdateRequired` and aborts the
+provider daemon **before it connects** (`runModelCatalogPreflight`). The gate
+runs only at daemon start / join, so a stale feed silently arms and providers
+drop one-by-one as they restart or the coordinator cycles them. This is an
+ops-renewal gap, not a design flaw — the guard on month-old signed pricing data
+is correct.
+
+**The fix is to re-date + re-sign the feed on a schedule.** Since #1268 the
+coordinator hot-reloads the feed on `SIGHUP` (`reloadCoordinatorConfig` swaps
+the WS admission catalog and the buyer-served `/v1/*` bytes atomically,
+fail-closed), so renewal is now **zero-disruption**: no coordinator restart,
+every provider WebSocket stays connected.
+
+## Security model — signing stays off the production host
+
+The Ed25519 feed-signing key (`streamvc-autotune-static-v4`) is what protects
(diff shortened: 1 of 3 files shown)


antfleet-ops (approved):
Reviewed the feed-renewal automation. Signing stays off the production host (key read from file, not env; pubkey derived in-memory vs trusted keyring). verify-directory gates before deploy; content-continuity guard is freshness-only; deploy is atomic + locked + clock-skew-guarded with exact-identity activation check and full rollback (current + .previous-target). Codex 3-lane audit iterated to 0 C/H/M; remote-read target values now shape-validated against the coordinator parser.

Issue #1235, opened 2026-08-27
Epic: Forward-resilient provider updates — converge the fleet + close the strand paths


Augustas11:
## Purpose

Converge the provider fleet on the current release and keep it converged without
manual per-node rescue. This is the **prevention + convergence** epic; recovery of
individual stranded nodes stays in #1184 / #1188 / #1189 / #1200 / #1203.

> **Status: diagnosis corrected and evidence-backed (see comments).** The original
> "forward-compat trapdoor" theory was refuted against code by four independent
> reviews and a live-evidence pass. The accepted diagnosis and plan are below.

## Privacy Rule

No provider names, hostnames, full ids, addresses, or payout details here.
Anonymous labels only; identifying detail stays in private operator notes.

## Accepted diagnosis — four distinct block classes, no single trapdoor

Live evidence (4 nodes) shows fleet fragmentation is not one bug but a *class* of
release/coordinator/environment failures, each stranding a different cohort:

1. **Pre-rebrand parser strand.** Clients ≤1.8.94 hard-reject every release
   ≥1.8.95 because Malibu identifier tolerance (`live.malibu.provider.cli`) landed
   in v1.8.95 (commit 9c78c988). The rejection surfaces as
   `compatibility_set_invalid` (a redaction of the real `cli_identifier` sublabel).
   No bridge release exists — this cohort cannot self-recover.
2. **Coordinator compat-set target missing.** The coordinator advertises a
   `recommended_binary_version` with no `recommended_compatibility_set_id`, so
   clients fail eligibility with `coordinator_compatibility_target_missing`
   (`AutoUpdater.swift:191`; coordinator emits the target only when
   `compatibility_set.target_id` is configured, `server.go:2975/2989`).
3. **Expired signed discovery head (fleet-wide).** The coordinator-independent
   self-heal rail is down everywhere (`discovery_head_expired`, hundreds of
   attempts/node). Head validity is 7 days; renewal is a manual approval ceremony.
4. **ACL write barrier (external node, incident #1189).** The watchdog's
   autoupdate-recovery write is ACL-rejected on the provider home dir for days;
   the node stays "zombie serving" while never updating.

Supporting facts: the published v1.8.105 release artifact is **valid** (verified);
coordinator `provider_hardware_profiles` is **stale** (refreshes only on
`cli_hello`), so version dashboards built on it mislead; three systems disagree on
[...]

PR #1277, opened 2026-08-28
Automate discovery-head renewal cadence + freshness alarm (#1235 child D)

Augustas11:
## Summary
Closes the chronic-lapse of the signed release-discovery **self-heal rail** (Epic #1235 block-class #3). The head has a 24h validity but was renewed only by manual dispatch, so it repeatedly lapsed — leaving the coordinator-independent recovery path down fleet-wide for days at a time (e.g. a ~5-day gap 08-19→08-25).

## What changed
- **`renew-release-discovery-head.yml`**: adds a **weekly `schedule:`** and bumps the default validity **24h → 168h** (the workflow/builder/verifier/client all already cap at 7 days), so the head is auto-dispatched well inside its window. Security posture is unchanged — the renewing job still runs in `environment: production-release` (antfleet-ops approval), seals OpenSSL, and scopes `contents: write` to that job only.
- **`discovery-head-freshness-alarm.yml`** (new): a **read-only, secret-free** scheduled check (every 6h) that fails/notifies when the client-visible head is expired or within 48h of expiry — backstopping a skipped/late approval **before** the rail lapses. It mirrors the client's exact selection (the 20 most-recent releases, strict `release-discovery-v1-[1-9][0-9]*` grammar, highest sequence), so it alarms on what clients actually see, not merely what exists.
- **`scripts/check-discovery-head-freshness.py`** (new): the freshness check (stdin head JSON → exit 1 if expired/imminent/malformed/empty), unit-tested.

## What this does NOT change
The **antfleet-ops approval gate stays intact.** This turns "remember to dispatch or the rail lapses" into "auto-dispatched weekly, approve when pinged, alarm catches a miss." Making the *scheduled* renewal fully unattended (it only re-signs an already-approved target) is a separate key-custody decision, deliberately not taken here.

## Testing / audit
- `actionlint`, both workflows valid YAML, `test-renew-release-discovery-head.sh` passes; freshness script unit-tested (valid/expired/imminent/malformed/empty); selection logic verified (highest sequence, rejects leading-zero/malformed).
- Codex 3-lane audit over the full diff → **0 CRITICAL / 0 HIGH**. The one MEDIUM (alarm scanned 200 releases while the client sees only 20 → could pass on a head clients can't see) and one LOW (looser tag grammar) are fixed by mirroring the client's 20-release window + strict grammar.

Evidence under `audits/2026-08-29-dh-automation/` (kept local, not in this PR).

### DIFF
--- a/.github/workflows/discovery-head-freshness-alarm.yml
+++ b/.github/workflows/discovery-head-freshness-alarm.yml
@@ -1,0 +1,72 @@
+name: Discovery head freshness alarm
+
+# Read-only backstop for the signed release-discovery self-heal rail. The renewal
+# workflow (renew-release-discovery-head.yml) auto-dispatches weekly at 168h
+# validity but still waits on the production-release approval; if that approval is
+# skipped/late, this alarm fails (and notifies) BEFORE the head lapses fleet-wide.
+# No secrets, no production-release environment, no signing — it only reads the
+# latest head's self-asserted expiry.
+
+on:
+  schedule:
+    # Every 6 hours: far more often than the weekly renewal, so a stale head is
+    # caught with well over a day of runway (alarm threshold is 48h).
+    - cron: "0 */6 * * *"
+  workflow_dispatch:
+    inputs:
+      min_hours:
+        description: "Alarm if the head expires within this many hours (default 48)"
+        required: false
+        default: "48"
+        type: string
+
+permissions:
+  contents: read
+
+concurrency:
+  group: discovery-head-freshness-alarm
+  cancel-in-progress: true
+
+jobs:
+  check:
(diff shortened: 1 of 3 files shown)


antfleet-ops (approved):
Reviewed the discovery-head renewal automation. The weekly schedule + 168h validity keeps the renewing job in production-release (approval gate unchanged); the read-only alarm is secret-free and now mirrors the client's exact 20-release window + strict grammar, so it alarms on client-visible staleness. Codex 3-lane audit 0 C/0 H; the MEDIUM (200 vs 20 window) and LOW (grammar) are fixed. Freshness script unit-tested.

Issue #1285, opened 2026-08-29
P0: Fresh-Mac provider install hangs — installer hard-depends on python3 (Command Line Tools stub) 


Augustas11:
## Summary (P0 — blocks new-provider onboarding)

The Malibu app's provider installer (`phase3-binary/dist/install.sh`) **hard-depends on `python3`** (71 invocations), but a **stock consumer macOS does not ship a usable `python3`** — `/usr/bin/python3` is an Xcode Command Line Tools *stub*. So a brand-new, non-developer provider Mac cannot run the installer, and the failure presents as a **silent hang at "Starting installer…"**.

## Root cause (verified in code)

- `require_tool` checks presence with `command -v python3` (`install.sh:require_tool`). On a stock Mac the CLT **stub** at `/usr/bin/python3` exists, so this check **passes** even though real Python is absent.
- `main()` then calls `python3` very early — the first real invocation is `validate_install_dir` (before any download, i.e. during the "Starting installer" phase).
- On a Mac without Command Line Tools, invoking the `python3` stub triggers macOS's **"Install Command Line Developer Tools"** system dialog. That dialog appears **behind the Malibu window**, so the app looks stuck at "Starting installer…". The install cannot proceed until CLT is installed.
- The app's progress monitor (`CLIInstallRunner.ActivityMonitor`) only recognizes `macprovider-cli` processes, so the entire pre-CLI phase already shows "Starting installer…" — masking that the installer is blocked on a system prompt.

## Why it's deterministic / fleet-wide

This affects **every fresh non-developer Mac** (the majority of real providers) — which is why multiple independent providers hit the identical hang on first install. It is NOT a network issue (GitHub API + downloads verified fast) and is NOT fixed by the curl-timeout change (#1281).

## Evidence to confirm (fast)

Ask an affected provider whether an **"install command line developer tools"** dialog is hidden behind the Malibu window. A yes confirms this root cause.

## Interim workaround (non-technical)

Provider brings the hidden CLT dialog to front (Cmd+Tab / minimize Malibu), clicks **Install**, waits for CLT to finish, then reopens Malibu. This should get past "Starting installer".

## Fix directions (in order of correctness)

1. **Remove the `python3` dependency from `install.sh`** — reimplement the 71 python3 sites in POSIX shell or a small bundled/compiled helper. Most correct; consumer installs must never require developer tools. (Largest effort — needs an audit of what each python3 block does; several parse JSON / inspect ACLs / launchd.)
[...]

PR #1286, opened 2026-08-29
Guard fresh-Mac installs against the python3 CLT stub (fixes the onboarding hang) — #1285

Augustas11:
## Problem (P0 — the real new-provider onboarding blocker; issue #1285)

A brand-new, non-developer Mac hangs at **"Starting installer…"** on first
provider install. Root cause: `install.sh` hard-depends on `python3` (71 calls),
but a stock Mac's only `python3` is the **Xcode Command Line Tools stub** at
`/usr/bin/python3`. `require_tool` passed it via `command -v`, and the first real
`python3` call (`validate_install_dir`) triggered the OS **"Install Command Line
Developer Tools"** dialog — which appears **behind the Malibu window** and blocks
the installer forever. Deterministic for every fresh non-developer Mac → explains
multiple independent providers with the identical hang. (Not a network issue;
not fixed by #1281.)

## Fix (short-term guardrail from #1285)

`ensure_python3_usable` runs right after the tool-presence loop, before the first
`python3` use. It detects the non-functional stub **without invoking it**
(`xcode-select -p` does not pop the install dialog): the stub is only
`/usr/bin/python3` **and** only when no developer toolchain is selected. In
consumer/GUI mode it opens the Command Line Tools installer for the user, then
exits with a dedicated **code 8**; Malibu maps 8 → `developerToolsRequired` with
an actionable message instead of a hidden hang. A real `python3` (Homebrew,
python.org, or an active CLT/Xcode toolchain) is accepted unchanged.

Converts a hidden-dialog hang into clear, self-service recovery: "click Install,
wait, then reopen Malibu."

## Not in this PR (durable follow-up, #1285)

Removing the 71-site `python3` dependency from `install.sh` (or bundling a
self-contained Python in Malibu.app) so consumer onboarding never needs Apple
developer tools at all.

## Tested

- `bash -n install.sh`
- `phase3-binary/dist/test/install_python3_clt_guard.test.sh` (new): real-python3
  accepted / stub+no-CLT → exit 8 / stub+CLT accepted / guard precedes first
  python3 use — all pass; wired into `make test-dist`
- `--dry-run` end-to-end still exits 0 on a real-python3 Mac (no success-path change)

{
  "issue": "https://github.com/Augustas11/macprovider/issues/1285",
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-003"],
  "requirements": ["SPEC-003-R001"],
  "authority_domains": ["provider-onboarding-identity"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "bash -n phase3-binary/dist/install.sh",
    "bash phase3-binary/dist/test/install_python3_clt_guard.test.sh"
  ],
  "journeys": ["not-required"]
}

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,7 @@ test-dist:
 	bash scripts/test-tier2-enforcement-safety.sh
 	bash ops/pearl-updater/test_transaction_gate_systemd.sh
 	bash phase3-binary/dist/test/check_baked_static_feed_sync.test.sh
+	bash phase3-binary/dist/test/install_python3_clt_guard.test.sh
 	bash scripts/test-catalog-release.sh
 	bash scripts/test-autotune-gate-matrix.sh
 	bash -n phase4-coordinator/dist/deploy-pearl-vps.sh
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,7 @@ test-dist:
 	bash ops/pearl-updater/test_transaction_gate_systemd.sh
 	bash phase3-binary/dist/test/check_baked_static_feed_sync.test.sh
 	bash phase3-binary/dist/test/install_python3_clt_guard.test.sh
+	bash phase3-binary/dist/test/install_1286_fresh_mac_e2e.test.sh
 	bash scripts/test-catalog-release.sh
 	bash scripts/test-autotune-gate-matrix.sh
 	bash -n phase4-coordinator/dist/deploy-pearl-vps.sh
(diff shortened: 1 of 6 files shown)


antfleet-ops (dismissed):
Re-audit gate met: HIGH resolved (guard never invokes the CLT stub), MEDIUM fixed (python3 removed from generic require_tool loop → die 8 for both stub and no-python3), LOW fixed (exit-8 classifier test). Verified on real no-CLT hardware: fast die 8, no hang. Advisory spec-index `check` red is the known SPEC-010-R005 drift, non-blocking.

antfleet-ops (approved):
Round-4 audit ACCEPT (0 C/H/M; 1 LOW carried, residual demoted to INFO by the auditor). Bounded stdin-mode python3 probe closes the reproduced hang. EMPIRICALLY PROVEN on real no-CLT hardware (redastare, macOS 26.5): NEW install.sh exits 8 in ~1s with the actionable CLT message, no hang. Advisory spec-index `check` red is the known SPEC-010-R005 drift, non-blocking.

PR #1294, opened 2026-08-30
Fix Malibu app install hang (beachball) — ps pipe deadlock + off-main progress polling

Augustas11:
## Summary

Fixes the Malibu app hanging with a spinning-rainbow beachball at **"Starting installer"** (elapsed 0:00) the instant a provider enters the invite and presses Next — the reason new providers could not onboard through the app.

**Root cause (verified):** the install-progress poller ran `ps -ax -o command=` via `macproviderProcessLines()` and called `waitUntilExit()` **before** draining the output pipe. macOS pipe buffers are ~64KB; when `ps` output exceeds that, `ps` blocks writing into the full undrained pipe and never exits → `waitUntilExit()` hangs forever. It ran on the `@MainActor` progress loop, so the UI froze. Measured: **99KB** of `ps` output on a Mac running Electron apps (deadlock) vs **18KB** on a plain Mac (no hang) — explaining why it hit some Macs and not others, independent of macOS version, model cache, or the install itself.

## Fix

- New `capturedProcessOutput(...)` drains stdout to EOF and routes stderr to `nullDevice`, with a bounded watchdog — a >64KB writer can no longer deadlock. `macproviderProcessLines()` uses it.
- `ActivityMonitor.snapshot()` is now `async` and runs the subprocess **off the main actor**; `installProgress` is assigned back on `@MainActor` only.
- `bestEffortBootstrapLaunchdProvider()` / `restoreLaunchdIfCorrectiveRecoveryOwned()` made async/off-main; the `launchctl bootstrap` wait is **bounded** (10s + terminate, resumes once) so a wedged launchctl cannot hang shutdown, while still performing the restore on shutdown/cancel (not skipped on Task cancellation).
- Swept the onboarding path; the main install runner already drains via `readabilityHandler` and is unchanged.

## Testing

- App build + **478 MalibuTests pass** (incl. `testCapturedProcessOutputDrainsLargeOutputWithoutDeadlock`, emitting >128KB).
- 3-lane audit (code / concurrency / architecture) — 0 CRITICAL/HIGH/MEDIUM on the changed surface.

## Follow-up (pre-existing, not in this diff)

`CLIInstallTeardown.runBundledUninstall` and `PayoutWalletFlow.runCLI` use the same read-after-wait subprocess-capture shape and should migrate to a shared bounded helper. Carried as a LOW: `capturedProcessOutput` is SIGTERM-only (fine for the cooperative `ps` caller).

### DIFF
--- a/phase3-binary/app/Sources/Malibu/Agent/MalibuAgent.swift
+++ b/phase3-binary/app/Sources/Malibu/Agent/MalibuAgent.swift
@@ -304,7 +304,7 @@ final class MalibuAgent: ObservableObject {
                     try await AutotuneRecommendationRunner.runHardwareAdmissionRecovery(cliURL: cliURL)
                 }
                 guard !Task.isCancelled, !self.isShuttingDown else {
-                    self.restoreLaunchdIfCorrectiveRecoveryOwned(attemptedCorrectiveRecovery)
+                    await self.restoreLaunchdIfCorrectiveRecoveryOwned(attemptedCorrectiveRecovery)
                     return
                 }
                 self.hardwareRecoveryOwnsLaunchdRestore = false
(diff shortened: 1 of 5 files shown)


antfleet-ops (approved):
App-only concurrency fix for the install-progress beachball (ps pipe deadlock + off-main polling). 3-lane audit clean on the changed surface (0 C/H/M); 478 MalibuTests pass incl. a >128KB no-deadlock test. The red 'check' is the pre-existing CONFORMANCE evidence drift (requirements[9]/[19] on server.go/install.sh) already failing on origin/main and untouched by this PR. Approving.

PR #1296, opened 2026-08-31
Fix fresh-install "no paid model cleared" (escaped artifact path) — 1.8.117

Augustas11:
## Fresh-install "no paid model cleared" — escaped artifact path (1.8.117)

Every fresh install via `install.sh` (Malibu app **and** curl) failed at Mac
check with `die 30` "no paid model recommendation cleared … nothing committed",
on **any** Mac regardless of RAM. This was misdiagnosed in the field as a
memory/swap/16GB limit; it is a config **serialization/read-back mismatch**.

### Root cause
`autotune --recommend --apply` correctly clears and applies the paid model every
time. But the model artifact path lives under `…/Application **Support**/…`,
which contains a space, so `ConfigApplier.yamlScalar` quotes it via
`JSONEncoder().encode` — which on this platform escapes forward slashes, writing
`model_artifact_path: "\/Users\/admin\/…"`. `install.sh`'s read-back strips the
quotes but not the escaping, so the value starts with `\`; its absolute-path
check `case "$p" in /*)` fails, the installer concludes no paid model cleared,
retries 3×, and dies 30 with nothing committed. Manual `autotune --recommend
--json` "works" only because it emits JSON (unescaped by a JSON parser) and never
goes through `install.sh`'s read-back.

### Fix (defense in depth)
- `ConfigApplier.yamlScalar` **and** `ServeCommand.yamlPathLiteral` (durable-store
  migration writer): encode with `.withoutEscapingSlashes` so quoted paths keep
  literal `/` (root fix).
- `install.sh` `read_config_{artifact_path,model,catalog_model_id}`: unescape
  `\/`→`/` after stripping quotes, so an already-escaped config (older CLI) still
  reads back a real absolute path. This alone unblocks onboarding with the
  existing CLI.

### Validation
End-to-end on a 16GB M3 MacBook Air: with the fixed `install.sh` and the
**unchanged** 1.8.116 CLI, the install clears the paid 3B model, registers, and
goes live (coordinator pool check `tier=provisional, state=ready`; `/v1/models`
serving). 3-lane audit (code/security/architecture): all PASS, 0 C/H/M.

### Deployment
Ships as **1.8.117** (binaryVersion + project.yml + release-builds.tsv). The app
pins its own version's pkg, so it is re-cut; the curl link picks up 1.8.117 as
latest.

{
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-023"],
  "authority_domains": ["installer-autotune-policy"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "cd phase3-binary && swift test --filter ConfigApplierTests",
    "cd phase3-binary && swift test --filter ServeCommandTests/testPersistMigratedArtifactPathWritesSpacedPathWithoutEscapedSlashes",
    "bash phase3-binary/dist/test/install_config_path_escape.test.sh"
  ],
  "journeys": ["not-required"]
}


antfleet-ops (approved):
LGTM — install-path escape fix validated E2E on a 16GB M3 (live, coordinator ready); 3-lane audit 0 C/H/M; both serializers + install.sh read-back covered; the only red is pre-existing advisory CONFORMANCE drift (requirements[9]/[19]) unrelated to this change.

PR #1304, opened 2026-08-31
Renew the discovery head's own target (freshness-only), not GitHub latest

Augustas11:
## Summary

Fixes the signed release-discovery **head renewal** so the coordinator-independent
self-heal rail does not go dark fleet-wide.

`renew-release-discovery-head.yml` is a **freshness-only** renewal: it re-signs a
strictly-greater sequence with a fresh expiry for the target the **current** signed head
already points at. Advancing the target to a newer release is a **separate,
coordinator-gated** workflow (`verify-live-coordinator-release-rollout.yml`).

**Bug:** the renewal resolved its target from GitHub `releases/latest` and rebuilt the
head against *latest's* manifests. When `latest` legitimately runs ahead of the signed
head — it does now: the live head targets **v1.8.111** while GitHub-latest is **v1.8.117**,
because advancing requires an operator coordinator flip that has not happened — the sign
step's `verify-release-discovery-transport.py --target-tag <latest>` and the
latest-vs-current `cmp` all fail closed. The Monday `0 16 * * 1` cron would therefore
**fail**, and the live head **expires 2026-09-04T23:14Z**, after which clients fail closed
(`discoveryHeadExpired`) and the signed self-heal rail is dark fleet-wide.

This is exactly the mutable-`latest` coupling **SPEC-020-R001** forbids ("Discovery MUST
begin from a signature-authenticated monotonic discovery head under the pinned release
trust root, not from mutable GitHub `latest` ordering"). SPEC-020's renewal prose had
mandated binding "the current latest immutable stable target" — that prose was the
normative root cause and is corrected here.

## Fix

- **`scripts/resolve-discovery-renewal-target.py`** (new) parses the **current** live
  transport head's `signed.target_compatibility_set_id` to resolve the exact `tag@commit`
  the head is bound to, with strict `re.fullmatch` validation and repository binding
  (fails closed on any malformed / off-repository target).
- **`renew-release-discovery-head.yml`** now: (1) selects the renewal base + append-only
  ceiling via the new selector, (2) downloads the current head and resolves **its own**
  target, (3) fetches **that** target release's `compatibility-set.json` +
  `compatibility-artifact-index.json` (via `releases/tags/<tag>`, never `releases/latest`)
  and re-validates it is a real, stable, immutable release bound to the exact commit the
  head names, then (4) re-signs. Succeeds even when GitHub-latest has moved.
- **`scripts/select-discovery-renewal-base.py`** (new) picks the renewal base from
  **public immutable** transports only (so a leftover draft is never the base), computes
  the ceiling across **both** published releases **and** git tag refs
  (`git ls-remote --tags`, collision-proof), and mints the **smallest** sequence above it
  (`ceiling + 1`, computed in Python and uint64-bounded — not signed-64-bit Bash) via the
  new `build-release-discovery-head.py --release-sequence`. It **fails closed unless the
  ceiling equals the public head**, so a renewal of the older target can never dominate a
  newer target's coordinator-gated rollout (which republishes a prebuilt, higher-sequence
  head). Verified against live state: current ceiling `2176955688681473` → mints
  `2176955688681474`; v1.8.117's prebuilt head is `2186103947198465`, so a later rollout
  still wins the client's greatest-sequence selection.

## Security invariants preserved (unchanged)

- ECDSA signature over canonical JSON verified with the pinned public key
  (`ops/pearl-updater/release-signing-public.pem`) before and after signing.
- Current head re-verified with `--allow-expired --require-immutable` against its own
  resolved target; new head verified with `--minimum-sequence`.
- Sequence monotonicity, 7-day expiry cap, canonical-JSON / duplicate-key rejection,
  append-only transport tags (no `--clobber`), and the immutable fixed `release-discovery`
  tag are untouched.
- `scripts/verify-release-discovery-transport.py` needed **no change** — it is already
  target-agnostic; the bug was purely the workflow's target resolution.

## Tests

- New `scripts/test-resolve-discovery-renewal-target.sh`: a **divergent-latest** target
  resolves cleanly; malformed set-id, short commit, non-semver tag, cross-repository
  target, missing `signed`, and duplicate-key inputs all fail closed.
- New `scripts/test-select-discovery-renewal-base.sh`: normal ceiling+1; a higher draft or
  orphan tag (incl. peeled `^{}` refs) fails the domination guard; uint64 exhaustion,
  missing-public, and missing-assets cases fail closed.
- `scripts/test-release-discovery-head.sh` extended for `--release-sequence` + mutual
  exclusion; `scripts/test-renew-release-discovery-head.sh` asserts no `releases/latest`
  target, current-head derivation, git-tag ceiling folding, and no signed-Bash arithmetic.
- Registered in `make test-dist`. Local: full discovery suite passes; workflow YAML
  validates; `py_compile` / `shellcheck` / `bash -n` clean; SPEC governance passes for
  SPEC-020.

## Audit

Full-fix diff reviewed by a three-lane codex audit (code / security / architecture);
**0 CRITICAL / 0 HIGH / 0 MEDIUM** on the final diff (one LOW test-coverage note closed
with committed cases). The audit caught and drove fixes for a sequence-domination HIGH (a
run-id-based renewal of the old target would leapfrog a newer release's earlier-built
prebuilt head) and two MEDIUMs (draft base selection, Bash uint64 overflow, tag collision).

## OPERATOR ACTION REQUIRED — deadline 2026-09-04T23:14Z

After this merges, an operator MUST run the fixed **Renew signed release discovery head**
workflow (with `production-release` approval) **before 2026-09-04T23:14Z**, or the live
signed head expires and the self-heal rail goes dark fleet-wide. This PR only fixes the
renewal logic; it does **not** sign, publish, or dispatch anything, and it does **not**
advance the target (that remains the coordinator-gated rollout's job).

Refs #1235.

{
  "behavior_change": "yes",
  "contract_change": "yes",
  "specs": ["SPEC-020"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "scripts/test-resolve-discovery-renewal-target.sh",
    "scripts/test-select-discovery-renewal-base.sh",
    "scripts/test-renew-release-discovery-head.sh"
  ],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/1235"
}

### DIFF
--- a/.github/workflows/renew-release-discovery-head.yml
+++ b/.github/workflows/renew-release-discovery-head.yml
@@ -1,9 +1,12 @@
 name: Renew signed release discovery head
 
 # Protected recurring renewal for append-only discovery transports.
-# Signs a strictly greater sequence for the current latest stable target and
-# publishes one new immutable prerelease. Never mutates existing transports or
-# the permanently immutable fixed release-discovery tag.
+# Freshness-only: signs a strictly greater sequence for the target the CURRENT
+# signed head already points at (resolved from the live transport, not from
+# GitHub "latest", which legitimately runs ahead until an operator rolls the
+# coordinator forward) and publishes one new immutable prerelease. Never
+# advances the target, mutates existing transports, or touches the permanently
+# immutable fixed release-discovery tag.
 
 on:
   workflow_dispatch:
@@ -68,7 +71,7 @@ jobs:
           "$sealed_bin" version | grep -E '^OpenSSL 3\.'
           printf 'bin=%s\n' "$sealed_bin" >> "$GITHUB_OUTPUT"
 
-      - name: Resolve latest stable target and current discovery ceiling
+      - name: Resolve the current head's own target and discovery ceiling
         id: target
         shell: bash
         env:
@@ -76,8 +79,14 @@ jobs:
           VALIDITY_HOURS_INPUT: ${{ github.event.inputs.validity_hours || '168' }}
         run: |
           set -euo pipefail
+          # Freshness-only renewal: keep the CURRENT signed self-heal head fresh
+          # for the exact target it already points at. The GitHub "latest" release
+          # legitimately runs ahead of the signed head until an operator rolls the
+          # coordinator forward (verify-live-coordinator-release-rollout.yml). This
+          # renewal must NOT advance the target, so it resolves the target from the
+          # current live transport's signed head, never from the GitHub latest API.
           work="$RUNNER_TEMP/renew-discovery"
-          mkdir -p "$work/latest" "$work/current"
+          mkdir -p "$work/target" "$work/current"
           [[ "$VALIDITY_HOURS_INPUT" =~ ^[1-9][0-9]*$ ]] || {
             echo "::error::validity_hours must be a positive integer" >&2
             exit 1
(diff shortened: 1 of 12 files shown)


antfleet-ops (approved):
Approving: discovery-head renewal cron fix. 3-lane audit clean (0/0/0). The red spec-index check is pre-existing corpus evidence-drift on main (also failing on last 3 main commits), unrelated to this diff which only touches SPEC-020 + discovery renewal scripts. Merge gated on swift test completing green.

Issue #1307, opened 2026-08-31
Provider CLI: artifact quantization is lost between recommendation and model adoption


erikHtoo:
## Goal

Ensure an actionable signed autotune recommendation uses the same artifact-aware RAM-fit decision when it is adopted, so the CLI does not offer an upgrade that immediately rejects itself.

Relates to the 16 GB physical provider journey. This is separate from interactive-context calibration in #1297.

## Reproduction environment

- Apple M4 Mac mini
- 16 GB unified memory
- macprovider-cli v1.8.109
- Current model: `meta-llama/llama-3.2-3b-instruct`
- Current state: `buyer_serving`

No credentials, provider tokens, account IDs, or host access details are included.

## Reproduction

Run a no-apply, installed-only recommendation:

```text
macprovider-cli autotune --recommend --check-only --installed-only \
  --no-submit-hardware-evidence --json
```

The signed recommendation selects:

- Catalog key: `meta-llama/llama-3.1-8b-instruct`
- Artifact model ID: `mlx-community/Meta-Llama-3.1-8B-Instruct-4bit`
- Proposed context: 4,000
- Reported memory headroom: 0 GB
- Eligibility: true

Pass that untouched recommendation to:

```text
macprovider-cli models adopt-recommendation --json \
  --recommendation-json recommendation.json
```

Observed result:

```json
{
  "type": "failed",
  "phase": "validating",
  "reason": "ram_unfit",
  "rollback_state": "not_needed",
  "target_model_id": "meta-llama/llama-3.1-8b-instruct"
}
```

The failure is safely before mutation: the config stays on 3B and the provider remains buyer-serving. The fail-closed behavior is correct; the contradictory actionable offer is not.

## Root cause

`ModelsAdoptRecommendationCommand` evaluates:

```swift
ModelFit.evaluate(modelID: recommendation.targetModelID, ...)
```

`targetModelID` is the canonical catalog key and does not contain the artifact's quantization marker. `ModelFit.estimateWeightSizeGB` therefore treats the 8B model as the full-precision fallback at two bytes per parameter, estimates roughly 16 GB of weights, and returns `wontFit` on a 16 GB Mac.

The signed recommendation also carries the exact artifact identity:

```text
mlx-community/Meta-Llama-3.1-8B-Instruct-4bit
```

That identity contains the quantization information used by the installed artifact and recommendation. The two stages are therefore applying RAM-fit logic to different identities.

## Implementation plan

[...]

PR #1317, opened 2026-09-01
Fix recommendation adoption RAM fit for signed artifacts

Augustas11:
## Summary

- Route `models adopt-recommendation` RAM validation through the signed candidate catalog row, matching the recommendation engine's `min_ram_gb` plus safety-margin rule.
- Preserve a test-only fallback that evaluates the signed artifact `model_catalog_model_id` so quantized artifacts are not re-sized as their display/catalog key.
- Add regression coverage for the reported 16 GB Llama 3.1 8B 4-bit adoption case and for signed catalog memory floors.

## Tests

- `cd phase3-binary && swift test --filter ModelsSubcommandTests`
- `cd phase3-binary && swift test --filter ModelFitTests`
- `git diff --check`
- Full `cd phase3-binary && swift test` is not green in this checkout: it first stops on an unrelated local MLX default metallib load failure in `StageForwardParityTests`.
- `cd phase3-binary && swift test --skip StageForwardParityTests` still reports unrelated coordinator compatibility-set / token-recovery failures; the touched test suite remains green.

{
  "issue": "https://github.com/Augustas11/macprovider/issues/1307",
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-023"],
  "authority_domains": ["installer-autotune-policy", "autotune-recommendation", "model-catalog-identity"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "cd phase3-binary && swift test --filter ModelsSubcommandTests",
    "cd phase3-binary && swift test --filter ModelFitTests",
    "git diff --check"
  ],
  "journeys": ["not-required"]
}

### DIFF
--- a/phase3-binary/Sources/macprovider-cli/ModelsSubcommand.swift
+++ b/phase3-binary/Sources/macprovider-cli/ModelsSubcommand.swift
@@ -423,8 +423,9 @@ struct ModelsAdoptRecommendationCommand: AsyncParsableCommand {
         guard !Self.hasBlockingWarning(recommendation.warnings) else {
             try fail("blocked_warning", exitCode: 2)
         }
+        let signedAuthority: RecommendationAdoptionAuthority?
         do {
-            try await Self.validateSignedAuthority(recommendation, configPath: config)
+            signedAuthority = try await Self.validateSignedAuthority(recommendation, configPath: config)
         } catch {
             try fail("signed_authority_invalid", exitCode: 2)
         }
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Approved via antfleet-ops. The fix is scoped to issue #1307 and targeted CLI tests cover the adoption RAM-fit regression; current CI failures are being inspected before merge.

PR #1336, opened 2026-09-02
Adopt Dependabot Go and Swift dependency bumps

Augustas11:
## Summary

Consolidates the low-risk, PR-CI-covered Dependabot version bumps into one
change. The two GitHub Actions **major-version** bumps were split into a
separate PR (checkout v4→v7, download-artifact 4.3.0→8.0.1) because PR CI does
not exercise the release/deploy/scheduled workflows where they mostly live and
they deserve independent revertability.

Each bump was reproduced from scratch (`go get` + `go mod tidy`,
`swift package resolve`) — the resulting lockfiles match Dependabot's branches.

| Supersedes | Bump |
|---|---|
| #1327 | coordinator `golang.org/x/net` 0.56.0 → 0.58.0 (pulls `golang.org/x/crypto` 0.54.0 → 0.55.0) |
| #1329 | coordinator `golang.org/x/text` 0.40.0 → 0.41.0 |
| #1332 | gateway `golang.org/x/text` 0.40.0 → 0.41.0 |
| #1331 | gateway `modernc.org/sqlite` 1.55.0 → 1.57.0 (indirect: `mattn/go-isatty`, `golang.org/x/sys`, `modernc.org/libc`) |
| #1333 | provider CLI `huggingface/swift-transformers` 1.0.0 → 1.3.3 |

The swift-transformers bump adds new transitive deps (swift-crypto, swift-asn1,
EventSource, yyjson); swift-jinja stays pinned at 2.4.2 and the stale pin
comment in `Package.swift` was updated from 1.0.0 to 1.3.3.

## Verification (local)

- `go build ./...` + `go vet ./...` — coordinator and gateway: clean
- `go test ./...` (true exit codes) — coordinator and gateway: pass, incl.
  money-path `billing`, `buyer`, `auth`, `requestlog`
- `make test-integration` — pass, incl. `TestSpec022V04StreamingSettlementReconcilerE2E`
- `internal/trustpool` constant-time gate — 3/3 green locally
- `swift build` — succeeds (pre-existing deprecation/Sendable warnings only);
  `Package.resolved` pin versions identical to Dependabot's #1333 branch

## CI note

- `spec-index / check` (advisory, not part of `ci-required`) will be red: the
  governance declaration validator forces `behavior_change: yes` for any
  non-governance path, which a dependency-only bump is not, and fabricating
  specs to satisfy it is disallowed. This matches every Dependabot PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1337, opened 2026-09-02
Bump GitHub Actions: checkout v4->v7, download-artifact 4.3.0->8.0.1

Augustas11:
## Summary

Splits the two **major-version** GitHub Actions bumps out of the Dependabot
dependency work (Go/Swift bumps are in #1336) so their risk is isolated and
they can be reverted independently.

| Supersedes | Bump |
|---|---|
| #1328 | `actions/checkout` v4 → v7 (45 SHA-pinned refs + 1 tag ref) |
| #1330 | `actions/download-artifact` 4.3.0 → 8.0.1 (6 SHA-pinned refs) |

SHAs match Dependabot's resolved pins. The checkout SHA is asserted by two dist
pin-guard scripts (`test-autotune-feed-freshness-alarm.sh`,
`test-renew-autotune-static-feed-signed.sh`); both updated to the new v7 SHA so
`make test-dist` stays green.

## Why a separate PR

- These are **major-version jumps** (`download-artifact` crosses four majors),
  which carry real breaking-change potential.
- Most refs live in **release, `malibu-release`, deploy, signing, renewal, and
  alarm** workflows that run on tags/schedule/dispatch — **not** on pull
  requests. PR CI does **not** exercise them, so a green check here does not
  validate the release/signing path.
- Isolating them keeps them independently revertible if a later release breaks,
  which a squashed bundle with the Go/Swift bumps would not allow.

## Before relying on this

Confirm the next release / signing / notarization run (and the scheduled
renewal/alarm workflows) actually succeed with checkout v7 and
download-artifact v8 — that is the coverage PR CI cannot provide.

## Verification (local)

- `make test-dist` — pass (incl. both updated pin-guard scripts)
- all 24 changed workflow YAMLs parse
- workflow diffs are action-ref lines only

### DIFF
--- a/.github/workflows/acceptance-candidate.yml
+++ b/.github/workflows/acceptance-candidate.yml
@@ -51,7 +51,7 @@ jobs:
       promotion_ready: ${{ steps.source.outputs.promotion_ready }}
     steps:
       - name: Checkout main-owned workflow controls
-        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
         with:
           ref: ${{ github.sha }}
           path: control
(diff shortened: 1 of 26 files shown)


antfleet-ops (approved):
GitHub Actions major bumps, isolated for revertability; ci-required green. make test-dist and YAML parse verified; release-path exercised only at next release.

PR #1356, opened 2026-09-03
Automate curl-channel install.sh republish on release

Augustas11:
## What

Automate republishing the public curl-channel installer to Pearl on every
non-prerelease release, so `https://get.malibu.tech/install.sh` no longer
silently lags behind the release it should serve.

## Why

`get.malibu.tech/install.sh` is a **static file on Pearl** (`/var/www/get/install.sh`)
that release automation never touched — it was hand-published. It had drifted
stale: at **v1.8.117** it kept serving the byte-identical **v1.8.115/116** copy,
missing the **#1296** escaped-path fix (`gsub(/\\\//,"/",value)` in three awk
parsers), so a fresh `curl … | bash` install on an escaped model-artifact path
(e.g. under `…/Application Support/…`) hit `die 30`. The `install-sh-parity-alarm`
workflow only *detects* this; nothing *fixed* it. (The channel was manually
resynced to v1.8.117 out of band; this PR prevents the recurrence.)

## Change

- **`scripts/publish-install-sh.sh`** (new) — reuses the existing
  `MALIBU_DOWNLOAD_SSH_KEY` Pearl root deploy key + pinned
  `scripts/dist/malibu-download-known_hosts`; backs up the current file,
  atomically installs the released copy, and verifies both the on-disk and the
  served bytes. It is **release-bound**: publishing requires `--tag`, refuses
  unless the local `phase3-binary/dist/install.sh` is byte-identical to that
  tag's committed copy, and refuses unless `gh` confirms the tag is GitHub's
  current *latest stable* release (fetches the tag fresh from origin; repo
  pinned; all validation runs before any success/no-op exit). It is idempotent
  (no-op when already in sync) and safe to run standalone as recovery.
- **`release.yml`** (`sign_publish`) — a `Republish curl-channel install.sh to
  Pearl` step, gated `needs.build.outputs.prerelease == 'false'`, modeled on the
  adjacent `Publish Malibu download alias to Pearl` step. Marked
  `continue-on-error: true` so a transient Pearl republish flake cannot fail an
  otherwise-good, already-published release — the parity alarm is the detector
  and the idempotent script is the standalone retry.

## Testing / audit

- Behavioral tests (all pass): byte-mismatch rejects, missing `--tag` rejects,
  non-latest tag (`v1.8.116`) rejects with "GitHub latest = v1.8.117",
  `releases/latest` correctly excludes the `v1.8.118` prerelease, matching
  tag + latest passes then no-ops, `--check-only` works read-only.
- Three-lane review (code / security / architecture) run to **0 CRITICAL /
  0 HIGH / 0 MEDIUM** over the full diff, across three fix rounds. The security
  HIGH (env-var command injection into the remote root shell) was closed by
  hardcoding the webroot + URL; the reentrancy MEDIUM by `continue-on-error`;
  and the "recovery could publish non-released bytes" MEDIUM by the `--tag`
  release-binding above.

## Governance note

`behavior_change: none` is the honest classification: this is release/publishing
infrastructure that makes the implementation honor the **existing** SPEC-003
contract ("`install.sh` is served from `main` via `get.malibu.tech`") — it adds
no product behavior and changes no requirement. The advisory
`check_spec_pr_declaration` validator has no passing declaration for a
non-governance release-infra change like this: `none` is rejected for
non-governance paths, and SPEC-003's only CONFORMANCE requirement (`SPEC-003-R001`)
is the onboarding-*skill* publication, not `install.sh`, so mapping to it would
be inaccurate (and AGENTS.md forbids fabricating a mapping to satisfy the
checker). Declaring honestly and accepting the advisory flag.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1986,6 +1986,38 @@ jobs:
             compatibility-artifact-index.json checksums.txt checksums.txt.sig \
             release-provenance.json macprovider-release-discovery.json
 
+      - name: Republish curl-channel install.sh to Pearl
+        if: needs.build.outputs.prerelease == 'false'
+        # sign_publish is not reentrant (it created the immutable release earlier
+        # in the job), so this post-publication Pearl step must not be able to
+        # fail an otherwise-good, already-published release. If the republish
+        # flakes, the install-sh-parity-alarm catches the drift and the
+        # idempotent scripts/publish-install-sh.sh reruns it standalone.
+        continue-on-error: true
+        env:
+          MALIBU_DOWNLOAD_VPS_HOST: ${{ secrets.MALIBU_DOWNLOAD_VPS_HOST }}
+          MALIBU_DOWNLOAD_SSH_KEY: ${{ secrets.MALIBU_DOWNLOAD_SSH_KEY }}
+          GH_TOKEN: ${{ github.token }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          # The public `curl -fsSL https://get.malibu.tech/install.sh | bash`
+          # channel is a static file on Pearl that nothing republishes on a
+          # release; it has silently lagged behind (e.g. missing the #1296
+          # escaped-path fix at v1.8.117, so a fresh install could hit die 30).
+          # Republish the released copy so the curl channel matches this release
+          # and the install-sh-parity-alarm stays green.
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Approved: 3-lane review (code/security/architecture) to 0 C/H/M; release-bound publish + idempotent recovery. Advisory spec-index declaration red is the known release-infra gap (documented in PR body).

Issue #1354, opened 2026-09-02
Warmup-gate deadlock on fresh-provider onboarding (config-gated) + ws:// silent-refuse + isolated-coordinator config foot-guns


Augustas11:
## Summary

A fresh provider deadlocks during onboarding when the coordinator's warmup gate
is enabled. Found by @SmtTheSE during the #1346 candidate E2E.

**This is config-gated and not a regression.** Production runs
`pool.warmup_gate_enabled: false`, and the CLI-side readiness wait involved has
shipped since **v1.8.117**, so the live fleet is unaffected and the current
release candidate is **not blocked** by it. But the binary default ships the
gate **on**, so any coordinator stood up from defaults (test/staging, a new
contributor, or prod if the gate is ever enabled) will deadlock fresh providers.
Fix on its own track behind the three-lane audit gate — this touches
`phase4-coordinator/internal/ws/` and the CLI handshake, both sensitive paths.

## The deadlock (only when `warmup_gate_enabled: true`)

1. Coordinator registers the provider as `StateDegraded` and starts the warmup
   gate, dispatching an `inference_request` over the provider WS —
   `phase4-coordinator/internal/ws/server.go:2818` (`initialState = StateDegraded`
   iff `WarmupGateEnabled`).
2. The CLI is not reading yet. `CoordinatorClient.connectAndRun` awaits
   `waitForCoordinatorServingCapability()` (~30s polling `/v1/pool/check`)
   **before** it installs `inferenceRelay` and starts the read loop. The warmup
   probe is never read; even if it were, `inferenceRelay` is still nil at that
   point, so it would be NAK'd `unknown_message_type`.
3. The safety net declines while a gate is pending: `markDegradedForWarmup`
   early-returns if `warmupGatePending` (`server.go:5510`), and other fallbacks
   gate on `warmupGatePending` (`server.go:4266`).
4. CLI gives up → `coordinator buyer-serving readiness was not confirmed` →
   closes the socket → warmup relay fails → reconnect → loops forever.

**Evidence (@SmtTheSE):** coordinator instrumentation shows the
`inference_request` frame written successfully; the CLI's
`MACPROVIDER_KEEPALIVE_DEBUG=1` log for the same window shows **no** matching
`ws_recv`. Setting `warmup_gate_enabled: false` took the provider to
`serving_buyers` / `routing_eligible: true` immediately, on the next connect.

## Why it is config-gated / not a regression

[...]

PR #1357, opened 2026-09-03
Make warm-up capability gate fail-open observe-only; fix onboarding deadlock (#1354)

Augustas11:
## Summary

Fixes the fresh-provider onboarding deadlock from issue #1354 by making the
FR-P8a admission warm-up gate **fail-open and observe-only** (SPEC-002 v1.6.0),
so onboarding is **never blocked** by the fitness probe — the explicit product
requirement for this change.

### The problem

The warm-up gate admitted a newly connected provider as `StateDegraded` (when
`pool.warmup_gate_enabled: true`) and withheld it from buyer routing until a
token-producing WS probe passed. The released provider CLI, however, awaits
coordinator buyer-serving readiness (`/v1/pool/check`) **before** it installs
its inference relay and starts reading the WebSocket — so the probe frame was
never answered, the provider could never leave `degraded`, readiness never
confirmed, and the CLI reconnect-looped forever.

Production runs the gate `false`, so the live fleet is unaffected — but
`config.Default()` ships it `true`, so any coordinator stood up from defaults
(test/staging, a new contributor) bricks fresh providers. Critically, the gate
runs at **admission on every `hello`**, so enabling it would also drain the
existing fleet out of routing as providers reconnect (deploys, updates, blips).

### The fix (fail-open, observe-only)

A provider that clears the trust/identity admission gates is now admitted
`ready` and immediately buyer-routable. The warm-up probe still runs in the
background and records a fitness verdict for telemetry / connection-event
history, but it **never** withholds, degrades, or marks a provider
`unavailable`, and heartbeat / `state_update` `ready` reports are no longer
clamped to `degraded` while a probe is in flight. Because the gate can no longer
block onboarding, `config.Default().Pool.WarmupGateEnabled` stays `true` safely
(no footgun, no change to the normative default).

**Scope guard — only the fitness verdict is fail-open.** Trust/identity gates
still withhold routing independently and were verified end-to-end (security
lane): catalog/autotune `hello` admission, model-hash + `require_hash_verified`,
attestation, encrypted-leg, admission ceiling/stale/sandboxed, and the
model-version floor all still exclude a provider at buyer-routing time
(`buyer/server.go` `handlePoolCheck` and candidate selection via
`tier2ProviderExcludedForConfig`) and at admission (`ws/server.go`). A provider
that genuinely cannot serve is removed by the in-flight circuit breaker
(FR-P11a) on real buyer traffic. The wake-from-sleep `warm_up` fallback (FR-P8)
is unchanged and now runs independently of the admission probe.

### Also in this PR

- **S2 — `ws://` no longer silently refused.** `CoordinatorClient.init?` now
  accepts `ws://` for loopback coordinators only (localhost/127.0.0.1/::1),
  matching `CoordinatorReadinessClient`, and emits an explicit stderr diagnostic
  instead of a silent `nil` (which surfaced as a misleading "check this Mac's
  internet connection"). Also rejects hostless `wss:///` and embedded userinfo.
- **SPEC-004 (smart router) reconciled to v0.3.4** so FR-P8a admission warm-up
  is no longer described as a routing-eligibility gate anywhere; wake-from-sleep
  `warm_up` degradation is attributed to FR-P5.

### Deferred (not onboarding blockers)

The issue's S5 config foot-guns are intentionally **not** in this PR: S5a
(compat-set `accepted_ids ≥ 2`) is intentional rollback-safety with a clear
error, not a bug; S5b (global YAML strict-decode) is genuinely risky on the live
money-path config loader and should not be rushed in.

### Known accepted limitation

Under fail-open, a provider that returns only well-formed empty completions
(`finish_reason: "stop"`, zero tokens) can remain routable: FR-P11a
deliberately does not count a clean empty `stop` as a fault. This is the
accepted trade-off of "never block onboarding on a fitness signal"; SPEC-002
documents it and notes a future bounded quality-path.

## Audit

Three-lane codex audit (code / security / architecture) over the full fix diff,
to the repo bar of 0 CRITICAL / 0 HIGH / 0 MEDIUM. Security lane: 0/0/0/0 and
independently verified trust-gate independence, ws:// masquerade resistance, the
redirect-not-followed property, and billing/quota interaction. Code and
architecture MEDIUM findings (wake-path decoupling, breaker-claim honesty,
stale SPEC-004/SPEC-002 wording, a weak timeout test, CLI URL hardening) were
all resolved and re-audited to 0 C/H/M.

## Testing

- `make test` (coordinator, gateway, integration, dist)
- `cd phase3-binary && swift test`
- Full `internal/ws` suite incl. rewritten warm-up/wake tests; targeted
  `internal/buyer` trust/routing tests.

{
  "issue": "https://github.com/Augustas11/macprovider/issues/1354",
  "behavior_change": "yes",
  "contract_change": "yes",
  "specs": ["SPEC-002", "SPEC-004"],
  "requirements": ["SPEC-002-R002"],
  "authority_domains": ["coordinator-admission-routing"],
  "arbitration": ["SPEC_BUG"],
  "tests": ["phase4-coordinator/internal/ws/server_test.go:TestWarmupProbeAdmitsProviderReadyAndObservesOnly"],
  "journeys": ["not-required"]
}

### DIFF
--- a/phase3-binary/Sources/macprovider-cli/CoordinatorClient.swift
+++ b/phase3-binary/Sources/macprovider-cli/CoordinatorClient.swift
@@ -454,7 +454,28 @@ actor CoordinatorClient {
         guard let rawURL = config.coordinatorURL, let url = URL(string: rawURL) else {
             return nil
         }
-        guard url.scheme == "wss" else {
+        // #1354 S2: production coordinators require `wss`, but an insecure `ws`
+        // scheme is allowed for loopback hosts only (localhost/127.0.0.1/::1) so
+        // isolated test/staging coordinators can be reached — matching the
+        // loopback exemption already honored by CoordinatorReadinessClient. Any
+        // other rejection is reported explicitly instead of a silent nil, which
+        // previously surfaced as a misleading "check this Mac's internet
+        // connection" status with no log line.
+        let scheme = url.scheme?.lowercased()
+        let host = url.host?.lowercased()
+        let isLoopback = host == "localhost" || host == "127.0.0.1" || host == "::1"
+        switch scheme {
+        case "wss":
+            break
+        case "ws" where isLoopback:
+            break
+        case "ws":
+            FileHandle.standardError.write(Data(
+                "FATAL coordinator URL uses insecure ws:// for non-loopback host \(host ?? "?"); use wss:// (ws:// is permitted only for localhost/127.0.0.1/::1)\n".utf8))
+            return nil
+        default:
+            FileHandle.standardError.write(Data(
+                "FATAL coordinator URL scheme \(scheme ?? "nil") is not supported; expected wss:// (or ws:// for a loopback coordinator)\n".utf8))
             return nil
         }
         self.coordinatorURL = url
--- a/phase3-binary/Sources/macprovider-cli/CoordinatorClient.swift
+++ b/phase3-binary/Sources/macprovider-cli/CoordinatorClient.swift
@@ -461,8 +461,22 @@ actor CoordinatorClient {
         // other rejection is reported explicitly instead of a silent nil, which
         // previously surfaced as a misleading "check this Mac's internet
         // connection" status with no log line.
-        let scheme = url.scheme?.lowercased()
-        let host = url.host?.lowercased()
+        //
+        // Parse with URLComponents so we can reject embedded userinfo and a
+        // missing host, matching CoordinatorReadinessClient.readinessURL — a bare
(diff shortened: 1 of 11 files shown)


antfleet-ops (approved):
Approved by antfleet-ops on behalf of the operator. Three-lane audit 0 C/H/M (security independently verified trust-gate independence, ws:// masquerade resistance, redirect property); make test green; governance manifest synced to SPEC-002 v1.6.0 / SPEC-004 v0.3.4. Fail-open observe-only warm-up gate resolves the #1354 onboarding deadlock without weakening trust gates.

Issue #1360, opened 2026-09-03
Package.resolved out of sync under Xcode 16.4 — locked release builds fail (async-http-client unpinned, #1336 swift-transformers bump)


Augustas11:
## Summary

`phase3-binary/Package.resolved` is out of sync with `Package.swift`'s dependency
graph **under the pinned release toolchain (Xcode 16.4)**. The locked release /
acceptance-candidate build (`-onlyUsePackageVersionsFromResolvedFile`) fails
because `async-http-client` is required by the graph but not pinned:

```
xcodebuild: error: Could not resolve package dependencies:
  an out-of-date resolved file was detected at .../phase3-binary/Package.resolved,
  which is not allowed when automatic dependency resolution is disabled;
  ... dependencies were added: 'async-http-client'
```

## Root cause

**#1336** (`4e54ea64`, Adopt Dependabot Go and Swift dependency bumps) bumped
`swift-transformers` **1.0.0 → 1.3.3**. Under Xcode 16.4, that graph transitively
requires `async-http-client`, but the `Package.resolved` regenerated in #1336 was
produced under a *newer* toolchain that does not pull it in, so the pin is missing.

## Impact

- **Locked / release / acceptance-candidate builds are broken** on `main`.
  First surfaced cutting the 1.8.119 acceptance candidate
  (run 33749166183, step "Build candidate CLI compatibility set").
- **Regular CI stays green** because it uses *automatic* resolution (adds the pin
  on the fly), which is why this was not caught at #1336 merge — exactly the
  "money-path dep-upgrade smoke" gap.

## Fix

Regenerate `Package.resolved` under **Xcode 16.4** (the release toolchain) so the
`async-http-client` pin is present, and commit it. Cannot be reproduced on newer
toolchains (Xcode 26.x resolves the graph without it).

Being fixed for the 1.8.119 candidate via a one-off resolve job on the 16.4
runner; this issue tracks the underlying `main` regression + the missing dep-lock
verification that let it merge.

## Follow-up

- Add a CI check that the locked resolution (`-onlyUsePackageVersionsFromResolvedFile`)
  succeeds under the release toolchain, so a future dep bump cannot merge with an
  incomplete `Package.resolved`.

Augustas11:
Fixed on main by https://github.com/Augustas11/macprovider/pull/1367 (`8b4f7ba0`).

[...]

PR #1361, opened 2026-09-03
Add resolve job for phase3-binary Package.resolved under release toolchain (#1360)

Augustas11:
## What

Add a `workflow_dispatch` job that regenerates `phase3-binary/Package.resolved`
under the **pinned release toolchain (Xcode 16.4)** and uploads it as an artifact.

## Why

The locked candidate/release build (`-onlyUsePackageVersionsFromResolvedFile`)
fails because `async-http-client` is required by the graph under Xcode 16.4 but
not pinned — a `Package.resolved`/`Package.swift` drift introduced by #1336
(swift-transformers 1.0.0 → 1.3.3). See **#1360**. Newer local toolchains
(Xcode 26.x) resolve the graph without it and cannot reproduce the correct lock,
so the lock must be regenerated on the release toolchain. This job does that;
it uploads the corrected lock (does not commit), and is reusable for future
dep bumps.

## Governance note

`behavior_change: none` — CI tooling only; no product behavior, spec, or contract
change. The advisory `spec-index` declaration check has no passing declaration for
a non-governance CI-path change (same as the install.sh and version-bump PRs); it
is not `ci-required` and does not block merge.

### DIFF
--- a/.github/workflows/resolve-package-lock.yml
+++ b/.github/workflows/resolve-package-lock.yml
@@ -1,0 +1,47 @@
+name: Resolve CLI Package.resolved (release toolchain)
+
+# Regenerate phase3-binary/Package.resolved under the pinned release toolchain
+# (Xcode 16.4), so the lock matches what the locked candidate/release build
+# (`-onlyUsePackageVersionsFromResolvedFile`) requires. Newer local toolchains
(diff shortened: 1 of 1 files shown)


antfleet-ops (approved):
CI helper to regenerate Package.resolved under the pinned release toolchain (#1360). Workflow-only, no runtime change.

Issue #966, opened 2026-08-09
Evaluate swift-transformers 1.3.3 as a separate token-exact migration


Augustas11:
## Goal

Evaluate `swift-transformers` independently from the MLX engine upgrade. Macprovider currently pins `1.0.0`; upstream latest is `1.3.3`.

This must not be silently bundled into the `mlx-swift-lm` pin bump because tokenizer changes can alter prompt tokens, receipt accounting, cache identity, stop behavior, and tool-template rendering.

## Candidate improvements

- `1.3.2`: BPE/Unigram and pre-tokenizer performance improvements.
- `1.3.3`: Unicode-scalar correctness fixes for BPE, Unigram, Bert NFD, and Japanese voiced-kana handling.

## Token-exact release gate

- [ ] Capture the current `1.0.0` token IDs, decoded text, and byte counts for the committed fixture corpus.
- [ ] Compare `1.3.3` on ASCII, emoji/grapheme clusters, combining marks, CJK/Japanese, code, JSON, and model special tokens.
- [ ] Classify every token-ID delta as intended upstream correctness or a blocker.
- [ ] Re-run chat-template/tool-schema fixtures for Qwen, GPT-OSS, Nemotron, and Llama-family rows.
- [ ] Re-run prompt/completion token accounting and receipt fixtures.
- [ ] Recompute tokenizer configuration/cache identity hashes where bytes or behavior change.
- [ ] Benchmark long-input tokenizer latency without changing model or MLX pins.
- [ ] Ship as a dedicated dependency PR with exact pin and `Package.resolved` revision.

## Do not

- Do not update `swift-transformers` in the same commit as `mlx-swift-lm`.
- Do not accept decode-text equality as token-ID parity.
- Do not preserve old Unicode bugs merely for parity; document intentional corrections and invalidate incompatible caches.

## References

- https://github.com/huggingface/swift-transformers/releases/tag/1.3.2
- https://github.com/huggingface/swift-transformers/releases/tag/1.3.3
- `phase3-binary/Package.swift`

Augustas11:
Implementation is in #967. Production pins remain unchanged; the PR adds fail-closed monitoring/runtime guards and keeps this tracker open until its upstream correctness/release gates pass.

Augustas11:
## Upstream refresh — 2026-08-25

| Tracker | Live state |
|---|---|
| huggingface/swift-transformers latest tag | still **`1.3.3`** (published 2026-05-16) |
[...]

PR #1372, opened 2026-09-04
deps(provider-cli): swift-transformers 1.3.3 → 1.3.4 under #966 token-exact gate

Augustas11:
## Summary

Advances the provider-CLI `swift-transformers` exact pin **1.3.3 → 1.3.4** under
the issue #966 token-exact gate.

The `1.0.0 → 1.3.3` bump had already landed on `main` via Dependabot
consolidation PR #1336 (`4e54ea64`) without the #966 gate, and is not yet in any
released tag. This runs that gate over the full `1.0.0 → 1.3.4` delta (1.3.3 and
1.3.4 tokenizer bytes are identical — 1.3.4 adds only `AutoTokenizer` revision
support, `Hub.Repo` Sendable, jinja→2.4.2, and a sampler fix in a path MLX does
not use) and lands at 1.3.4.

`mlx-swift-lm` / `mlx-swift` are frozen. `Package.resolved` changes only the
`swift-transformers` entry; `swift-huggingface` stays at `main`'s `0.9.0`.

## Token-exact gate — PASS

Full method + raw per-model outputs: `audits/2026-09-04-swift-transformers-134/GATE.md`.
Minimal SwiftPM harness importing only `Tokenizers`, built at 1.0.0 and 1.3.4,
encoding a 20-row Unicode corpus (ASCII, emoji/grapheme clusters, combining
marks, CJK/Japanese incl. voiced + half-width kana, code, JSON, special tokens)
against each model's real `tokenizer.json`. Comparison is on token IDs.

| Model (catalog family) | tokenizer | 1.0.0 → 1.3.4 token IDs |
|---|---|---|
| Qwen3-8B | BPE + NFC | **identical** |
| Qwen2.5-Coder-7B | BPE + NFC | **identical** |
| Llama-3.2-3B | BPE (byte-level) | **identical** |
| GPT-OSS-20B | BPE (o200k) | **identical** |
| Gemma-4-26B | BPE + metaspace | **5/20 rows — intended Unicode corrections** |
| Nemotron-3-Nano-30B | BPE (`TokenizersBackend`) | 1.0.0 cannot load; 1.3.4 loads (new capability) |

Every byte-level BPE catalog family is token-exact. Gemma's 5 deltas are all
intended upstream Unicode corrections — two fix genuine 1.0.0 bugs (a dropped
skin-tone emoji `👌🏻`; a leaked `▁` metaspace artifact in `vs▁café`). Per #966,
these corrections are accepted and old bugs are not preserved for parity.

**Cache invalidation:** Gemma prefill/KV caches keyed on 1.0.0 token IDs for
Unicode-heavy inputs must be invalidated. Byte-level families are unaffected.
**Billing:** Gemma prompt/completion counts shift only on Unicode-heavy inputs,
only downward (1.0.0 over-counted); 1.3.x has not shipped, so no in-flight
receipts assumed 1.0.0 Gemma counts.

Long-input tokenizer latency improves (Qwen ~1.9×, Gemma ~20×). `macprovider-cli`
builds clean against 1.3.4 (`swift build`, exit 0).

## Verification

- `swift build` (phase3-binary) at 1.3.4 — clean.
- Token-ID parity captured 1.0.0 vs 1.3.4 across 6 catalog tokenizer families.
- `swift package resolve` keeps `swift-huggingface` 0.9.0 and prunes only inert
  orphan pins carried by #1336 (async-http-client / swift-xet / nio stack — no
  live dependency references them). This PR keeps the minimal tokenizer-only
  `Package.resolved` diff; orphan pruning is left to a separate hygiene PR.

## Governance note

This is a dependency-manifest bump, mirroring #1336, which merged with no
`SPEC-GOVERNANCE` declaration block and an advisory-red `spec-index / check`
while `ci-required` gated the merge. No SPEC, AUTHORITY, CONFORMANCE, or
product-behavior path changes here, so no declaration block is added and the
advisory `check` is expected red as on #1336.

Refs #966


Augustas11:
## 3-lane codex audit (full PR diff)

Bar: 0 CRITICAL / 0 HIGH / 0 MEDIUM. **Result: PASS.**

- **Code lane:** 0 C/H/M. LOW — gate reproduction doesn't pin exact model snapshot revisions/hashes. INFO — pre-existing orphan pins in `Package.resolved`.
- **Architecture lane:** 0 C/H/M. Confirmed landing at 1.3.4 (vs 1.3.3) is sound; the Gemma KV impact is a transient cache miss, not corruption (see adjudication below).
- **Security lane:** raised one **MEDIUM** — *KV disk-cache envelope keys identity on model/catalog/tokenizer-file hashes + MLX revision but not the `swift-transformers` library revision, so a 1.0.0-era Gemma cache could be promoted post-upgrade for a different token stream.*

### Adjudication of the security MEDIUM — refuted by code (not a valid finding)

The security lane assumed a promoted cold-cache entry is reused based on envelope identity alone. It is not. Reuse is gated on the **live incoming token IDs**:

- `ConversationCache.begin()` computes `lcp = longestCommonPrefix(entry.canonicalPromptTokens, incomingTokens)` (`ConversationCache.swift:225`) and misses on `prefix_diverged` when `lcp < lcpThreshold`.
- A promoted cold candidate is fed through this **same** predicate — no identity-only second path (`ConversationCache.swift:166`, `:172`).
- The cached KV is **trimmed to exactly `lcp`** (`ConversationCache.swift:239`), so only positions whose token IDs are byte-identical between the cached and live sequences are ever reused; `cached_prompt_tokens` billed = `lcp`.

Consequence for the Gemma scenario: a stale 1.0.0 Gemma cache whose Unicode-heavy token IDs differ from the 1.3.4 tokenization prefix-diverges → cache **miss**, or reuses only the identical leading token IDs (correct by construction). No wrong continuation state, no mismatched receipt accounting. Worst case is a performance-level cache miss. The architecture lane independently reached the same conclusion.

Adding `swift_transformers_revision` to the envelope identity (next to the existing `mlx_swift_lm_revision` / `mlx_version`) remains a reasonable **LOW** defense-in-depth hardening for the KV subsystem (SPEC-037), tracked separately — it is not a blocker for this bump.

LOW/INFO findings are carried per policy.

antfleet-ops (approved):
Approved: #966 token-exact gate; 3-lane audit clean; ci-required green.

Issue #1233, opened 2026-08-27
SPEC-043 live-readiness before Trusted Pool announcement


Augustas11:
## Context

Follow-up after https://github.com/Augustas11/macprovider/issues/1160.

Issue 1160 covered SPEC-043 creator-MVP evidence and the production promotion gate (Gates A–D). That work is done:

- signed isolated-candidate `JOURNEY-TRUSTED-POOL-CREATOR-MVP` envelope (workflow `33037491210`, `run_id` `1` on `fd509043`)
- consumed sibling `PoolPromotionTransitionV1` `spec043-ppt-20260827T032524Z-1` (workflow `33039343666`)
- SPEC-043-R001 through R012 are **conformant**
- spec-level `implementation_status` stays **partial**, `production_status` stays **not-deployed**, spec-level `evidence[]` stays empty

Isolated-candidate CONFORMANCE is **not** a live Trusted Pool launch. Buyer-facing wording stays **Trusted Pools**. Do not call this a Privacy Pool.

## Remaining before any live announcement

Ops (not more evidence-gate PRs):

- [ ] Store a signed production on-call readiness record in the launch environment (`POST /admin/trust-pools/on-call-readiness`, CLI `coordinator-cli trust-pool-oncall`). Operator HTTP/CLI production promote already fail-closes without a current row (#1232).
- [ ] Assign a reviewed-artifact lifecycle owner on the live pool (`POST /admin/trust-pools/pools/{id}/reviewed-artifact-lifecycle`). Separate from digest-bound reviewed-artifact upsert.
- [ ] Run a production timing-floor remeasure with `scripts/measure-pool-rejection-timing-floor.py --environment production --allow-production`. Isolated-candidate oracle proof is not that remeasure. Offline `--samples-json` must not be treated as a production remeasure.

Code / spec (optional or later, still launch blockers):

- [ ] Wire on-call into mapped `Store.PromotePool` / `validatePromotion` (needs recapture+sibling+consume; HTTP/CLI is already gated).
- [ ] Optional: fail-close operator HTTP production promote when reviewed-artifact lifecycle owner is missing (same unmapped wrapper pattern as #1232).
- [ ] Immutable custody-class recording from the R002 registration.
- [ ] Full R006 re-verification before reactivation.
- [ ] Unresolved SPEC-042 conformance rows, or an explicit decision to accept them as launch blockers.

Until announce:

- keep `trusted_pools.enabled` off, or any enabled pool in a non-routeable lifecycle
[...]

PR #1232, opened 2026-08-27
Fail-close operator HTTP production promote without current on-call

Augustas11:
## Summary
- Operator HTTP/CLI `POST /admin/trust-pools/pools/{id}/promote` fail-closes for non-candidate launch environments when the matching on-call readiness row is missing or expired.
- Candidate promote stays ungated so the isolated-candidate journey remains valid. In-process `Store.PromotePool` is still ungated; wiring mapped `validatePromotion` needs a recapture window.
- Isolated-candidate CONFORMANCE is not a live Trusted Pool launch. Related to https://github.com/Augustas11/macprovider/issues/1160. Keep issue 1160 open.

## Test plan
- [x] `cd phase4-coordinator && go test ./internal/trustpool`
- [x] `python3 scripts/check_spec_governance.py --root . --base-ref origin/main`
- [x] Three-lane `omc ask codex` (code/security/architect): 0 CRITICAL / 0 HIGH / 0 MEDIUM

SPEC-GOVERNANCE-DECLARATION-BEGIN
{
  "schema_version": "spec-pr-governance-v1",
  "behavior_change": "yes",
  "contract_change": "yes",
  "specs": ["SPEC-043"],
  "requirements": ["SPEC-043-R008", "SPEC-043-R011"],
  "authority_domains": ["trusted-pool-creator-onboarding"],
  "arbitration": ["DECISION_REQUIRED"],
  "tests": [
    "phase4-coordinator/internal/trustpool/live_readiness_test.go:TestRequireOnCallReadinessForPromotion",
    "phase4-coordinator/internal/trustpool/live_readiness_test.go:TestAdminHandler_PromoteCandidateSkipsOnCallReadiness",
    "phase4-coordinator/internal/trustpool/live_readiness_test.go:TestAdminHandler_PromoteProductionRequiresOnCallReadiness",
    "phase4-coordinator/internal/trustpool/live_readiness_test.go:TestAdminHandler_PromoteProductionRejectsExpiredOnCallReadiness"
  ],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/1160"
}
SPEC-GOVERNANCE-DECLARATION-END

PR #1387, opened 2026-09-05
trustpool: fail-close operator promote without production reviewed-artifact lifecycle owner (#1233)

Augustas11:
## What

Fail-close operator HTTP/CLI production `promote` when a Trusted Pool has no
current **production reviewed-artifact lifecycle owner**. This is a SPEC-043
live-readiness launch blocker tracked in #1233.

Operator production promote already fail-closes without a current on-call
readiness row (#1232). The reviewed-artifact lifecycle owner — also a
SPEC-043-R008 production-enablement precondition ("reviewed creator
distribution package ... artifact digest") — was **not** gated, so production
promote could succeed with no lifecycle owner assigned.

## How

- Add `Store.RequireReviewedArtifactLifecycleForPromotion`, mirroring
  `RequireOnCallReadinessForPromotion`: an **unmapped wrapper** gate in
  `handlePromotePoolGuarded`, outside the evidence-mapped
  `Store.PromotePool` / `validatePromotion` transaction, so **no CONFORMANCE
  recapture** is triggered. It fail-closes a non-candidate pool promote when
  the production lifecycle owner is missing, candidate-class, or overdue.
  `ErrReviewedArtifactLifecycle` already maps to
  `409 reviewed_artifact_lifecycle_rejected`.
- Both live-readiness gates in the wrapper now fail closed on **any** non-nil
  error, not only their sentinel. Previously a non-sentinel read error
  (missing table, malformed row, storage error) fell through to the mapped
  promote, which does not re-check these rows — a real fail-open. This also
  closes the identical pre-existing latent fail-open on the on-call gate in
  the same wrapper. (Found and fixed during the three-lane audit below.)
- Runbook updated to document the new gate and its mapped-vs-unmapped boundary.

Candidate pools and in-process `Store.PromotePool` are intentionally unchanged;
wiring the mapped `validatePromotion` stays deferred to a recapture window.

## Tests

- `TestRequireReviewedArtifactLifecycleForPromotion` (candidate skips, missing
  fails closed + `PromotePool` still succeeds, candidate-class fails closed,
  current production allows, overdue fails closed, unknown pool passthrough)
- `TestAdminHandler_PromoteProductionRequiresReviewedArtifactLifecycle` behavior
  folded into the existing on-call admin test + overdue and corrupt-row admin
  tests (`...RejectsOverdueReviewedArtifactLifecycle`,
  `...FailsClosedOnUnreadableLifecycle`)
- `go test ./internal/trustpool -count=1` green (incl. the candidate journey)

## Audit

Three-lane codex audit over the full fix diff, second round after the fail-open
fix: **0 CRITICAL / 0 HIGH / 0 MEDIUM** across code, security, and architecture
lanes. One LOW carried (security): the gate does not re-validate `owner != ''`
from a directly DB-tampered row; the upsert path already enforces non-empty
owner, so this is only reachable by bypassing the write path.

### DIFF
--- a/docs/runbooks/trusted-pool-creator-launch.md
+++ b/docs/runbooks/trusted-pool-creator-launch.md
@@ -76,7 +76,12 @@ Also still missing before any live announcement:
 - a reviewed-artifact lifecycle owner for the live pool
   (`POST /admin/trust-pools/pools/{id}/reviewed-artifact-lifecycle`, CLI
   `coordinator-cli trust-pool-artifact-lifecycle`). This is separate from the
-  digest-bound reviewed-artifact upsert.
+  digest-bound reviewed-artifact upsert. Operator HTTP/CLI production promote
+  now fail-closes without a current production lifecycle owner (missing,
+  candidate-class, or overdue), the same unmapped-wrapper gate as on-call.
+  In-process `Store.PromotePool` still does not consult it; wiring
+  `validatePromotion` needs a recapture window because that function is
+  evidence-mapped.
 - a production timing-floor remeasure. Isolated-candidate oracle proof is not
   that remeasure. Use `scripts/measure-pool-rejection-timing-floor.py`; it
   refuses production hosts unless `--environment production --allow-production`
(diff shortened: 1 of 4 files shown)


antfleet-ops (approved):
Approving on behalf of antfleet-ops. Reviewed: reviewed-artifact lifecycle promote gate mirrors the on-call unmapped-wrapper pattern (no CONFORMANCE recapture); both wrapper gates now fail closed on any non-nil error (closes the fall-through fail-open). Three-lane audit clean at 0 C/H/M; SPEC-043-R008 governance declaration validated.

Issue #1397, opened 2026-09-06
F1: autotune --apply candidate-log filename overflows NAME_MAX on long usernames → onboarding blocked


Augustas11:
**Found in:** #1381 (BYOM v0.1 candidate E2E, @SmtTheSE) — finding **F1**, independently code-verified. **Not BYOM-specific:** this is the autotune onboarding path and affects **all** provider onboarding, including the 1.8.120 candidate under acceptance in #1396.

## Symptom

On a Mac whose username is long (macOS derives the account name from the Apple ID email — ~28 chars is ordinary), `macprovider-cli autotune --recommend --apply` fails with:

```
Error: The item couldn't be saved because the file name
"Users-<28-char-username>-Library-Application-Support-macprovider-models-mlx-community-
Llama-3.2-3B-Instruct-4bit-<40-char-sha>-<64-char-sha>-18080-1788598349-C801D332.log"
is invalid.
```

The candidate-serve **log filename is 258 bytes against the 255-byte `NAME_MAX`** — over by 3.

## Why it's fatal (not cosmetic)

`--apply` is the **only** source of `model_artifact_sha256`, and `serve` refuses to join a coordinator without it:

- `phase3-binary/Sources/macprovider-cli/MacProviderCLI.swift` — `"coordinator join requires model_artifact_sha256 from autotune --recommend --apply"` → `throw ExitCode(2)`.

So on such a Mac **provider onboarding cannot complete at all**, and the error names no cause and suggests no fix. The provider population is consumer Macs, so long Apple-ID-derived usernames are the norm, not the exception — a CI runner named `runner` (6 chars) clears the limit, which is why this was never seen in CI.

## Root cause

`phase3-binary/Sources/macprovider-cli/CandidateProviderRunner.swift`:

- `logFileURL(model:port:in:)` builds `"\(safeModelName(model))-\(port)-\(timestamp)-\(suffix).log"` and immediately does `try Data().write(to: logFileURL, options: .atomic)` — the throw point.
- `safeModelName(_:)` only maps non-`[A-Za-z0-9._-]` scalars to `-` and collapses runs. It performs **no hashing and no length bound**.

[...]

Issue #1396, opened 2026-09-06
Physical acceptance + resilience test of the 1.8.120 candidate (@SmtTheSE)


Augustas11:
**Owner:** @SmtTheSE · **Supersedes acceptance scope of** #1358 (that run validated the 1.8.119 candidate and did its job — see "Why a new issue"). **Recovery-net fixes under test:** #1369.

## Task

Physical (real-Mac) acceptance of the **signed 1.8.120** provider candidate before broadcast promotion. This is the resilience half of the release gate: the 1.8.120 candidate carries the installer/updater **recovery-net fixes** (#1369) that have been unit-tested + 3-lane-audited but have **never run on hardware**. Promotion is a fleet-wide auto-update, and the recovery net's job is to catch a *failed* auto-update — so it must be validated on a real Mac, not just in CI.

## Signed candidate

- **Run:** https://github.com/Augustas11/macprovider/actions/runs/34010040066 (both jobs green: build + protected signer)
- **Artifact:** `acceptance-candidate-b5c93f897a5dfc45aeb8523868f80009317a53a5` (55 MB, Developer-ID signed + notarized + stapled + bound: Malibu.app + macprovider-cli + coordinator/gateway)
- **Source commit:** `b5c93f897a5dfc45aeb8523868f80009317a53a5` (main), tag `v1.8.120`, `binaryVersion = 1.8.120`
- **Download:** `gh run download 34010040066 -n acceptance-candidate-b5c93f897a5dfc45aeb8523868f80009317a53a5` (or the Actions UI)

⚠️ **The artifact retention is ~24h (expires 2026-09-07).** If it has expired by the time you pick this up, ping and it is re-cut in ~23 min (re-dispatch `acceptance-candidate.yml` pinned to the current main tip).

## Why a new issue (not #1358)

#1358 validated the **1.8.119** candidate (`e4916367`, 2026-09-03). That run *passed the candidate itself* (clean 117→119 auto-update, honest truthfulness surfaces) but **surfaced the broken recovery net** — which is exactly what #1369 then fixed. #1369 (`9d616ea0`) merged **after** that test, so none of the fixes were in what Steven ran. The 1.8.120 candidate differs from the 1.8.119 one by **~13,000 lines in `phase3-binary`**, concentrated on the resilience surface:

- `dist/install.sh` +495 lines — `recovery_bootstrap_service` (F6: `launchctl bootstrap` I/O recovery)
[...]

Issue #1381, opened 2026-09-04
BYOM v0.1 candidate E2E (real-Mac): live Ollama loopback + coordinator admission + Malibu consumption


Augustas11:
Parent epic: #1240 · Runbook: `test/e2e/byom/CANDIDATE-E2E-RUNBOOK.md` (main `58c08ca4`)

## Task

Run the BYOM v0.1 candidate E2E on the full candidate stack on a real Mac — the last-mile validation unit tests and the 3-lane audits can't cover: **live Ollama loopback runtime + real coordinator admission + the Malibu app consuming the CLI-owned economics projection.** All 8 BYOM slices are merged to `main` (audit-clean, 0 C/H/M each); this run is what gates promotion.

The full step-by-step (commands, schema assertions, fail-closed checks, disablement proof, exit criteria) lives in the committed runbook — follow it directly:

`test/e2e/byom/CANDIDATE-E2E-RUNBOOK.md`

Summary of the four parts:
- **A. Provider CLI onboarding** — `models discover → evaluate → offer --dry-run → offer → admission status → withdraw` against live Ollama loopback + real coordinator; assert closed schemas + honest non-earning fields; non-catalog can't earn; endpoint/host material rejected.
- **B. Coordinator admission + routing gate** — non-`settlement_capable` states hidden from default paid routing and `/v1/pool/check`; catalog decisions require trusted catalog authority evidence; no ledger credit/debit/payout-ready row for any BYOM route (v0.1 enables no earning).
- **C. Malibu app consumption** — signed-projection-only economics; malformed/duplicate-key/unknown-field fails closed; `catalog_priced` rows show rates **with** the "No provider credit yet" disclosure even when switchable; verified-vs-provider-reported identity distinct; VoiceOver reads the full disclosure.
- **D. Disablement proof** — exercise `docs/runbooks/byom-disablement-rollback.md` switches; each disables its surface without losing provider state.

## Why this matters

BYOM v0.1 is fail-closed by construction (earning is impossible regardless of state), so the risk isn't payouts — it's the new provider-visible surfaces (CLI commands, coordinator admission endpoints, Malibu economics UI) behaving correctly on real runtimes. A real-Mac run catches what the hermetic `make test-byom-e2e` harness can't.

## Gates a clean run satisfies

- **#1245** — live non-MLX (Ollama loopback) adapter proof.
- **#1248** — current automated + real-Mac evidence toward promotion.

## Exit criteria / deliverables

[...]

PR #1399, opened 2026-09-06
fix(byom): operator guidance (F2/F5) + hermetic E2E harness (F8) (#1381)

Augustas11:
Addresses #1381 findings **F2**, **F5**, and **F8** (from @SmtTheSE's BYOM v0.1 candidate E2E).

## F2 — offer-refusal guidance was a dead end (message-only)

`models offer` refused an `evaluation_required` candidate with "run models evaluate and models offer --dry-run --json before submitting." But `canSubmit` passes only when the evaluation digest is non-empty **or** the `evaluation_required` warning is absent — and `models evaluate` persists nothing, so that warning never clears. Following the message loops forever. The guidance now names the input that works: run `models evaluate <ref> --json` and pass its `evaluation_digest_sha256` via `--evaluation-digest-sha256`.

## F5 — unstable-id dead end (message-only)

A first `discover` on a fresh namespace returns a `byom_unstable_*` candidate id. `models offer` then **provisions the salt and re-discovers a stable id**, so the unstable id an operator copied no longer matches (`selectCandidate` matches `candidateID || servedModelRef || displayName`). The `candidateNotFound` / `candidateUnstable` guidance now tells the operator to target by `served_model_ref`, and that re-running `discover` alone won't help (discovery is read-only, SPEC-046-R001).

## F8 — `make test-byom-e2e` couldn't pass on `main`

Two independent causes:

- **Cause 1** — the harness set `--local-discovery-namespace-path` but never created the salt file, so `discover` emitted `byom_unstable_*` and the `offerable` assertion was unreachable. Seed the salt before any command runs (Steven's patch).
- **Cause 2** — the catalog fixture `qwen3-8b` (~16 GB estimate) reports `does_not_fit` on CI runners (~14 GB) and ordinary dev Macs, blocking the offer/economics flow the harness actually tests. A fixture swap is not viable: `catalogKey(for:)` matches the served-ref by lowercase equality against catalog `key`/`model_id`, `qwen3-8b` is the smallest **short-key** model, and the 3B model's only key is the org-prefixed `meta-llama/llama-3.2-3b-instruct` — an Ollama-style served name won't match it, so a swap would null `catalog_model_key` and break the `catalog_priced` assertions. Instead, add a **BYOM-discovery-scoped, hermetic-E2E-only** RAM override `MACPROVIDER_BYOM_E2E_DETECTED_RAM_GB` that the harness sets; the fit **logic still runs** against the supplied RAM. It is scoped to BYOM's advisory local-readiness signal — it does **not** touch `ModelFit.detectRAMGB()` or autotune's global fit gating, and no money-path or coordinator decision consumes BYOM `fitState` (v0.1 enables no earning). Worst case in production: a provider misleads their own local fit display / OOMs their own machine.

## Tests

`swift test` (BYOMDiscoveryTests): guidance-message assertions (F2/F5) + RAM-override fallback semantics (positive int honored; 0/negative/non-numeric/empty → real detection). Build + these tests green locally. `make test-byom-e2e` green must be confirmed on the Ollama-loopback + coordinator rig (a #1396 exit criterion; not runnable in per-PR CI, per the harness README).

## Audit

3-lane audit (code / security / architecture) via Claude subagents (codex at usage limit this round). **All three PASS — 0 CRITICAL / HIGH / MEDIUM.**

- **Security** grep-confirmed **no coordinator/gateway consumer of `fit_state`** — settlement is gated on cryptographic catalog material, so the override's blast radius is self-inflicted-only (own local display / own OOM). Salt uses a CSPRNG at 0600.
- **Architecture** confirmed the env-var seam is *required* (a Python→CLI subprocess boundary can't be crossed by in-process DI/test-initializers; an undocumented test-named env is lower-surface than a permanent CLI flag), verified the override is scoped to the two BYOM `fitState` sites (the 5 other `detectRAMGB` callers untouched), and independently **refuted the fixture-swap alternative** against the baked catalog.
- **Code** verified F2/F5 guidance is factually accurate against `canSubmit`/`selectCandidate`/`submitOffer` and that the salt seed satisfies `readNamespace`'s exact checks.

Carried (LOW/INFO follow-ups, non-blocking): (a) the RAM override isn't build-gated — a future hardening could fence it behind `MACPROVIDER_BYOM_ALLOW_INSECURE_LOOPBACK_COORDINATOR` (architecture deemed this no-action for this batch given the self-inflicted blast radius); (b) `candidateUnstable` is also reached on `namespacePermissionInvalid`, where a permissions fix is additionally needed — the message steers to `served_model_ref` (the primary recovery) but doesn't name that case; (c) the env read sits in a domain helper rather than composition-root injection (refactor quality).

## Context

Batches into the enriched **1.8.120** acceptance (#1396) alongside F1 (#1397, merged). Not a promotion by itself.

## Governance

`behavior_change: none` in the buyer/settlement/spec sense — operator-facing guidance text, a test harness, and a BYOM-only advisory-fit test seam. No spec, contract, schema, or money-path surface changes.

### DIFF
--- a/phase3-binary/Sources/macprovider-cli/BYOMDiscovery.swift
+++ b/phase3-binary/Sources/macprovider-cli/BYOMDiscovery.swift
@@ -1211,11 +1211,11 @@ enum BYOMModelAdmissionError: Error, Equatable, CustomStringConvertible {
         case .missingAdmissionIdentity(let providerID):
             return "provider admission signing identity is missing for \(providerID); start provider enrollment before offering BYOM models"
         case .candidateNotFound:
-            return "BYOM candidate was not found in local discovery"
+            return "BYOM candidate was not found in local discovery; target the offer by its served_model_ref (for example ollama:llama3.2:3b) rather than a candidate id. models offer provisions the local namespace and re-resolves a stable id, so an id copied from a first discover on a fresh namespace (a byom_unstable_… value) will no longer match"
         case .candidateUnstable:
-            return "BYOM candidate id is unstable; run models discover --json to initialize the local namespace"
+            return "BYOM candidate id is unstable; target the offer by its served_model_ref (for example ollama:llama3.2:3b) instead of the byom_unstable_… id. models offer provisions the local discovery namespace and resolves a stable id; re-running models discover alone will not, because discovery is read-only (SPEC-046-R001)"
         case .candidateNotOfferable:
-            return "BYOM candidate is not offerable; run models evaluate and models offer --dry-run --json before submitting"
+            return "BYOM candidate is not offerable; run models evaluate <ref> --json and pass the returned evaluation_digest_sha256 to models offer via --evaluation-digest-sha256. models evaluate does not persist state, so the evaluation_required warning clears only when you supply that digest; also resolve any blocking readiness warning"
         case .invalidEvaluationDigest:
             return "evaluation digest must be a 64-character lowercase SHA-256 hex value"
         case .invalidWithdrawalReason:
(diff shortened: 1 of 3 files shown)


antfleet-ops (approved):
Approving BYOM batch (#1381 F2/F5/F8). 3-lane audit (code/security/architecture, Claude subagents; codex rate-limited) all PASS at 0 C/H/M — security grep-confirmed no coordinator/gateway consumes fit_state, architecture confirmed the env seam is required by the subprocess boundary and the fixture swap is genuinely blocked. Build + BYOMDiscoveryTests green; ci-required green. Only red is advisory spec-index (behavior_change: none). make test-byom-e2e validated on the Ollama rig (a #1396 exit criterion).

PR #1403, opened 2026-09-07
fix(se): omit claimed.model_id so SE attestation fits the 1024-byte cap

### LINKED ISSUE #1400: Tier-2 SE attestation silently disabled by hardcoded 1024-byte handshake cap (envelope 1028-1030B on M4)
Found in the 1.8.120 physical acceptance (#1396, @SmtTheSE); originally raised as F7 in the BYOM v0.1 E2E (#1381). Now with a precise root cause.

## Symptom

```
WARN se_attestation unsupported reason=token_exceeds_handshake_cap bytes=1028 cap=1024
```

Tier-2 Secure Enclave attestation is **silently disabled** — the code returns `nil` and continues, so the provider looks healthy while hardware-backed identity is unavailable.

## Root cause

`Tier2Attestation.swift:28` hardcodes `maxHandshakeTokenBytes = 1024`. On a MacBook Air M4 the attestation envelope measured **1028, 1029, and 1030 bytes across attempts** — a 4–6 byte overrun of a hardcoded cap.

Two things make this worse than a plain static limit:
1. **Intermittent** — the envelope size *varies between attempts on the same machine*, so it fails non-deterministically rather than always/never.
2. **Silent** — the only signal is a single `WARN` line in installer output; nothing surfaces that hardware attestation is off.

## Impact

On affected hardware, providers run with Tier-2 SE attestation unavailable without any clear signal. Not a 1.8.120 regression (pre-existing), and it does not endanger a working provider — but it undercuts the hardware-trust posture on current Apple silicon.

## Suggested direction

Right-size the handshake cap to the actual envelope (with margin) rather than a round 1024, and/or make the degrade explicit (a durable status, not just a WARN) so an operator/coordinator can see attestation is disabled. Confirm the envelope-size variability source (may be M4-specific).

Augustas11:
## Summary

Fixes #1400 without raising the SPEC-008 §7.4 1024-byte handshake cap.

On current catalog model IDs (~40 chars) the SE `attestation_token` JSON measured 1028–1030 bytes. The provider omitted the token (`WARN token_exceeds_handshake_cap`) and continued, so the session looked healthy while SE key-custody (`self_signed`) never ran.

The overflow is **not M4-specific** and is **not intermittent pass/fail** — DER signature length only changes the WARN byte count. The unbounded term was `claimed.model_id`, which Pillar C never consumes on the SE path (`model_hash` is omitted; hello already carries `model_id`). Dropping that field lands the envelope at ~975–980 bytes.

**Do not bump the cap.** It is coordinator-enforced (`maxHandshakeMetadataBytes`); a Swift-only bump would convert today's graceful omit into WS close 4001.

## Test plan

- [x] `swift test --filter SEAttestation` — 18 tests, 0 failures
- [x] Realistic envelope with a 2000-char `model_id` still fits and keeps 32 bytes of headroom
- [x] Omit-on-overflow still covered via oversized `binary_version`
- [ ] Physical: 1.8.120+ provider on the #1396 M4 Air should no longer print `token_exceeds_handshake_cap` and should present an SE token

## Audit

Native three-lane review (code / security / architect) — **0 CRITICAL / 0 HIGH / 0 MEDIUM**. `omc ask codex` hit the Codex usage limit until 10:34 AM; lanes ran as native Cursor subagents over the same diff.

Carried LOW/INFO (not blocking):
- Remaining headroom is ~44–49 bytes on canonical `mp-` IDs; a 64-char `provider_id` plus max-length semver can still hit the omit guard (defense-in-depth kept).
- SPEC-008 §7.4a's SE `claimed` key set is not enumerated (example at §7.4 is the MDA format and still accurate). Follow-up, not this PR.
- MDA generator still has no 1024-byte self-check (pre-existing; spec says inline MDA generally cannot fit).
- Coordinator-side omit vs "no SE hardware" is still indistinguishable (`not_required`). Follow-up observability, not this PR.

This restores spec-intended SE presentation. It does **not** mint `hardware` attestation — SE remains `self_signed` key custody.

{
  "behavior_change": "yes",
  "contract_change": "none",
  "issue": "https://github.com/Augustas11/macprovider/issues/1400",
  "specs": ["SPEC-008"],
  "requirements": ["SPEC-008-R001"],
  "authority_domains": ["tier2-trust-evidence"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "swift test --filter SEAttestation (18 tests, 0 failures)",
    "SEAttestationGeneratorEnvelopeTests.testRealisticEnvelopeFitsHandshakeCap (2000-char model_id, 32-byte headroom)",
    "SEAttestationGeneratorEnvelopeTests.testOversizedBinaryVersionOmitsTokenRatherThanExceedCap"
  ],
  "journeys": ["not-required"]
}


antfleet-ops (approved):
Approved. Real overflow vs the SPEC-008 1024-byte handshake cap; dropping unused claimed.model_id is the correct fix (a cap bump would close 4001). 0 C/H/M across code/security/architect. Merge when required checks are green.

antfleet-ops (approved):
Re-approve after merging origin/main (SPEC-010-R002 + SPEC-016-R002 spec-index repairs). Original SE envelope fix unchanged.

PR #1410, opened 2026-09-07
fix(release): bundle MLX Metal resources into acceptance-candidate Malibu.app

Augustas11:
## What

The acceptance-candidate signer (`scripts/sign-acceptance-candidate.sh`) assembled `Malibu.app` with only `macprovider-cli` + catalog payloads in `Contents/MacOS`, never the adjacent `mlx.metallib` and SwiftPM resource bundles (`mlx-swift_Cmlx.bundle`, `swift-nio_NIOPosix.bundle`) that the standalone tarball payload already ships. The app-bundled CLI loads MLX Metal kernels from files adjacent to it, so an **app-installed** provider on a build cut through this flow could not run Metal inference.

- **v1.8.117** was cut via `release.yml` (which does copy these, guarded by `test-release-security-posture.sh`) → unaffected.
- **v1.8.120** was the first cut through `acceptance-candidate → promote` → shipped a metallib-less `Malibu.app`. Caught by the post-publication `verify-tier2-provider-release` gate **before any fleet rollout** (coordinator still recommends 1.8.117).

## Fix

- Copy `mlx.metallib` + the two SwiftPM bundles from `$cli_work` into `Malibu.app/Contents/MacOS`, then **codesign the metallib and every nested `*.bundle`** with the Developer ID identity **before** the non-deep outer app sign (mirroring `release.yml:954-963`). This is required: the shipped v1.8.117 nested bundles each carry their own `Contents/_CodeSignature`, and the non-deep outer sign does not add them — without explicit nested signing, `codesign --verify --strict --deep` and Apple notarization would reject the app.
- Fail closed after stapling if `mlx.metallib` is absent.
- Add a source guard in `test-acceptance-candidate-security.sh` so the acceptance-candidate app-assembly path is covered (previously only `release.yml` was guarded — the gap that let this ship).

## Verification

- `scripts/test-acceptance-candidate-security.sh` (new guard) → PASS; proven to FAIL when the sign-script copy is reverted.
- `scripts/test-tier2-provider-release.sh`, `test-release-security-posture.sh`, `test-compatibility-set-manifest.sh`, `test-malibu-independent-release.sh` → PASS.
- 3-lane audit (code / security / architecture) to 0 C/H/M.

Release/packaging path; no runtime code or schema change. A new version will be cut through the full flow to supersede the broken v1.8.120 DMG (immutable public release — not patched in place).

### DIFF
--- a/scripts/sign-acceptance-candidate.sh
+++ b/scripts/sign-acceptance-candidate.sh
@@ -223,6 +223,41 @@ cmp "$cli_work/compatibility-set.json" "$app/Contents/Resources/compatibility-se
 rm -rf "$app/Contents/Resources/compatibility-set-local" "$app/Contents/Resources/catalog-release"
 cp -R "$cli_work/compatibility-set-local" "$app/Contents/Resources/compatibility-set-local"
 cp -R "$cli_work/catalog-release" "$app/Contents/Resources/catalog-release"
+# Malibu.app's bundled macprovider-cli loads MLX Metal kernels and SwiftPM
+# resources from files adjacent to it in Contents/MacOS, exactly like the
+# standalone tarball payload assembled below. Copy them in before the app is
+# signed/notarized/stapled so they are sealed into the shipped bundle; without
+# them the app-bundled CLI cannot run MLX Metal inference.
+[[ -f "$cli_work/mlx.metallib" && ! -L "$cli_work/mlx.metallib" ]] ||
+  die "provider payload lacks mlx.metallib for the Malibu.app bundle"
+install -m 0644 "$cli_work/mlx.metallib" "$app/Contents/MacOS/mlx.metallib"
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Approve: acceptance-candidate Malibu.app now bundles AND signs mlx.metallib + nested SwiftPM bundles before the outer sign (mirrors release.yml), fixing the metallib-less DMG that shipped as v1.8.120. 3-lane audit (code/security/architecture) round 2 all 0 C/H/M; round-1 HIGH (missing nested signing) resolved and re-verified against v1.8.117's independently-signed bundles. Regression guard added. Carried LOW: align app-copy to the *.bundle glob release.yml uses (follow-up, no unsigned-payload gap today since sign loop globs + --deep self-gate).

PR #1094, opened 2026-08-20
Allow Pearl runtime-only healthz ahead of a staged CLI release

Augustas11:
## Summary

v1.8.104 sign/publish notarized, then failed closed: live Pearl `/healthz` is `v1.8.103` while coordinator still advertises `1.8.102`. The staged pre-publication gate required those two versions to be equal.

That equality conflates Pearl runtime with the advertised CLI. `v1.8.103` is a Pearl-only tag; we are not deploying the 1.8.104 coordinator (poolmanifest/rewards) just to unstick this CLI. Recommendation stays `1.8.102`.

- Allow `previous advertised CLI <= /healthz < candidate` at staged pre-publication.
- Keep `/healthz >= candidate` allowed (Pearl-first deploy, already tested).
- Reject `/healthz` older than the advertised CLI.
- Keep `recommended_binary_version` locked to the previous advertised CLI until post-publication.

The existing unpublished `v1.8.104` draft/tag still points at `a6a999b2` and cannot use this script. After merge: delete the unpublished draft, retag signed `v1.8.104` onto the new main, dispatch `release.yml` again. Do not `PATCH draft=false` on the current draft.

Live proof of the new gate against coordinator.malibu.tech (feeds + healthz `v1.8.103` + recommended `1.8.102`): pass.

Codex lanes: code 0/0/0, architect 0/0/0, security r2 0/0/0 (r1 MEDIUM was the already-allowed Pearl-first path).

SPEC-GOVERNANCE-DECLARATION-BEGIN
{
  "schema_version": "spec-pr-governance-v1",
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-020"],
  "requirements": ["SPEC-020-R001"],
  "authority_domains": ["provider-autoupdate"],
  "arbitration": ["CODE_BUG"],
  "tests": ["scripts/test-live-coordinator-release-gate.sh"],
  "journeys": ["not-required"],
  "issue": "https://github.com/Augustas11/macprovider/issues/614"
}
SPEC-GOVERNANCE-DECLARATION-END

Made with [Cursor](https://cursor.com)

PR #1419, opened 2026-09-07
fix(release): decouple coordinator binary version from CLI release in rollout gate

Augustas11:
## What

`verify-live-coordinator-release-gate.py` failed the **post-publication** rollout gate whenever the live coordinator's own `/healthz` binary version was older than the provider-CLI release being rolled out — e.g. coordinator **v1.8.119** rolling out CLI **1.8.121** → `/healthz version 'v1.8.119' is older than release 1.8.121`.

## Why it's wrong

The Pearl-runtime **coordinator** and the **provider CLI** are independent artifacts that share the `vX.Y.Z` tag space but ship on **independent cadences**. The coordinator does not need to be rebuilt in lockstep with every CLI release (especially a packaging-only one).

- SPEC-034: *"independent Malibu/CLI marketing versions negotiate by capability, not equality."*
- SPEC-020: `recommended_binary_version` is operator-set and advertised regardless of the coordinator's own binary version.
- The check already allowed the coordinator to be **ahead** of a staged CLI (#1094) — it simply never handled the **behind** case post-publication.

## Fix

Gate the `/healthz`-version-older-than-release failure to **pre-publication only** (where it still guards against a coordinator regressed below the previously-advertised stable). Post-publication correctness is enforced by the existing **`recommended_binary_version == release`** check, which is authoritative.

Test updated: a coordinator older than the CLI release is now accepted post-publication; added a pre-publication regression that still rejects a coordinator older than the previous stable.

## Verification
- `scripts/test-live-coordinator-release-gate.sh` → PASS
- `scripts/test-acceptance-promotion.sh` → PASS
- 3-lane audit (code / security / architecture) to 0 C/H/M.

### DIFF
--- a/scripts/test-live-coordinator-release-gate.sh
+++ b/scripts/test-live-coordinator-release-gate.sh
@@ -220,11 +220,23 @@ if run_guard "$work/missing-sig" >"$work/missing-sig.out" 2>&1; then
 fi
 grep -q 'fixture coordinator response is missing for /v1/rate-card.sig' "$work/missing-sig.out"
 
-make_fixture "$work/stale-healthz" v1.8.68 v1.8.67
-if run_guard "$work/stale-healthz" >"$work/stale-healthz.out" 2>&1; then
-  fail "accepted a coordinator older than the shipped CLI release"
+# Post-publication: the coordinator's own /healthz binary version is an
+# independent Pearl-runtime artifact that shares the vX.Y.Z tag space with the
+# provider CLI but ships on its own cadence. A coordinator OLDER than the shipped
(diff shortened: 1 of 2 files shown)


antfleet-ops (approved):
Approve: decouple coordinator binary version from CLI release in the post-publication rollout gate. 3-lane audit (code/security/architecture) all 0 C/H/M; security confirmed no crypto/signature/digest weakening and no downgrade surface (recommended==release + feed digests + Ed25519 all preserved); architect confirms SPEC-034/020/042-R170 support the decoupling and pre-publication rollback guard is retained. Unblocks the v1.8.121 signed discovery-head rollout.

PR #1423, opened 2026-09-07
fix(byom): treat evaluation_required as advisory; unify submit/dry-run blockers (SPEC-047-R002)

Augustas11:
## Summary

The provider CLI had two predicates deciding whether a BYOM offer can be submitted, and they disagreed on `evaluation_required`:

- `canSubmitLocalDryRun` (dry-run predictor) treated it as **advisory** → `would_submit: true`.
- `canSubmit` (the real offer-submit gate) treated it as a **hard block** unless an evaluation digest was supplied.

PR #1399 changed only the dry-run side, so the dry-run **over-promised**: it told the provider `would_submit: true` for a candidate the real `models offer` (without a digest) would refuse with `candidateNotOfferable`. This is a provider-facing truthfulness gap and it was the cause of the red `make test-byom-e2e` on `main`.

**SPEC-047-R002 is authoritative:** *"Offer submission MAY omit the evaluation digest only when the dry-run response explicitly labels the offer as more likely to be rejected or confined to non-earning states."* Every v0.1 `local_inventory_only` candidate is non-earning, so the digest is optional — `canSubmit` was the surface out of conformance.

## What changed

- **Single source of truth** `BYOMDiscoveryWarning.submitBlockingWarnings` (ordered) / `.submitBlockingWarningCodes` (Set). `canSubmit`, `canSubmitLocalDryRun`, and `reasonCode(for:)` all derive from it, so the dry-run can never again over- or under-promise the real submit. `evaluation_required` is deliberately absent.
- The shared set **adds `adapter_unavailable` + `adapter_timeout`**, which `canSubmit` previously failed to block and on which the dry-run diverged — both now fail closed. (Found by the code-review audit lane.)
- `canSubmit` drops its now-unused `evaluationDigestSHA256` parameter.
- `candidateNotOfferable` guidance leads with the real hard blockers; the digest is framed as an advisory strengthening step (still names `--evaluation-digest-sha256`).
- Tests: allow a missing digest for a non-earning candidate; table-test every hard blocker (incl. the two new adapter codes) failing closed **with and without** a digest; the E2E harness asserts `would_submit: true` + honest `reason_code: catalog_binding_unverified` (SPEC-047-R003) with `evaluation_required` still surfaced in `warnings`.
- **CI:** wire the previously manual-only `make test-byom-e2e` into the `phase3-binary` swift-tests job, and extend the change detector (`ci-detect-changed-paths.sh`) + its parity test to gate on `test/e2e/byom/*`, so this harness cannot silently regress again. (It was manual-only, which is why the drift went unnoticed for a day.)

## Testing
- Swift: `BYOMAdmissionTests` 21/0, `BYOMDiscoveryTests` 25/0, `BYOMOfferDryRunTests` 8/0
- `make test-byom-e2e`: **PASS** (was RED on `main`)
- Detector parity suite: all scenarios pass, incl. the new BYOM harness case

## Audit
Three lanes (code / security / architecture) via codex over the full fix diff: **0 CRITICAL, 0 HIGH, 0 MEDIUM.** Carried LOW (documented): a future signed promotion journey (SPEC-047-R008) should include one non-earning offer where the dry-run reports `would_submit: true`, `evaluation_required` stays in warnings, and the signed offer omits the digest — coverable by existing journey categories, no new state required.

Refs #1381

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -424,6 +424,14 @@ jobs:
         working-directory: phase3-binary
         run: swift test --parallel
 
+      - name: Run BYOM CLI onboarding E2E (hermetic)
+        # #1381 F8: this hermetic harness (fake Ollama loopback + in-process
+        # coordinator) is the only automated cover for the CLI-owned BYOM
+        # admission surface end to end. It was manual-only and silently
+        # regressed; gate it here so canSubmit/dry-run drift fails a required
+        # check. Reuses the SwiftPM build warmed by the swift test step above.
+        run: make test-byom-e2e
+
       - name: Install reviewed XcodeGen artifact
         shell: bash
         run: |
--- a/phase3-binary/Sources/macprovider-cli/BYOMDiscovery.swift
+++ b/phase3-binary/Sources/macprovider-cli/BYOMDiscovery.swift
@@ -18,6 +18,29 @@ enum BYOMDiscoveryWarning: String, Codable, Sendable {
     case namespacePermissionInvalid = "namespace_permission_invalid"
 }
 
+extension BYOMDiscoveryWarning {
+    /// Hard blockers for a BYOM offer submission, highest-priority first.
+    ///
+    /// Single source of truth so the real submit gate
+    /// (`BYOMOfferSubmissionBuilder.canSubmit`), the dry-run predictor
+    /// (`canSubmitLocalDryRun`), and the dry-run `reason_code` selector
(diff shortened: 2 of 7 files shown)


antfleet-ops (approved):
Approving: SPEC-047-R002 conformance fix. canSubmit/dry-run/reasonCode now share one blocking-warning source of truth; evaluation_required is advisory for non-earning v0.1 offers; adapter_unavailable/adapter_timeout now fail closed. 3-lane audit 0 C/H/M, CI green including the newly-gated make test-byom-e2e.

PR #1425, opened 2026-09-07
test(byom): VoiceOver reads catalog_priced rates with the non-earning caveat (#1381 Part C)

Augustas11:
## Summary

Closes the app-rendering + VoiceOver half of **#1381 Part C**. The tester's physical acceptance left the Malibu app's `catalog_priced` economics presentation and its VoiceOver read-through unverified — his self-hosted SQLite coordinator could not produce a catalog-bound candidate. But that half is **fixture-testable and needs no trusted-catalog coordinator**, so this pins it deterministically.

SPEC-047-R004 requires provider-facing surfaces to state that a non-settlement model is not earning-eligible. The app shows a *"No provider credit yet; catalog and receipt checks are still required."* caveat beside `catalog_priced` rates — but nothing guaranteed a screen-reader user hears the caveat **together with** the rates.

## What changed

- **`MalibuModelRow.economicsRateLines`** — single source of truth for the provider rate lines. Both the visible `Text` rows and the composed accessibility label derive from it, so the shown copy and the VoiceOver announcement can't drift apart.
- **`MalibuModelRow.economicsAccessibilityLabel`** — composes the rate lines + the non-earning caveat into **one** announcement, so a `catalog_priced` (non-settlement) row's rates can never be voiced without the caveat. `nil` when there is no economics content.
- **`ModelRowView`** — renders the visible rates via `ForEach(economicsRateLines)` and applies the composed label to the grouped economics element (`.accessibilityElement(children: .ignore)`) with a stable `accessibilityIdentifier("byom.economics.summary")` hook.
- **Tests** — `catalog_priced` non-settlement voices rates **and** the full disclosure (announcement is exactly the rate lines in order, caveat last; every visible rate line appears verbatim); `settlement_capable` voices rates **without** the caveat. Locale-independent (the rate decimal separator is locale-derived).

## Testing
Full `MalibuTests/ModelManagementTests` via xcodebuild: **96 tests, 0 failures**.

## Audit
Three lanes (code / security / architecture) via codex over the full diff: **0 CRITICAL, 0 HIGH, 0 MEDIUM.** A first pass flagged a visible-vs-a11y rate-string duplication (LOW) — resolved by the shared `economicsRateLines` source; re-audit clean.

## Scope note
This closes the **app-rendering / accessibility** half of Part C. A live **trusted-catalog rig** (signed catalog authority + catalog-bound candidate + hardware-verification trust row) is still what a true end-to-end acceptance would use to prove the coordinator actually promotes a real candidate to `catalog_priced` on hardware — that remains a separate, heavier follow-up and is not required to validate this composition.

Refs #1381


antfleet-ops (approved):
Approving: #1381 Part C app-rendering/accessibility. catalog_priced rates and the non-earning caveat are composed into one atomic VoiceOver announcement (shared economicsRateLines source); settlement_capable omits the caveat. 96/0 MalibuTests, 3-lane audit 0 C/H/M, CI green.

PR #1443, opened 2026-09-08
fix(install): stop false rollback_failed on fresh first install (#1421)

### LINKED ISSUE #1421: install.sh reports rollback_failed on a fresh first install; recovery remediation not always runnable
**Surfaced by:** @SmtTheSE during 1.8.120 physical acceptance ([#1396 acceptance report](https://github.com/Augustas11/macprovider/issues/1396#issuecomment-5562764083), finding 2). Non-blocking for the 1.8.121 release — split out here so it survives the #1396 close.

## Problem

Every failed install that reached service start reported an automatic-rollback **failure**, even on a **first** install where there is nothing to roll back to:

```
[macprovider-recovery] provider service is active even though it was inactive before the failed install
[macprovider-install] ERROR: automatic rollback failed; recovery data was preserved at ...
[macprovider-install] Run exactly: bash '.../install-recovery-<ts>/recover.sh'
```

On a fresh machine "inactive before the install" is simply the truth, but the code takes the restore path and reports failure. Worse, the printed remediation is **not always runnable**: the recovery directory lives under `~/.config/macprovider/`, which does not reliably survive, so the printed `recover.sh` command returns *No such file or directory*.

Also emitted on **every** attempt, including successful ones:

```
WARNING: could not persist rollback lifecycle state before restoring the previous install
```

Nothing broke as a result — but a rollback that cannot persist its lifecycle state is the precondition for the exact wedge class the recovery net (`recover-update`, #1369) exists to recover from.

## Verified still present on `main` (e81776cc)

- `phase3-binary/dist/install.sh:6421`, `:6537` — `provider service is active even though it was inactive before the failed install`
- `phase3-binary/dist/install.sh:7447` — `WARNING: could not persist rollback lifecycle state before restoring the previous install`
- `phase3-binary/dist/install.sh:7459` — `ERROR: automatic rollback failed; recovery data was preserved at ...`

## Acceptance

- A failed **first** install (no prior installation) does not report `automatic rollback failed`; it distinguishes "nothing to restore, just clean up staged files" from "restore of a previous install failed."
- The printed `recover.sh` remediation is only emitted when the recovery directory it names actually exists and is runnable; otherwise a remedy that works is printed.
- The `could not persist rollback lifecycle state` warning is understood/fixed — either it stops firing on the happy path, or (if it is load-bearing) the lifecycle state is persisted durably so a rollback interrupted here cannot wedge.

Augustas11:
Fixes #1421.

## What was reported
During 1.8.120 physical acceptance (#1396 finding 2), every failed install that
**reached service start** reported an automatic-rollback FAILURE even when the
rollback was clean, and the printed `recover.sh` remediation was often already
gone by the time an operator ran it. A benign `could not persist rollback
lifecycle state` warning was also observed.

## Root causes (verified against `main`, not the 1.8.117-era baseline the report was taken on)
1. **Bootout race.** `recover.sh`'s `stop_loaded_service` booted out a
   just-created service and the caller asserted `service_loaded` with **no settle
   wait**. launchd keeps reporting a booted-out service as loaded for a short
   window, so a failed *first* install (nothing was loaded before it) tripped
   `provider service is active even though it was inactive before the failed
   install` → `recover.sh` exit 70 → `automatic rollback failed` on a clean
   rollback. The installer's own repair-preflight path already polls up to 2s
   after bootout; `recover.sh` did not.
2. **Stale remediation.** On `recover.sh` failure, `rollback_install_transaction`
   printed `Run exactly: bash <dir>/recover.sh` and returned **without** disarming
   the interrupted-install recovery agent. That agent (`observe.sh`) takes the
   same install lock, re-runs `recover.sh` after the installer exits, and on
   success `rmtree`s the recovery dir — so the printed path went stale.
3. **Fresh-install wording.** A failed first install (`EXISTING_INSTALL_WAS_PRESENT=0`)
   used upgrade-shaped wording ("restoring the previous provider installation").
4. **Lifecycle guard.** The state machine only allowed `uninstalled → installing
   (installer)`. A reinstall-after-uninstall whose cutover fails *before* the
   `installing` checkpoint then tried `uninstalled → rollback_in_progress
   (installer)`, which was rejected — so the installer's rollback silently failed
   to persist its intent (the observed warning). NB: the report's "on every
   attempt, including successful ones" is not reproducible on `main` — the string
   lives only in the rollback path.

## The fix
- `stop_loaded_service`: bounded post-bootout settle loop (20×0.1s) mirroring the
  repair-preflight quiesce loop; fails closed if the service never unloads.
- `rollback_install_transaction`: branch wording on `EXISTING_INSTALL_WAS_PRESENT`;
  on `recover.sh` failure keep the recovery agent armed and print an accurate
  remedy (armed auto-retry; `recover-update` for upgrades / re-run installer for
  fresh; `recover.sh` path while it exists).
- `ProviderLifecycleState.swift`: extend the fail-closed guard to allow
  **installer-only** `uninstalled → {installing, rollback_in_progress}`. Every
  other writer and target state stays fail-closed.

## Tests
- New `install_recovery_bootout_settle.test.sh` — race settles+succeeds,
  never-unloads fails closed, already-unloaded fast path, foreign-identity
  no-bootout, 5-arg legacy shape. Proven to **fail** against the pre-fix helper.
  Wired into `make test-dist`.
- `ProviderLifecycleStateTests` — installer rollback-from-uninstalled allowed;
  non-installer / foreign-target stays fail-closed.
- Updated `install_upgrade_evidence_rollback.test.sh` for the new remediation.
- Full local suites green: settle test, upgrade rollback fault matrix,
  ProviderLifecycleStateTests (19), ProviderLifecycleShellConformanceTests (17),
  UninstallCommandTests (25), `swift build`.

## Audit (full combined diff, 3 lanes)
Code 0/0/0 · Security 0/0/0 · Architecture 0/0/0. Security lane explicitly
confirmed the guard change keeps every non-installer writer fenced and adds no
stale-writer overwrite path.

Carried LOWs (non-blocking):
- Remediation prints `macprovider-cli recover-update` unqualified (consistent
  with existing installer guidance; no injection/escalation — same operator,
  fixed string). Defense-in-depth absolute-path printing is a separate pattern.
- The fresh-install rollback wording branch is exercised via the shared code
  paths but not a dedicated fault-matrix case (the matrix harness is
  upgrade-shaped); the branch is a simple message selection.

{
  "issue": "https://github.com/Augustas11/macprovider/issues/1421",
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-001", "SPEC-020"],
  "requirements": ["SPEC-020-R004"],
  "authority_domains": ["provider-autoupdate", "native-app-lifecycle"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "phase3-binary/dist/test/install_recovery_bootout_settle.test.sh",
    "phase3-binary/dist/test/install_upgrade_evidence_rollback.test.sh",
    "phase3-binary/Tests/macprovider-cliTests/ProviderLifecycleStateTests.swift::testInstallerRecordsRollbackFromUninstalledForFailedReinstall",
    "phase3-binary/Tests/macprovider-cliTests/ProviderLifecycleStateTests.swift::testUninstalledStaysFailClosedForNonInstallerAndForeignTargets"
  ],
  "journeys": ["not-required"]
}

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,6 +183,7 @@ test-dist:
 	bash phase3-binary/dist/test/install_fresh_evidence.test.sh
 	bash phase3-binary/dist/test/install_recommend_retry.test.sh
 	bash phase3-binary/dist/test/install_upgrade_evidence_rollback.test.sh
+	bash phase3-binary/dist/test/install_recovery_bootout_settle.test.sh
 	bash phase3-binary/dist/test/install_launchd_migration.test.sh
 	bash phase3-binary/dist/test/install_lifecycle_state.test.sh
 	bash phase3-binary/dist/test/install_transaction_lock.test.sh
--- a/phase3-binary/Sources/macprovider-cli/ProviderLifecycleState.swift
+++ b/phase3-binary/Sources/macprovider-cli/ProviderLifecycleState.swift
@@ -261,7 +261,18 @@ struct ProviderLifecycleStateRecord: Codable, Equatable, Sendable {
 
         guard let previous else { return }
         if previous.state == .uninstalled {
-            guard writer == .installer, state == .installing else {
+            // A prior uninstall left an `uninstalled` tombstone. The installer is
+            // the only writer allowed to drive lifecycle out of that state:
+            // normally it records `installing` to begin a fresh install, but a
+            // reinstall whose cutover fails before it reaches the `installing`
+            // checkpoint must still be able to persist the `rollback_in_progress`
+            // intermediate state. Forbidding it made the installer's own rollback
+            // silently fail to record its intent on exactly a reinstall-after-
+            // uninstall -- the precondition for the wedge class recovery exists to
+            // clear (#1421). Every other writer, and every other target state,
+            // stays fail-closed.
+            guard writer == .installer,
+                  state == .installing || state == .rollbackInProgress else {
                 throw ProviderLifecycleStateError.invalidTransition(from: previous.state.rawValue, to: state.rawValue)
             }
         }
--- a/phase3-binary/Tests/macprovider-cliTests/ProviderLifecycleStateTests.swift
+++ b/phase3-binary/Tests/macprovider-cliTests/ProviderLifecycleStateTests.swift
@@ -166,6 +166,67 @@ final class ProviderLifecycleStateTests: XCTestCase {
         XCTAssertNotEqual(resumed.previousTransitionID, restarting.transitionID)
     }
 
+    func testInstallerRecordsRollbackFromUninstalledForFailedReinstall() throws {
+        // #1421: a reinstall after an uninstall tombstone whose cutover fails
+        // before the `installing` checkpoint must still be able to persist the
+        // `rollback_in_progress` intermediate state, so the installer's rollback
+        // records its intent instead of silently failing (the wedge precondition).
+        let fixture = try Fixture()
+        let store = ProviderLifecycleStateStore(url: fixture.recordURL)
+        _ = try store.transition(
+            to: .uninstalled,
+            reasonCode: "uninstall_services_stopped",
+            writer: .installer,
+            operationID: "uninstall-1"
+        )
+        let rolledBack = try store.transition(
+            to: .rollbackInProgress,
+            reasonCode: "install_admission_failed",
+            writer: .installer,
+            operationID: "reinstall-1"
+        )
+        XCTAssertEqual(rolledBack.state, .rollbackInProgress)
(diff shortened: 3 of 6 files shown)


antfleet-ops (approved):
Approving: bootout-race root cause is correct and the fix mirrors the existing repair-preflight settle loop; the lifecycle guard relaxation stays installer-only and fail-closed for every other writer (confirmed by the new tests and the 3-lane audit: code/security/architecture all 0 CRITICAL/HIGH/MEDIUM). New regression test is wired into make test-dist.

PR #1465, opened 2026-09-09
feat(release): BYOM v0.2 slice 2b-ii — artifact feed as a release asset, Pearl deploy envelope, post-activation renewal (#1453)

Augustas11:
## Summary

BYOM v0.2 **slice 2b-ii** (epic #1453): the SPEC-023 artifact feed becomes a **release asset**, is staged into the Pearl release envelope by the coordinator deploy, and the scheduled feed renewal keeps working after activation. Completes the distribution surfaces behind slices 2a (generator, #1461) and 2b (coordinator serving + live gate, #1463); `catalog-release.py status` now reports no pending distribution surface. **No release is cut and no catalog bytes change**: every path is conditional on `release.json` binding the feed, and today's rate-card-bound release produces byte-identical outputs.

### The Stage A shape (locked by the first audit round)
A release is **artifact-bound iff its `release.json` `feeds` contains `autotune-artifacts.json`** — the single binding authority everywhere. Then, and only then, `autotune-artifacts.json` + `.sig` are **GitHub release assets**, never provider-payload members and never `compatibility-artifact-index` roles: the deployed updater and installer enforce the exact nine-name `catalog-release/` set and the exact seventeen index roles, so either widening would fail-close every installed provider before the new binary runs (SPEC-023 §3.7.8 Stage A). The CLI's fallback is the snapshot compiled into the binary (§3.7.2), which slice 2c bakes.

### What lands
- **Acceptance path.** `acceptance-candidate.yml` captures the pair from the candidate checkout as two more unsigned inputs; `verify-unsigned` makes the build boundary exact in both directions (an artifact-bound provider archive must be accompanied by the pair, a rate-card-bound one must not be) and checks the feed's digest and length against the archived `release.json` record; `sign-acceptance-candidate.sh` authenticates the pair with `catalog-release.py verify-directory` over a separate nine-payload-files-plus-pair catalog directory (digest, sidecar signature, signer equality with the candidate feed) and appends it to `release_assets` (checksums, provenance, `release-assets.txt`).
- **Release workflow.** The direct publish path copies the pair from `phase3-binary/dist/static/` after checking it against `release.json`; the promoted-restore, provider-runtime, and signing reconstructions re-derive the binding from the provider archive itself; `pearl-release.json` `catalog.files` binds the pair.
[...]

PR #1463, opened 2026-09-09
feat(coordinator): BYOM v0.2 slice 2b — serve the SPEC-023 artifact feed at /v1/catalog-artifacts + live release gate (#1453)

Augustas11:
## Summary

BYOM v0.2 **slice 2b** (epic #1453), on top of the merged slice 2a (#1461, `fdc6c83f`): the coordinator **serves** the SPEC-023 §3.7 artifact feed and the live release gate **verifies** it. No release is cut and no catalog bytes change; the checked-in `dist/coordinator.yaml` deliberately does not name the new config pair, so production behaviour is unchanged until the activation deploy. Release-asset publication (`package.sh`, `release.yml`, `pearl-release.json` producer, deploy upload) is slice 2b-ii; CLI consumption with baked fallback is 2c.

### Coordinator serving (`internal/buyer`, `internal/config`, nginx)
- Optional config pair `autotune.catalog_artifacts_path` / `autotune.catalog_artifacts_sig_path`. Unset: `/v1/catalog-artifacts` and `.sig` answer 404 and `/v1/autotune-release` keeps the exact v0.1 shape (§3.7.6 rule 6). Set: requires the three base feeds; validation both-or-neither.
- Loaded with the base feeds at boot and on SIGHUP reload (`LoadAutotuneFeeds`). The coordinator refuses bytes it cannot bind to the candidate catalog it serves beside them: signer `key_id` equality with the candidate sidecar (§3.7.2 — a valid signature by a second concurrently trusted key fails), `version` / `generated_at` / `policy_version` equality, `candidate_catalog_sha256` equal to the served candidate bytes, `release_id == version` (§3.7.4), the closed identity matrix + global hash uniqueness + `primary_artifact_id` rules, §3.7.5 primary-artifact consistency with the candidate rows. Closed schema is presence-aware (a present `null` on an optional string field, or an omitted `verified_at`, is rejected exactly as the generator rejects it). A mismatch fails startup; on reload it keeps the prior served set.
- Literal bytes through the shared feed handler (same rate limit and cache headers). nginx: exact `location = /v1/catalog-artifacts` and `.sig` allow-through blocks before the `/v1/` catch-all, same allowlisted shape as `/v1/rate-card`; asserted by `check_nginx_catalog_routes_test.sh` and the Go ordering test.
[...]

PR #1461, opened 2026-09-09
feat(catalog): BYOM v0.2 slice 2a — artifact feed generator, class rate rows, ledger v3 (#1453)

Augustas11:
## Summary

BYOM v0.2 **slice 2a** (epic #1453): the release-generator half of the SPEC-023 v0.10.0 catalog pipeline — artifact feed generation, class rate expansion, and ledger v3. **No release is cut, no catalog bytes change**: the committed 4-feed release still verifies unchanged; artifact-feed generation is proven hermetically in a scratch release signed with a throwaway key. Coordinator route (2b) and CLI consumption (2c) follow as separate PRs.

### Artifact feed (SPEC-023-R004, §3.7)
- Operator source `phase3-binary/catalog/autotune/autotune-artifacts-source.json` (`macprovider.autotune-artifacts-source.v1`; its `models` block is the published feed's `models` block verbatim) → `catalog-release.py generate` emits `autotune-artifacts.json` in the exact §3.7.3 closed shape with release binding filled in.
- Enforced at generation and verify: closed schema at every level; format × algorithm × source × runtime tuple matrix; global `(hash_algorithm, hash)` uniqueness across model keys; `artifact_id` grammar; cross-release rebinding check against the previous signed feed (in-place rebind, and remove-then-reintroduce-with-different-bytes, both fail; new id + `blocked` retirement is the legal path; first artifact-bound release needs no previous feed); GGUF `source_ref.digest == "sha256:" + hash`; primary-artifact consistency with the candidate row for `listed`/`recommendable` (`candidate` rows may stay `declared`); `rate_class` on every `recommendable` key; signer key-id equality with the candidate feed (a valid signature from a second concurrently trusted key fails).
- Seeded with the 10 current rows' primary MLX artifacts (hash/revision copied from the signed candidate rows, `verified`), `rate_class` per §3.3.1, and no GGUF artifacts yet.

### Class rate rows (SPEC-023-R005, §3.3.1)
- `rate-card-source.json` (`macprovider.rate-card-source.v1`, never published): all 12 published rows explicit (three credit fields each), `classes` for `class-3b/8b/20b-moe/120b-moe` only (the 30b-moe / 32b rows currently disagree with each other — reconciling them is §13 Q15 operator pricing, which rule 8 forbids doing silently), release-global share/multiplier.
[...]

Issue #1453, opened 2026-09-08
BYOM v0.2: activate catalog-matched earning — artifact sets to settlement_capable on hardware


Augustas11:
## Status as of 2026-09-18 — keep open until CLI release

v0.2 **implementation and signed SPEC-046/047 promotion are on `main`** (`0435e118`, #1583). SPEC-046-R001..R008 and SPEC-047-R001..R008 are `conformant` via signed journey-results. This issue stays open for the CLI release cut only.

**Not a close blocker:** [#1569](https://github.com/Augustas11/macprovider/issues/1569) (live `ollama_loopback` Gemma as non-earning runtime) is a v0.3 follow-on.

## Close-out to-do (ordered)

Pipeline: `docs/runbooks/byom-journey-evidence.md`. Do not edit `specs/CONFORMANCE.json` by hand.

- [x] **1. Pin the physical admission capture to a `source_sha`.** `~/byom-admission-run/run-manifest.json` (`run_id` `a03a715cb5c541ef9f5a6375ed3f0a4a`, 12/12, CLI `1.8.123`). Recaptured with `source_sha` = `9b5ddbc14ebed939cf2297cc0e99d662ea52e96b` so promotion preflight matches current selectors.

- [x] **2. Hermetic discovery evidence run (SPEC-046).** `JOURNEY-PROVIDER-BYOM-DISCOVERY` run_id `byom-discovery-20260918T033355Z`, 10/10, CLI `1.8.123`.

- [x] **3. Capture redacted evidence artifacts (unsigned).** Landed in #1580: `journeys/evidence/provider-byom-discovery-20260918T033355Z.redacted.json`, `journeys/evidence/network-model-admission-20260916T135000Z.redacted.json`.

- [x] **4. Typed capture validation.** `ADMISSION_CONTRACT.typed_capture_validation` was already on; no placeholder fixture swap required.

- [x] **5. Build unsigned journey-results + preflight.** Done in the production-release workflows and locally before sign.

- [x] **6. Operator sign.** `production-release` runs [35306375216](https://github.com/Augustas11/macprovider/actions/runs/35306375216) (SPEC-046) and [35306377225](https://github.com/Augustas11/macprovider/actions/runs/35306377225) (SPEC-047). Augustas11 dispatched; antfleet-ops approved the environment.

[...]

PR #1468, opened 2026-09-10
feat(cli): consume the SPEC-023 artifact feed with the compiled-in fallback; shared conformance corpus; SPEC-023 v0.10.2 (#1453 slice 2c)

Augustas11:
## Summary

BYOM v0.2 **slice 2c** (epic #1453): the provider CLI **consumes** the SPEC-023 artifact feed with the compiled-in fallback, and a **shared conformance corpus** pins the feed's closed schema, identity matrix, release binding, and primary consistency across the generator (Python), the coordinator (Go), and the CLI (Swift). Completes slice 2 behind 2a (generator, #1461), 2b (coordinator serving + live gate, #1463), and 2b-ii (release assets + deploy + renewal, #1465). Bundles **SPEC-023 v0.10.2** (§6 `warnings[]` lists the four artifact-feed codes and scopes its blocking sentence per §3.7.6 rule 6; §3.7.2 scopes the `release.json` signer leg to generation and manifest-holding consumers) and the matching **Malibu** allow-list change, both surfaced by the independent review. No release is cut; the committed snapshot bakes `nil` (rate-card-bound), so every provider path is byte-for-byte v0.1 until an artifact-bound release is cut.

### Generator (`scripts/catalog-release.py`)
- `generated_swift` bakes `bakedArtifactFeedBase64: String?` (base64 of the exact signed bytes, so no string escape can alter them) and `bakedArtifactFeedSignerKeyID` (the signer the sidecar carries, which the generator has already required to equal the `release.json` binding) — both `nil` for a rate-card-bound release (§3.7.2: the fallback is compiled into the binary, never a payload member). `generate` writes them; `verify` checks drift; the committed snapshot is regenerated.

### CLI (`phase3-binary/Sources/macprovider-cli/AutotuneArtifactFeed.swift`)
- `ArtifactFeed` strict decode: duplicate keys rejected, exact key sets at every level, presence-aware optional strings (a present `null` is a wrong-typed field; `verified_at` required and nullable), §3.7.4 identity matrix, global hash uniqueness, `artifact_id` grammar, primary must be `mlx_safetensors`.
- `bind(to:…)`: signer identity **equality** with the candidate feed's authenticated signer (§3.7.2 — a valid signature by a second concurrently trusted key fails) plus, on the compiled-in path, equality with the release-manifest-bound signer (a genuine three-way identity); `version` / `policy_version` equality, `generated_at` equality as the **exact release stamp string** (now also in Go; Python already did), `candidate_catalog_sha256` against the **selected** candidate bytes (§3.7.4), §3.7.5 primary consistency, model keys ⊆ candidate rows.
- `AutotuneStaticInputs.loadArtifactFeed(candidate:)` reuses the shared `loadSignedStatic` (fetch `/v1/catalog-artifacts` + sidecar, keyring, strict decode **before** the policy check per §3.5 order, older-than-baked / future / 30-day rules) and adds the four §3.7.6 classes as `AutotuneRecommendWarning` cases. Rule 5 in code: integrity / update-required / stale ⇒ no usable feed (`value == nil`, bytes kept for diagnostics). Rule 6 in code: the four warnings are absent from every blocking set, and a `nil` baked feed yields no fetch and no warnings. Freshness (§3.7.6 rules 3–4) is applied to **whichever bytes were selected, the compiled-in fallback included**, so an offline binary gets no usable feed once its baked feed is 14 days old. **Production caller:** `loadRecommendationInputs()` loads the artifact feed for the same selected candidate release beside the three v0.1 feeds; its warnings merge into every `autotune --recommend` / `--recommend-prefetch` / `--consume` and `models adopt-recommendation` transcript, and `models catalog-economics` reports them on stderr (SPEC-044's projection codes are a closed v0.1 enum) and runs discovery with a matcher built from that qualified selection.
- `BYOMCatalogMatcher.catalogKey(for:runtimeSource:)` resolves a served reference through the artifact set (HuggingFace repo ids, GGUF library tags) after the candidate rows — only through a `verified` artifact whose `allowed_runtime_sources` include the reporting adapter (§3.7.4; `declared` / `blocked` never match), of a `listed` / `recommendable` row (§3.2 ladder: `candidate` / `blocked` rows are never BYOM-matchable), and only when exactly one model key answers to the reference (a repo id / library tag is a lossy projection of content-addressed identity; an ambiguous reference mints no identity); identity only, never admission; no new SPEC-046 discovery warning codes. `models discover` is offline by design: its compiled-in matcher uses `bakedUsableArtifactFeed(now:)`, the one offline qualified selection (bound with the three-way signer identity **and** fresh at run time), nil under exactly the conditions in which the loader would yield no usable feed; `BYOMDiscoveryRunner` accepts an injected matcher for transcripts that fetched.

### Shared corpus (`scripts/tests/fixtures/artifact_feed_conformance.json`)
- One candidate catalog, one valid feed, forty-seven mutation cases with SPEC references (optional `candidate_ops` let a case pair the feed with a different catalog, e.g. an empty `models` map is accepted only with a catalog that has no `listed` / `recommendable` row — SPEC-023 §3.7.3 applies non-empty to each model's `artifacts`, not to `models`) (including the same instant spelled as a different RFC3339 form, fractional / comma-fraction stamps, an impossible `verified_at`, untrimmed release strings, `size_bytes` beyond int64 / as a float, `min_ram_gb` as a boolean, a double-slash model key, and acceptance variants). One `generated_at` grammar in all three validators (seconds precision, explicit zone, no fraction — the form `generate` stamps), one `min_ram_gb` domain ((0, 1 048 576]), `size_bytes` int64 round-trip, and trimmed header strings. Read by `ArtifactFeedConformanceCorpusTest` (Python, `validate_artifact_feed`), `TestCatalogArtifactsSharedConformanceCorpus` (Go, `LoadAutotuneFeeds`), and `testSharedConformanceCorpus` (Swift, decode + bind). Closes the LOW carried from slice 2b's audit.

## Audit
Three-lane codex loop over the FULL diff (`audits/2026-09-10-byom-v02-slice2c/`), bar 0 C / 0 H / 0 M; the anchored loop stopped after four rounds and an independent cold-context review followed.

| Round | code-reviewer | security-reviewer | architect | Closed by |
|---|---|---|---|---|
| R1 | 0 C / 3 H / 4 M | 0 C / 0 H / 2 M / 2 L / 1 I | 0 C / 0 H / 4 M / 2 L | `939d7ff7` — production caller for the feed (`loadRecommendationInputs`), verified/adapter-gated matching, manifest-bound baked signer (three-way identity), base64 bake, strict decode before policy, exact `generated_at` stamps in Go and Swift |
| R2 | 0 C / 1 H / 1 M / 3 L | **0 C / 0 H / 0 M** / 1 L / 1 I | 0 C / 0 H / 2 M / 2 L / 1 I | `7ba8578f` — one qualified selection: fallback bytes aged like fetched bytes, `bakedUsableArtifactFeed(now:)` for offline discovery, injected matcher for `catalog-economics`, row-status gating; one timestamp grammar in all three validators |
| R3 | 0 C / 0 H / 1 M / 1 L / 1 I | not re-fired (at bar) | 0 C / 0 H / 1 M / 1 L / 2 I | `012a7182` — shared `min_ram_gb` domain, ambiguity-rejecting served-reference matching, trimmed header check in Python, raw-byte duplicate-key test |
| R4 | 0 C / 0 H / 1 M | not re-fired (at bar) | **0 C / 0 H / 0 M / 0 L / 0 I** | `e858d59f` — empty `models` / `artifact_bindings` permitted per §3.7.3 / §3.7.8; corpus `candidate_ops` |
| Independent (cold-context, 3 lanes) | 0 C / 1 H / 5 M / 3 L / 3 I | 0 C / 0 H / 1 M / 3 L / 4 I | 0 C / 2 H / 6 M / 2 L / 2 I | `63e7757d` — Malibu `knownRootWarnings` (HIGH: the closed allow-list would have rejected every recommendation document at activation); content-addressed artifact matching per SPEC-047 §R001 (HIGH); no-trap compiled-in snapshot; `QualifiedArtifactFeed`; one matching authority; SPEC-023 v0.10.2; ICU `$` anchoring; corpus 55 + `case_count`; bake-signer test |
| Closure verification (cold-context, 3 lanes) | 0 C / 0 H / 1 M / 1 L / 2 I | **0 C / 0 H / 0 M** / 1 L / 4 I | 0 C / 0 H / 2 M / 0 L / 3 I | `6099576d` — artifact leg decides every key the usable feed covers (row leg = rule-6 fallback only); revisions collected over all snapshot entries, deterministic enumeration; §3.7.6 class 2 + R004 carry the §3.7.2 scoping; artifact classes excluded from `warning_state`; adopt skips the fetch; corpus 56 |
| Final verification (code + architect; security at bar) | **0 C / 0 H / 0 M** / 1 L / 3 I | at bar | 0 C / 0 H / 1 M / 2 L / 1 I | `46854810` — AC-CAT-1 carries the §3.7.2 scoping (the last restatement); uncapped names-only revision collection; runbook transcript list; `matchedArtifact(...)` provenance hook |

Carried (documented, not merge-blocking):
- **LOW** — the v0.1 loader now strict-decodes before the policy check (§3.5 order). For a fetched v0.1 feed that is BOTH schema-invalid and policy-incompatible the emitted code moves from `*_update_required` to `*_integrity_failure`; both are in every blocking set, so gating is unchanged. Pinned by `testSchemaInvalidLiveFeedIsAnIntegrityFailureNotAnUpdateRequirement`.
- **LOW / INFO** — the BYOM matcher now mints identity only for `listed` / `recommendable` rows (SPEC-023 §3.2 ladder, SPEC-047 §R001). Wire-visible for `models discover` on a catalog carrying a `candidate` or `blocked` row; zero fleet impact today (all ten committed rows are `recommendable`).
- **INFO** — `identity_state: catalog_matched` on the v0.1 row leg remains a local, advisory name-level match carrying `catalog_match_unverified`; the artifact leg is content-addressed (revision / digest). SPEC-047 continues to resolve by verified hash and to treat a provider-offered `catalog_model_key` as advisory.
- **INFO** — only the `mlx_cache` and `ollama_loopback` adapters consult the matcher (pre-existing v0.1 shape); a GGUF library tag never matches until the adapter reports the layer digest (slice 3).
- **INFO** — the shared corpus asserts accept/reject verdicts, not rejection classes; the three harnesses now assert the same `case_count` but not a shared schema version.
- **INFO** — the MLX artifact leg is gated on the observed snapshot **directory name** (the HuggingFace revision), not a locally computed hash; local `catalog_matched` stays advisory with `catalog_match_unverified`, and the coordinator resolves by verified hash (SPEC-047 §R001). Slices 3–4 treat `QualifiedArtifactFeed` provenance as identity input to that resolution, never as local verification.
- **INFO** — `ArtifactFeed.Model.primary` force-unwraps an invariant `decode` guarantees but the type does not (memberwise init is module-internal).
- **INFO** — all three validators reject a repeated adapter in `allowed_runtime_sources` and a whitespace-only `quantization`; §3.7.3 names neither (fail-closed, generator-enforced, corpus-pinned).
- **INFO** — the shared corpus pins accept/reject verdicts, not the §3.7.6 class (two hand-written Swift tests pin the integrity-vs-update boundary).
- **INFO** — the catalog-match set is non-monotonic in artifact-feed trust: with a usable feed only the artifact leg answers; once the compiled-in feed ages past 14 days between release cuts the v0.1 name-level row match returns (§3.7.6 rule 5's own fallback), always advisory with `catalog_match_unverified`.
- **Behaviour note** — on an artifact-bound release, a catalog key the usable feed covers is matched only through its artifact's immutable reference half; an Ollama model whose name merely equals such a key's row key no longer reports `catalog_matched` (it was never a catalog match per SPEC-047 §R001). Keys the feed does not cover keep v0.1 name matching.
- **INFO** — the `models catalog-economics` transcript's artifact-warning stderr lines have helper-level coverage only (no command-level test; needs injectable static inputs on the subcommand).

## Validation
`swift test --filter 'AutotuneArtifactFeedTests|AutotuneCommandTests|BYOMDiscovery|ModelCatalogEconomics|ModelsSubcommand|AutotuneRecommendTests'` 338 tests, 0 failures (27 in `AutotuneArtifactFeedTests`); Malibu `xcodebuild test … -only-testing:MalibuTests/ModelManagementTests` 96 tests, 0 failures; `scripts.tests.test_catalog_artifact_feed` 139 tests OK; `go vet` + `go test ./internal/buyer` (full package) ok; `python3 scripts/catalog-release.py verify` (committed release unchanged, snapshot regenerated with a `nil` bake); `bash scripts/test-catalog-release.sh` PASS; `check_spec_governance.py` passed; local full `swift test` reaches the known env-only MLX metallib abort with 0 failures before it (CI covers the exact-frame goldens).

BYOM epic gates:
- Parent epic: #1453 (slice 2c)
- Slice issue: #1453
- Authority boundary: SPEC-023 §3.7 consumption in the provider CLI (+ v0.10.2 §6 / §3.7.2 amendment); SPEC-046 discovery identity (matcher only, content-addressed per SPEC-047 §R001); SPEC-044 Malibu decode allow-list (no new codes on the projection)
- SPEC requirements: SPEC-023-R004, SPEC-046-R003 (identity_state semantics unchanged; no new discovery warning codes), SPEC-044-R002 (projection codes unchanged; Malibu root-warning allow-list widened), SPEC-047-R001 (`catalog_matched` never minted from a mutable name)
- User-facing state/copy: none (warning codes are SPEC-023 recommend warnings, localisation-safe enum strings)
- Machine schema: none new on the wire; four new `AutotuneRecommendWarning` values defined by SPEC-023 §3.7.6
- Security/privacy: signer equality, release binding, fail-closed artifact capabilities; no key material; test keys generated at test time
- Runtime/liveness: no behaviour change while the baked feed is `nil`; artifact-feed failures never block paid recommendation or coordinator join (§3.7.6 rule 6)
- Identity/economics/settlement provenance: no billing or settlement change; only PRIMARY artifacts can ever bind pricing/settlement (SPEC-023 §3.7.7, enforced by later slices)
- Rollout/rollback/compatibility: baked `nil` = v0.1; Stage A nine-name map and seventeen index roles untouched
- Explicitly deferred or not applicable: SPEC-010 multi-artifact identity + GGUF runtime hash (slice 3), catalog binding + coordinator decision path (slice 4), first artifact-bound release cut (operator)

{
  "behavior_change": "yes",
  "contract_change": "yes",
  "issue": "https://github.com/Augustas11/macprovider/issues/1453",
  "specs": ["SPEC-023", "SPEC-044", "SPEC-046", "SPEC-047"],
  "requirements": ["SPEC-023-R004", "SPEC-044-R002", "SPEC-046-R003", "SPEC-047-R001"],
  "authority_domains": ["installer-autotune-policy", "provider-byom-discovery"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "phase3-binary/Tests/macprovider-cliTests/AutotuneArtifactFeedTests.swift",
    "phase4-coordinator/internal/buyer/catalog_artifacts_conformance_test.go",
    "scripts/tests/test_catalog_artifact_feed.py",
    "scripts/tests/fixtures/artifact_feed_conformance.json",
    "phase3-binary/app/Tests/MalibuTests/ModelManagementTests.swift"
  ],
  "journeys": ["not-required"]
}


antfleet-ops (dismissed):
Reviewed slice 2c: four anchored codex rounds, an independent cold-context three-lane review, a closure pass, and a final pass — records in audits/2026-09-10-byom-v02-slice2c/. Security lane at 0/0/0 since the closure pass; code and architecture at 0/0/0 on the final commit. SPEC-023 v0.10.2 bundled with the implementation per the repo rule.

antfleet-ops (dismissed):
Re-approving after the spec-index regeneration commit (f6d0db38, specs/README.md only).

antfleet-ops (approved):
Re-approving after the rebase onto main (836d7e59, 422fc2f1); no content change to the slice, generated spec index re-derived.

Issue #1478, opened 2026-09-11
BYOM v0.3: LM Studio + llama.cpp loopback adapters with GGUF hash (discovery, not production earning)


Augustas11:
@SmtTheSE — this is yours.

Parent epic: #1453 (BYOM v0.2). This is the deferred v0.3 adapter slice, **not** a rewrite of v0.2 slices 5–7.

## Premise

A provider who already runs LM Studio or llama.cpp on loopback can `models discover` / `models evaluate` those local GGUF models as first-class `lmstudio_loopback` / `llamacpp_loopback` candidates, with the CLI hashing the served GGUF bytes itself (`macprovider.gguf-file.v1`). Catalog match is hash-based and may light up if the bytes match a verified artifact. Production `settlement_capable` / paid credit is **out of scope** until #1453 slice 7 has paid an Ollama or MLX catalog-matched node.

`openai_compatible_loopback` is not a substitute. It is opaque on purpose: no artifact bytes, `identity_state: opaque_endpoint`, offer submit refuses it, and a `verified` catalog artifact must not allow that runtime (SPEC-023 §3.7.4).

## Current code (do not rebuild)

Implemented adapters in `phase3-binary/Sources/macprovider-cli/BYOMDiscovery.swift`: `mlx_cache`, `ollama_loopback`, `openai_compatible_loopback` only.

Names already reserved, not implemented:

- SPEC-046-R002 enum: `lmstudio_loopback`, `llamacpp_loopback`
- SPEC-023 GGUF `allowed_runtime_sources` already permits both
- GGUF hashing is Ollama-store only: `BYOMModelAdmission.artifactEvidence` returns `nil` unless `runtimeSource == "ollama_loopback"`

Copy the Ollama adapter over the existing #1246 harness. Do not reimplement URL admission, redirect policy, redaction, or parser limits:

- `BYOMLoopbackOriginValidator`
- `BYOMDiscoveryHTTPBounds`
- `StrictJSONParser` / `BYOMDiscoveryJSON`
- `BYOMDiscoveryPrivacy`

## Scope

1. Thin `lmstudio_loopback` and `llamacpp_loopback` discovery adapters.
2. Well-known loopback defaults, same shape as Ollama's `http://127.0.0.1:11434` (not the no-default openai-compatible adapter):
   - LM Studio: `http://127.0.0.1:1234`
   - llama.cpp `llama-server`: `http://127.0.0.1:8080`
[...]

Issue #1346, opened 2026-09-02
Mac E2E: pre-release candidate (malibu-cli + coordinator) on a local isolated stack (@SmtTheSE)


Augustas11:
## Why this issue exists

The changes to the **provider CLI** (`malibu-cli`/`macprovider-cli`) **and the coordinator** merged since the last release (**v1.8.117**) are on `main` and **unreleased**. Linux CI proved the unit contracts, but the real behavior of the full candidate — the CLI rebrand/`doctor`/telemetry, and the coordinator-side onboarding/revoke/admission changes — needs a live E2E before a release candidate can be cut.

This is the **CLI-side counterpart to #1341** (which covers the Malibu.app GUI dashboard). @SmtTheSE — you drive the whole candidate stack.

**Approach: run the entire candidate locally.** Build **both** the candidate coordinator and the candidate CLI from `origin/main` HEAD, and run them together on an isolated node you control. Both sides are the candidate, so **all** post-117 changes are exercised — CLI *and* coordinator (#1305/#1306/#1313). Everything is local and isolated: your own coordinator, your own SQLite DB, **no Postgres, payout off, settlement observe**. **No production contact, no real USDC, no signing keys, no exposed endpoints.**

**Do not cut a release. Validation only.**

## What changed since v1.8.117 (all on `main`, unreleased)

**CLI:** #1343 `malibu-cli` rebrand + ownership-guarded symlink · #1335 CLI money-path telemetry truth under drift · #1319/#1318 read-only `doctor` · edb4d3c7 autotune context calibration · 3360d2cf adoption RAM-fit · #7931dfaf headless install + stranded-provider adoption · #1339 gpt-oss-120b catalog.

**Coordinator:** #1305 operator onboarding funnel · #1306 API-key revoke console messaging · #1313 promoted-provider admission telemetry stays live after promotion · #1299 fail-closed billing recovery/settlement.

## Hard constraints

1. **Do not cut a release.** Build from current `origin/main`.
2. **Do not run this against your live earning provider.** You just joined as a provider — a second `macprovider-cli` instance contends over the control socket, launchd, and auto-repair and **can disrupt a live provider**. Use an **isolated node**: a separate macOS user account (own `HOME` + launchd domain) or a throwaway box.
3. Everything runs locally on `127.0.0.1` with an isolated config — no prod, no keys, no network exposure.

## 1. Build both from HEAD

```bash
git fetch origin main && git checkout --detach origin/main
[...]

Issue #1246, opened 2026-08-28
BYOM epic gate: centralize adapter safety, privacy, and liveness controls


Augustas11:
Parent epic: #1240

## Gate

All BYOM runtime adapters must use one shared safety layer rather than per-adapter network and parsing logic.

## Required Controls

- Loopback-only endpoint validation.
- No LAN, public, wildcard, private non-loopback, link-local, multicast, unresolved, redirected, proxied, hostname-expanded, or Unix-socket targets.
- No port scanning.
- Short timeouts.
- Bounded response headers, decoded body bytes, JSON nesting, and parser work.
- Redacted diagnostics.
- Closed warning codes.
- Discovery is read-only.
- Evaluation has bounded timeout/token/request/output limits and no production serving mutation.

## Required Artifact

A shared adapter harness with unit tests for every rejection and bound above. Adapter implementations for Ollama, LM Studio, llama.cpp, and OpenAI-compatible loopback should be thin wrappers around that harness.

## Pass/Fail

Passes when:

- The first adapter PR proves the shared safety harness independently from any one runtime.
- Adding a new adapter does not require reimplementing URL admission, redirect policy, redaction, or parser limits.

Fails when:

- An adapter can contact non-loopback addresses.
- Raw endpoints, paths, credentials, prompts, completions, or adapter error bodies leak into JSON/logs.
- Discovery starts runtimes, downloads weights, edits provider config, or changes the current serving model.

Augustas11:
## Gate satisfied — one shared adapter safety layer, every rejection and bound proven

Closing on code + tests + runtime evidence, mapped to the Required Artifact and Pass/Fail rows.

### Shared harness (single place, no per-adapter network or parsing logic)

All runtime access in `phase3-binary/Sources/macprovider-cli/BYOMDiscovery.swift` goes through:
- `BYOMLoopbackOriginValidator` (`validatedHTTPOrigin` / `isSafeLoopbackHTTPURL`) — admits only literal `127.x.x.x` dotted-quad or `::1` with an explicit port over plain `http`; used by discovery, evaluation, the admission client, and the Ollama adapter.
- `BYOMDiscoveryHTTPBounds` (64 KiB headers, 256 KiB decoded body) + `BYOMURLSessionHTTPClient` (no redirects, no proxy, short per-request timeout).
- `StrictJSONParser` with a hard nesting bound (depth 32) feeding `parseOllamaTags` / `parseChatCompletions`.
[...]

PR #1480, opened 2026-09-11
feat(byom): v0.3 lmstudio_loopback + llamacpp_loopback adapters with CLI-side GGUF hash (#1478)

SmtTheSE:
## Summary

BYOM v0.3 adapter slice (#1478, parent epic #1453): `lmstudio_loopback` and `llamacpp_loopback` as thin copies of the Ollama adapter over the #1246 harness, with the CLI hashing the served GGUF itself (`macprovider.gguf-file.v1`). Discovery + evaluate + offer `--dry-run` only; no earning claim (`earning_path_class: no_earning_path_in_v0_1`).

### Locator protocol (`BYOMArtifactDigest.swift`)
The digest pipeline (descriptor-bound identity, streamed SHA-256, cache) was already runtime-neutral; only *served name to file* was Ollama-specific. `BYOMGGUFArtifactLocator` puts that step behind a protocol. `BYOMOllamaModelStore` conforms unchanged, every Ollama-named resolver method is kept as a shim, and `BYOMArtifactDigestTests` passes untouched, the refactor is behaviour-preserving by construction.

### `lmstudio_loopback`
- Inventory: LM Studio's native `GET /api/v0/models` (the OpenAI-shaped `/v1/models` parses too; only `id` is required). `compatibility_type` / `arch` / `quantization` / `max_context_length` sanitized exactly as Ollama's `details`.
- Artifact leg: `BYOMLMStudioModelStore` resolves the served **id** to exactly one `.gguf` under the operator-declared models root (default `~/.lmstudio/models`, `MACPROVIDER_LMSTUDIO_MODELS_ROOT`), symlink-resolved and root-contained, failing closed when zero or several files answer. MLX-format entries have no GGUF to `runtime_reported`. `identity_state` is `catalog_matched` / `artifact_hash_available` / `runtime_reported`, never `opaque_endpoint` (SPEC-023 §3.7.4).
- Default `http://127.0.0.1:1234`, `--lmstudio-origin` / `--skip-lmstudio` on all six `models` subcommands.

### `llamacpp_loopback`, two decisions worth reading
1. **Served reference is the file stem, never the path; the served path is the proof.** llama-server's `/v1/models` id is the loaded file's *path* unless `--alias` is set (confirmed live; `model_alias` carries it too), and the harness already refuses any reference containing `/`, a path on the wire would leak the operator's filesystem into `candidate_id`, display names and evidence JSON. Each id is reduced to its stem before the safe-reference check, and `BYOMLlamaCppModelStore` resolves that stem only where the **operator** says, in one of two modes, **neither with a default**:
   - **(a) `--llamacpp-model-root`** / `MACPROVIDER_LLAMACPP_MODEL_ROOT`, exactly one `<stem>.gguf` under the directory, 1-2 levels, symlink-resolved, root-contained; zero or several means no identity.
   - **(c) `--llamacpp-model-path`** / `MACPROVIDER_LLAMACPP_MODEL_PATH`, the one pinned file; overrides (a), needs no root, and its stem must equal what the server serves (else no identity, never a wrong one).
   In both modes the path llama-server reports in `/props.model_path` must equal the operator-resolved file before any digest is read, computed, or re-validated (discovery, evaluate, offer, and again before the signed package leaves); it is never stored or emitted. A same-stem file the runtime is not serving is never hashed. Selection is decided once from one source: a CLI flag ignores both environment variables, and two selectors from the same source are an explicit error. With neither set every candidate is `runtime_reported`. Both modes are tested and proven on hardware (#1478); recommendation is (a) as the documented operator path with (c) as the strict alternative, a docs-only choice.
2. **Fingerprint before inventory.** llama-server's default `:8080` is also macprovider's own `serve` port and both answer `/v1/models`, an unfingerprinted default-on adapter would list a provider Mac's own served model as a llama.cpp candidate. The adapter requires `GET /props` with `default_generation_settings` first; anything else on the origin is `unavailable`, never a candidate. `/props` also supplies the context window.

Default `http://127.0.0.1:8080`, `--llamacpp-origin` / `--skip-llamacpp` / `--llamacpp-model-root` / `--llamacpp-model-path`.

### Shared wiring
`artifactEvidence`'s Ollama-only guard to `BYOMDiscoveryEnvironment.artifactHashingRuntimes` (three runtimes; `openai_compatible_loopback` deliberately absent). Evaluate hashes on evaluate for all three, like Ollama. Both adapters are default-on like Ollama, so the CLI-driven tests that isolate one adapter with `--skip-ollama` now also pass `--skip-lmstudio --skip-llamacpp`, no other test changed.

## Audit
The 3-lane Codex audit on this PR (HIGH binding, two MEDIUM, four LOW/INFO) is fully addressed in commits `141593de` through `9d0c7833`; the per-finding response with live re-verification is on the PR thread.

## Validation
- `swift test --filter 'BYOM|ModelsSubcommand'` to **171/171** (19 new in `BYOMLoopbackAdapterTests.swift`: non-loopback rejected before dispatch, skip means no row + zero requests, hashed vs `runtime_reported` and never opaque, ambiguity fails closed, `/props` fingerprint gates inventory, pathtostem with no path on the wire, symlink escape + nil root rejected, pinned file wins over root and refuses a stem mismatch, hashed through discovery with only a pinned file, malformed means warning, dry-run claims no settlement).
- Broader suites green on this machine; the MLX-runtime suites cannot run here (the #1346 metallib toolchain gap) and rely on CI.
- `scripts/check_spec_governance.py` passed; `git diff --check` clean; `Package.resolved` untouched.

## Hardware pass (scope item 7), done, evidence on #1478
Real-Mac discover to evaluate to discover to offer `--dry-run` against **live llama-server 0.4.0 and live LM Studio**, one 138 MB GGUF served by both. Both `health_result: passed`; both reach `artifact_hash_available`; the CLI's `macprovider.gguf-file.v1` is byte-identical to `shasum -a 256` and **identical across the two runtimes** (same bytes, two paths, one digest). Both dry-runs report `earning_path_class: no_earning_path_in_v0_1`. All safeguards exercised live: no-root means `runtime_reported`, default `:8080` means `unavailable`, non-loopback means `rejected` with no origin on the wire, all-skipped means no rows.

Confirmed live: LM Studio `:1234` default and `/api/v0/models` shape match the parser field-for-field, and the inferred idtodirectory rule needed no adjustment; llama-server's `/v1/models` id **is the file path**, which is why the stem design exists. Two things hardware surfaced were fixed in `d3bbd79f`: LM Studio's default *embedding* model was listed as a candidate (non-`llm` types now skipped), and four CLI-driven tests that isolated Ollama via `--ollama-origin` reached the real `:1234` once LM Studio was running (every `Models*Command.parse` block now carries both skip flags).

**Both llama.cpp modes ship**; which the runbook recommends is your call (docs-only). MLX-format LM Studio entries stay `runtime_reported` (servable, no GGUF); non-`llm` entries are dropped. Option (c) evidence is on #1478 alongside the rest.

## What this does not do
No SPEC-047 state-machine, billing, `source_ref.kind`, Malibu app, catalog cut, LAN/scan, weight download, or serve-config change. No candidate from either adapter can reach `settlement_capable`.

BYOM epic gates:
- Parent epic: #1453 (v0.3 adapter slice)
- Slice issue: #1478
- Authority boundary: SPEC-046-R002 loopback discovery adapters + SPEC-010 v1.7 R007(a) CLI-side artifact digest
- SPEC requirements: SPEC-046-R002, SPEC-046-R003, SPEC-046-R007, SPEC-010-R007, SPEC-023 §3.7.4
- User-facing state/copy: none new (existing `identity_state` / guidance values)
- Machine schema: none new (`runtime_source` values already reserved in the SPEC-046-R002 enum)
- Security/privacy: operator-rooted hashing; runtime never selects the hashed file; no filesystem path reaches the wire; llama-server fingerprint prevents self-discovery on `:8080`
- Runtime/liveness: n/a
- Identity/economics/settlement provenance: none, dry-run asserts `no_earning_path_in_v0_1`
- Rollout/rollback/compatibility: `byom-disablement-rollback.md` rows 12-13; both adapters skip-able; unsetting the llama.cpp root disables its artifact leg
- Explicitly deferred or not applicable: settlement (#1453 slice 7)

{
  "behavior_change": "yes",
  "contract_change": "none",
  "issue": "https://github.com/Augustas11/macprovider/issues/1478",
  "specs": ["SPEC-046", "SPEC-010", "SPEC-023"],
  "requirements": ["SPEC-046-R002", "SPEC-046-R003", "SPEC-046-R007", "SPEC-010-R007"],
  "authority_domains": ["provider-byom-discovery"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "phase3-binary/Tests/macprovider-cliTests/BYOMLoopbackAdapterTests.swift",
    "phase3-binary/Tests/macprovider-cliTests/BYOMArtifactDigestTests.swift",
    "phase3-binary/Tests/macprovider-cliTests/BYOMDiscoveryTests.swift"
  ],
  "journeys": ["not-required"]
}


Augustas11:
3-lane Codex audit completed against PR #1480 (`origin/main` `7128d120` -> head `e44ac107`). Verdict: **request changes / do not merge yet**.

## Blocking finding

**HIGH correctness / MEDIUM security: llama.cpp artifact evidence is not bound to the file the runtime is serving.**

`BYOMLlamaCppDiscovery` fingerprints `/props` but discards `model_path`, then builds candidates from a stem-only `servedModelRef`. `BYOMLlamaCppModelStore` strips path-shaped runtime IDs down to the filename stem and resolves any same-stem GGUF under the operator root or pinned path. That means a misconfigured or hostile loopback llama.cpp server can serve one file while MacProvider hashes and submits/catalog-matches another same-stem local file.

Evidence:

- `phase3-binary/Sources/macprovider-cli/BYOMDiscovery.swift:4074` fetches `/props`, but the path is not used to bind artifact evidence.
- `phase3-binary/Sources/macprovider-cli/BYOMDiscovery.swift:4144` / `:4151` build evidence from the stem-only served ref.
- `phase3-binary/Sources/macprovider-cli/BYOMArtifactDigest.swift:430` reduces runtime IDs to stems.
- `phase3-binary/Sources/macprovider-cli/BYOMArtifactDigest.swift:447` / `:461` resolve a matching same-stem pinned/root GGUF.
- `phase3-binary/Tests/macprovider-cliTests/BYOMLoopbackAdapterTests.swift:158` through `:186` currently lock in the mismatch: runtime reports `/Users/someone/private-models/tiny-q4.gguf`, while `allowed/tiny-q4.gguf` upgrades to `artifact_hash_available`.

Suggested fix: keep filesystem paths off the wire, but use them internally as proof. For llama.cpp, canonicalize `/props.model_path` or a path-shaped `/v1/models.id`; for pinned mode require exact canonical equality with the pinned file, and for root mode require the runtime path to be contained in the operator root and equal the resolved hit before emitting digest evidence. If runtime path evidence is absent or unverifiable, leave the candidate `runtime_reported`. Add negative regression tests for same-stem different-path root and pinned mismatches.

## Additional security findings

**MEDIUM: environment pin overrides explicit CLI root.**

`BYOMDiscoveryEnvironment.production` resolves `llamacppModelRoot` and `llamacppModelPath` independently, while `BYOMLlamaCppModelStore` always gives the pin precedence. So `--llamacpp-model-root /approved` can be silently defeated by inherited `MACPROVIDER_LLAMACPP_MODEL_PATH`.

Suggested fix: resolve root/path selection atomically. If either CLI selector is supplied, ignore both environment selectors, or reject conflicting CLI selectors with an explicit error.

**MEDIUM: unloaded LM Studio models are marked ready and offerable.**

LM Studio’s documented `/api/v0/models` response lists loaded and downloaded models and includes `state`, but the PR discards `state` and assigns every LLM `readiness_state: "ready"`. Offer submission permits unevaluated candidates, so a downloaded but unavailable model can produce an offerable candidate.

Suggested fix: parse a closed set of LM Studio states. Only loaded instances should be `ready`; unloaded/not-loaded entries should map to `requires_preparation` or equivalent local-only state and block offer submission until serviceability is proven.

## Lower-priority follow-ups

- LOW: digest cache persists private absolute model paths; PR #1480 extends that exposure to LM Studio and arbitrary llama.cpp pinned paths. Prefer a non-reversible keyed path token or root-relative locator plus metadata.
- LOW: directory traversal budget does not bound allocation because `contentsOfDirectory` materializes arrays before `maxEntriesVisited` checks. Use a streaming enumerator and count all entries.
- LOW: the generic resolver still exposes an Ollama-specific forced cast accessor; remove it or make it optional/throwing.
- INFO: LM Studio comments imply `/v1/models` fallback, but discovery only requests `/api/v0/models`; either implement the fallback or narrow the comment.

## Verification run during audit

- `swift test --filter BYOMLoopbackAdapterTests` passed: 12 tests, 0 failures.
- The code lane also reported `swift test --scratch-path /private/tmp/pr1480-swift-scratch --filter 'BYOM|ModelsSubcommand'` passed: 164 tests.

These passes do not clear the blocker because the current tests encode the unsafe same-stem binding behavior.

SmtTheSE:
## Audit response: all findings addressed, five commits, live re-verification of the HIGH

Head moves `e44ac107` to `9d0c7833`. 171/171 BYOM and ModelsSubcommand tests, broader suite green (MLX-runtime suites excluded on this machine per the #1346 metallib gap), governance check clean, `Package.resolved` untouched. Every item below is one commit with its own tests; the HIGH was re-run against a real llama-server with the audit's exact scenario.

| Finding | Commit | What changed |
|---|---|---|
| HIGH: llama.cpp evidence not bound to the served file | `141593de` | The served path is now proof at every evidence site. |
| MEDIUM: environment pin overrides explicit CLI root | `7e6c1980` | Selection decided once, from one source. |
| MEDIUM: unloaded LM Studio models marked ready | `98a52bff` | Only `state: "loaded"` is ready. |
| LOW: digest cache persists private paths | `1f3b9cc9` | Persists a SHA-256 path token; cache schema v2. |
| LOW: traversal budget does not bound allocation | `141593de` | Streaming enumerator, every entry counted. |
| LOW: forced-cast Ollama accessor | `141593de` | Removed. |
| INFO: promised `/v1/models` fallback not implemented | `98a52bff` | Implemented, on 404 only. |
| Runbook rows 12 and 13 | `9d0c7833` | Reflect all of the above. |

### HIGH: binding to the served file

You were right that the stem-only design kept paths off the wire but never used the runtime's report as proof. `BYOMGGUFArtifactLocator.resolveArtifact` now takes `runtimeArtifactPath`. Ollama and LM Studio ignore it (they bind through manifest and directory from the name). llama.cpp requires it: the path must be absolute and its stem must equal the served stem; then in pinned mode it must equal the pinned file, and in root mode it must be contained in the root and equal the single stem-resolved hit. No path, a relative path, a stem mismatch, or any inequality means no artifact leg, never a wrong one.

The path is fetched fresh from `/props` at each site: discovery (from the fingerprint body already in hand), evaluate, offer (`artifactEvidence` is now async and takes the HTTP client), and again immediately before the signed package leaves, so `validateCurrent` also checks that the runtime is still serving the bound file. It is never stored or emitted; the wire assertion that no path appears is kept in every test.

The two tests that encoded the unsafe behaviour are replaced by your negative case and its positive: `testLlamaCppSameStemFileUnderRootIsNotHashedForAFileServedElsewhere` (decoy under root, file served elsewhere, digest already cached: stays `runtime_reported` and `computeEvidence` throws) and `testLlamaCppIsHashedOnlyWhenTheServedPathIsTheRootResolvedFile` (hashed only when they coincide; `validateCurrent` with a different served path fails closed). Pinned mode has the same pair.

Live, against llama-server 0.4.0, with a same-stem decoy under `--llamacpp-model-root` whose bytes differ from the served file:

| Scenario | Result |
|---|---|
| A. root holds the decoy; server serves a file elsewhere | evaluate `passed`; `runtime_reported`; no digest computed, no cache file written |
| B. root is where the served file lives | `artifact_hash_available`; digest `5a1395716f79...` equals the served file's shasum, not the decoy's `3e2c052a...`; cache persists `path_sha256`, no `path` |
| C. pin the decoy; server serves elsewhere | `runtime_reported` |
| D. `--llamacpp-model-root` and `--llamacpp-model-path` together | exit 1: `--llamacpp-model-root and --llamacpp-model-path are mutually exclusive; pass exactly one` |
| D. inherited `MACPROVIDER_LLAMACPP_MODEL_PATH` plus explicit `--llamacpp-model-root` | environment ignored, root wins |

### MEDIUM: selector precedence

`BYOMLlamaCppArtifactSelector.resolve(cliRoot:cliPath:environment:)` owns the decision. Any CLI selector ignores both environment selectors. Two selectors from the same source, CLI or environment, are a described error rather than a precedence rule. The six `models` subcommands resolve it before building the environment, and the store no longer reads the environment. `testLlamaCppEnvironmentPinCannotDefeatAnExplicitCLIRootEndToEnd` covers the rogue-pin case you described.

### MEDIUM: LM Studio load state

`state` is a closed enum; exactly `loaded` is serviceable. `not-loaded`, an unrecognized value, and an absent field all map to `readiness_state: requires_preparation` with the blocking warning, so the candidate is `local_only` until `models evaluate` drives a load and proves a completion. A non-string `state` is `malformed`, not silently ready. The real `/api/v0/models` capture on #1478 shows `"state": "not-loaded"` before the model was loaded, which is the value this maps.

### LOW and INFO

Cache: `BYOMArtifactFileIdentity.path` is in memory only and excluded from equality; `path_sha256` is persisted and compared, which still separates same-inode files on different paths. Schema bumped to v2 so existing v1 files with raw paths are discarded on load and rewritten on the next store. Traversal: `FileManager.enumerator`, descendants skipped past two levels, every entry counted against the budget before any filtering. Accessor: removed; the `init(store:cache:)` shim remains for existing tests. Fallback: `/api/v0/models` 404 falls through to `/v1/models`; nothing from the fallback carries `state`, so nothing from it is ready. Any other non-200 is still `unavailable`.

Redacted JSON for A through D is attached on #1478 alongside the earlier evidence.

Augustas11:
Follow-up verification on the updated head `9d0c7833` looks good from the prior audit perspective.

I re-checked the contributor response against the follow-up diff and then ran the focused local gate in a clean temporary checkout of the PR head:

```bash
cd phase3-binary
swift test --scratch-path /private/tmp/pr1480-followup-swift-scratch --filter 'BYOM|ModelsSubcommand'
```

Result: **passed** — 171 selected tests, 0 failures.

The important regressions from the audit are now covered in the selected suite, including:

- llama.cpp same-stem decoy under root stays `runtime_reported` instead of hashing the wrong file.
- llama.cpp hashes only when the served path equals the root-resolved/pinned file.
- missing served path or stem mismatch has no artifact leg.
- explicit CLI llama.cpp root cannot be defeated by inherited environment pin.
- LM Studio only marks `loaded` models ready/offerable, and `/v1/models` fallback remains non-ready.
- digest cache persists `path_sha256` rather than the absolute path and discards v1 cache files.

So I consider the earlier blocking audit finding resolved on `9d0c7833`.

Remaining process caveat: GitHub currently reports no status checks for this branch and the PR is marked `BEHIND`, so the next step should be to update/rebase the branch and let the normal checks run. From the audit/follow-up verification side, this is unblocked once that branch hygiene is handled.

antfleet-ops (approved):
Approving on the completed audit trail: the 3-lane Codex audit (1 HIGH, 2 MEDIUM, LOW/INFO) was fully addressed across commits 141593de→9d0c7833, with live re-verification of the served-file binding on real llama-server + LM Studio, and re-confirmed on the final head (171/171 BYOM+ModelsSubcommand tests, blocking finding resolved). Rebased cleanly onto origin/main (zero file overlap with intervening commits, so the audited diff is preserved). Discovery-only slice: no earning path, no settlement, dry-run asserts no_earning_path_in_v0_1. Merging once ci-required is green.

Issue #1486, opened 2026-09-11
BYOM v0.2 slice 7 — settlement journey on hardware (capture handoff → @SmtTheSE)


Augustas11:
Handoff of **BYOM v0.2 slice 7** (epic #1453). Capture half → @SmtTheSE; operator signing stays with @Augustas11.

Full brief committed at `audits/2026-09-11-byom-v02-handoffs/SLICE7_SETTLEMENT_JOURNEY_HANDOFF.md` (main @ `4bbb7eed`).

**Goal:** drive one candidate to `settlement_capable` on real Apple Silicon, capture `JOURNEY-NETWORK-MODEL-ADMISSION`, then (operator) sign it; promote SPEC-047-R001..R008; cut a release per `docs/runbooks/provider-cli-release-verification.md`.

**@SmtTheSE (capture):** run the full candidate stack locally on real hardware (as with #1346) — discover → offer → offer-time catalog match → operator decision → route → settlement — and capture the run manifest + evidence. redastare canary Mac available via @Augustas11.

**@Augustas11 (not delegable):** the dual-control `settlement_capable` decision and the operator signing of `JOURNEY-NETWORK-MODEL-ADMISSION` — operator credentials.

**Constraints:** GGUF cannot reach `settlement_capable` until SPEC-010 R007(e) — pick a settleable candidate or capture the GGUF stop case explicitly. Release-asset proof per the runbook (SHA-256 byte identity of both embedded `macprovider-cli` binaries; verify the updater path; never `codesign --force --deep`).

Note: @SmtTheSE is not a repo assignee yet; @Augustas11 to add collaborator access for formal assignment.

SmtTheSE:
Acknowledged. Taking the capture half. Read the handoff at `4bbb7eed`, the journey contract, the evidence runbook, and the manifest fixture.

## What the capture actually requires

`JOURNEY-NETWORK-MODEL-ADMISSION` is twelve steps, sixteen observations that must be true, five that must be false, and `observations.money_path_zero_rows` with all ten ledgers at zero, `request_log` included. So step 9 proves the `settlement_capable` state and that the route-time snapshot and receipt gates are armed; it does not move money. No gateway and no buyer traffic in the run, which keeps the rig to coordinator, Postgres, and the provider CLI.

[...]

PR #1495, opened 2026-09-12
byom: physical-provider driver for JOURNEY-NETWORK-MODEL-ADMISSION (#1486)

SmtTheSE:
## Summary

BYOM v0.2 slice 7 (#1486, parent epic #1453): a physical-provider driver for `JOURNEY-NETWORK-MODEL-ADMISSION`. `run-cli-onboarding-e2e.py --journey-evidence` runs the twelve journey steps against a real candidate CLI, a self-hosted coordinator with two distinct operator actors, and the coordinator Postgres, and writes a `macprovider.byom-journey-run.v1` run manifest that the existing `capture-byom-journey-evidence.py --journey admission` path consumes unchanged. The runner is aligned to SPEC-047 v0.1.9 (`offer_rejected` reserved and unreachable; 15 true observations). The R1 audit findings (4 HIGH, 4 MEDIUM) and the R2 findings (4 HIGH, 2 MEDIUM) are addressed in the last two commits; per-finding responses are on the thread.

### What the driver asserts, step by step
Every assertion below was read from the coordinator and CLI source, not assumed; the FakeRig in the unit test carries the same contract-exact document shapes so a drift in either side fails the test.

1. **Offer dry-run**: settleable MLX candidate reaches `catalog_matched`; dry-run claims `no_earning_path_in_v0_1` and appends no event.
2. **Signed offer**: submit returns `offer_submitted` from the coordinator with a `coordinator_event_id`; the runner waits for the coordinator's own probe edge (`sandbox_probe_only`), reads the head, attempts a second live offer, and requires `HTTP 409` on the CLI's stderr with the same `coordinator_event_id` afterwards (appended nothing, measured).
3. **Opaque endpoint**: the CLI refuses to offer an `opaque_endpoint` candidate (`not offerable`), and the coordinator never sees it; `local_default` / `local_inventory_only` guidance is captured.
4. **Sandbox probe only**: no money-path row for the candidate while probing.
5. **Network visible unpriced**: `network_visible_unpriced`, `economics_state` unpriced, four money fields zero.
6. **Catalog matched, not settlement**: the settleable candidate reaches `catalog_priced` with the signed catalog rates shown (`rate_source` live_signed or baked_signed, `catalog_economics_permitted == true`) and `settlement_capable == false`; no buyer debit, no ledger row.
7. **Revocation on drift**: operator-supplied `--drift-hook`; the candidate reaches `revoked` with `provider_guidance.transition_reason_code` inside the closed SPEC-047-R006 drift set; an operator-origin revoke is rejected as not-drift. Re-offer appends a fresh signed event.
8. **Withdrawal**: `models admission withdraw` reaches `withdrawn` (new `model_admission_withdraw.v1` validator in `byom_journey_evidence.py`); re-offer appends a fresh signed event.
9. **Settlement-capable case**: `catalog_priced` then `settlement_capable` through dual control at `/decisions/<id>/approve` with the closed approval body (evaluated head + idempotency key): the proposer's own approval must be exactly `409 dual_control_required`, then a distinct actor under `auth.operator_keys` approves; route-snapshot and receipt-verdict ledgers stay at zero.
10. **Admission status presentation**: the catalog-matched candidate presents an earning path; the GGUF candidate (`--gguf-ref`, required) is the novel non-catalog candidate and must report `earning_path_class: no_earning_path_in_v0_1` (SPEC-010 R007(e)); both carry state meaning and next action.
11. **Transition validity**: the head is re-read, one operator next-state outside `ADMISSION_ALLOWED_NEXT_STATES[current]` must be refused with exactly `409 invalid_transition`, and state and head are unchanged. The fresh-evidence-on-re-entry invariant (R001/R006) is proven on the reachable `revoked` (step 7) and `withdrawn` (step 8) paths; there is no rejection leg (v0.1.9).
12. **Redaction review**: six surfaces (runner log, verbatim CLI and operator-surface transcript, every captured document, the coordinator's offer listing, provider and coordinator log bytes appended during the run) are checked for both operator secrets, the DSN and its password, the shared evidence scanner (credential shapes, URLs, absolute and home paths, IPv4/IPv6 literals, localhost), the synthetic probe prompt and chat request shapes, and chat completion shapes; only then are `secret_field_persisted`, `raw_prompt_logged`, `raw_completion_logged` set `false`. SPEC-047-R008 is attached here.

Money path: the ten `money_path_zero_rows` ledgers (including `request_log`) are read from Postgres before and after; any non-zero delta fails the run and **no manifest is written**.

### Design decisions
- **Fail closed, publish nothing on failure.** `--out` must be new or empty; the manifest is written atomically once, only after all twelve steps and the ledger check pass. A failing run cannot leave a partial manifest that the capture tool could ingest.
- **Secrets by environment variable name only.** `RigConfig` carries the *names* of the operator-secret and DSN variables; values never enter argv, the manifest, or the transcript. The ledger DSN is not on the `psql` command line either: each read writes it to a 0600 libpq service file in a private 0700 directory that lives only for that read, with a closed allowlist of libpq keywords. Every subprocess (CLI, drift hook, psql) gets a scrubbed environment with the operator secrets, the DSN variable and all `PG*` dropped; psql keeps a minimal allowlist plus `PGSERVICEFILE`/`PGSERVICE`. The transcript records path arguments as `<path>` and admin exchanges as origin-relative routes. `RigConfig.validate()` refuses a non-loopback admin origin, a shared secret between the two actors, and the same actor name twice.
- **Transport is a Protocol.** `RigTransport` (`cli`, `cli_raw`, `admin_post`, `admin_get`, `ledger_counts`, `induce_drift`, request-log markers, `surfaces`) has one physical implementation (`PhysicalRig`: subprocess CLI, urllib admin with per-actor bearer, psql ledgers) and one fake for tests. Adding a hermetic-harness transport later is a new class, not a rewrite.
- **Drift is an operator hook.** The coordinator cannot drift itself; step 7 calls `--drift-hook` once and measures the outcome. Without the hook `PhysicalRig.induce_drift` raises with the flag named; it never sets an observation without measurement.
- **The operator surface is asked in the coordinator's own grammar.** Reason codes are `operator_*` (`^operator_[a-z0-9_]{2,56}$`), provider ids `^[a-zA-Z0-9_.-]{1,64}$`, candidate ids `^byom_[a-z2-7]{52}$`, event ids 64 hex, pending ids 32 hex, and the runner refuses to send anything outside that grammar, so a runner defect can never be recorded as a coordinator verdict. Every refusal the journey relies on is pinned to the exact status and `error.code`; the fake coordinator in the tests enforces the same validation (grammars, bound fields on approval, stale-head at approval, 404 for an unknown provider/candidate pair) and the negative tests prove look-alike refusals do not pass.
- **Every captured CLI document goes through `validate_captured_cli_document` and `assert_captured_document_redacted`**, the same guards the capture tool applies, so the manifest cannot carry a document the downstream tool would reject.

## Validation
- `make test_admission_journey_runner` (added to `test-dist`): 26/26. Full twelve-step pass with a capturable manifest; self-approval accepted, or refused with 400/401/409-no_pending look-alikes, fails step 9; illegal transition accepted, or refused with stale_head/invalid_request, fails step 11; reason codes inside the grammar and the runner's pre-wire guard; approval body and route match the coordinator (old shape is 400); withdrawal reason closed to the SPEC-047 enum; step 10 requires the GGUF candidate with no earning path; a leak in any of the six step 12 surfaces fails the run naming category and surface only; physical rig reviews only log bytes appended during the run; DSN never in psql argv (service file 0600/0700, removed after the read); both DSN forms parsed, unknown libpq keys fail closed; non-zero ledger publishes no manifest; operator-origin revocation is not drift; `--out` must be empty; config refuses non-loopback / shared secret / same actor; physical rig fails closed without the drift hook; fake ids are production-shaped and the runner refuses a `byom_unstable_` id; fake validates and routes by provider/candidate and binds approvals to the record; approval after an intervening event is `409 stale_head`; a refused duplicate that appends an event fails step 2; URL/path/IP/localhost in any surface fails step 12; transcript redacts path arguments; every subprocess gets the scrubbed environment.
- `test_byom_journey_evidence` 117/117 (three new admission-capture regressions at the `build_evidence` boundary), `test_byom_contract_lock`, `test_byom_onboarding_oracle`, `test_spec_governance`: OK.
- `scripts/check_spec_governance.py` passed; `git diff --check` clean; ASCII only; `Package.resolved` untouched. Rebased onto `origin/main` at `4c99f9d7`.

## What this PR does not do
No physical run: per the 2026-09-13 model change on #1486 the operator runs the capture in-session on operator hardware; this PR's deliverable is the audited driver. `ADMISSION_CONTRACT.typed_capture_validation` is now `True`; the admission fixture's eleven stub captures are replaced by full-shape documents from the fake-coordinator run (synthetic, but shaped by the same typed validators real captures must pass), so the evidence boundary enforces the closed schemas today. The operator's first physical run supplies the real captures. No coordinator, CLI, SPEC or schema change.

BYOM epic gates:
- Parent epic: #1453 (v0.2 slice 7)
- Slice issue: #1486
- Authority boundary: SPEC-047 network-model-admission (journey driver only; no state-machine change)
- SPEC requirements: SPEC-047-R001 through R008 as attached per step in `STEP_REQUIREMENT_IDS`
- User-facing state/copy: none
- Machine schema: none new (`macprovider.byom-journey-run.v1` already defined; `model_admission_withdraw.v1` validator added for an existing CLI document)
- Security/privacy: secrets by env-var name; loopback-only admin origin; redaction review is a hard step; no manifest on failure
- Runtime/liveness: n/a
- Identity/economics/settlement provenance: asserted, not created; money-path ledgers must stay at zero
- Rollout/rollback/compatibility: test tooling only; nothing ships in the CLI or coordinator
- Explicitly deferred: the physical run (operator-side per #1486)

{
  "behavior_change": "yes",
  "contract_change": "none",
  "issue": "https://github.com/Augustas11/macprovider/issues/1486",
  "specs": ["SPEC-047"],
  "requirements": ["SPEC-047-R001", "SPEC-047-R002", "SPEC-047-R003", "SPEC-047-R004", "SPEC-047-R005", "SPEC-047-R006", "SPEC-047-R007", "SPEC-047-R008"],
  "authority_domains": ["network-model-admission"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "scripts/tests/test_admission_journey_runner.py",
    "scripts/tests/test_byom_journey_evidence.py"
  ],
  "journeys": ["JOURNEY-NETWORK-MODEL-ADMISSION"]
}


Augustas11:
## 3-lane audit on `e37eea46` — request changes (0 CRITICAL / 4 HIGH / 4 MEDIUM)

Thanks Steven — the twelve-step structure is right, the SPEC-047 v0.1.9 alignment is clean, and your three source-confirmed corrections (offer has no replay field → duplicate is a fresh package/409; opaque endpoint refused by the CLI builder pre-coordinator; re-read the head before each decision) all hold.

The blocker is a single theme the audit kept hitting: **the fake coordinator in `test_admission_journey_runner.py` is more permissive than the real coordinator**, so these assertions pass in CI but would either fail on real hardware or record a "passed" observation the run didn't actually prove. Since the whole point of this driver is evidence a reviewer can trust, the assertions have to pin the exact status + error code, and the fake must mirror the real validation.

### HIGH
1. **Operator `reason_code`s are not in the `operator_*` namespace** — `admission_journey.py:501,513,569,572,615`. The real coordinator validates `reason_code` as `^operator_[a-z0-9_]{2,56}$`; the runner sends `experimental_disclosure`, `catalog_binding_verified`, `dual_control_settlement`, `matrix_probe`. On hardware every decision returns `invalid_request` before any step is proven — but the fake rig doesn't enforce the grammar, so the runner tests stay green. **Fix:** use closed `operator_*` reasons throughout and make the fake reject non-`operator_` reasons.
2. **`approve()` omits `expected_coordinator_event_id` and `idempotency_key`** — `admission_journey.py:370`. Both are required by the real coordinator (`phase4-coordinator/internal/ws/model_admission_operator.go:111`), so step 9 cannot complete against the real API. **Fix:** send the evaluated/pending head + a fresh approval idempotency key, and make the fake reject incomplete approval bodies.
3. **Self-approval refusal asserts `code >= 400`** — `admission_journey.py:577`. Any 4xx satisfies it (incl. the `400 invalid_request` from finding #2, or `401/404`), so dual control isn't actually proven. **Fix:** assert `code == 409` and `error.code == "dual_control_required"`; add a negative test where a same-actor approval that returns a *different* 4xx fails step 9.
4. **Illegal-transition asserts `code >= 400`** — `admission_journey.py:616`. `transition_matrix_enforced` is set true on any failure (stale head, malformed reason, auth), not the required `409 invalid_transition`. **Fix:** re-read the head, then require `code == 409` and `error.code == "invalid_transition"`; add tests proving `invalid_request`/`stale_head` do not satisfy step 11.

### MEDIUM
5. **Ledger DSN passed in `psql` argv** — `admission_journey.py:218` (`subprocess.run(["psql", dsn, ...])`). The DSN (`MACPROVIDER_JOURNEY_LEDGER_DSN`) is credential-bearing and visible via process inspection, contradicting the runbook's "secrets never in argv." (Security lane rated this HIGH.) **Fix:** pass the connection through `PGDATABASE`/libpq env or a 0600 service file; keep the DSN out of argv.
6. **Step 10 novel-candidate substitution** — `admission_journey.py:594`. If `--gguf-ref` is omitted the runner falls back to the opaque endpoint and allows `local_inventory_only`, yet still sets `earning_path_disclosure_verified=true`. **Fix:** require the novel non-catalog ref for the physical journey and require `earning_path_class == "no_earning_path_in_v0_1"` for step 10.
7. **Withdrawal reason `operator_withdrawal` is not in SPEC-047's withdrawal enum** — `admission_journey.py:550` — and the typed validator leaves the field open at `scripts/byom_journey_evidence.py:1124` (bare `require_text`). **Fix:** use a valid withdrawal reason (e.g. `provider_requested`), add a closed withdrawal-reason enum to the validator, and test rejection of unknown reasons.
8. **Step 12 redaction observations set from a thin slice** — `admission_journey.py:629` (`blob = transcript + json.dumps(status)`). `secret_field_persisted`, `raw_prompt_logged`, and `raw_completion_logged` are set false after checking only the runner transcript plus one status document — not the offer packages, coordinator events, or provider/coordinator logs the step claims to cover. **Fix:** feed the concrete log/artifact surfaces through the shared redaction scanner and only then set those observations.

### Not blocking
No CRITICAL. LOW/INFO: none.

Fix on the branch, then I'll re-run the 3-lane audit against the full diff as it will land. Once it clears 0/0/0, this gets antfleet-ops approval and merges (CI is already green). The two hardware-capture handoffs from my side (signed same-release feed set under `streamvc-autotune-static-v4`, SSH to redastare) are independent of this and unaffected.

SmtTheSE:
## Audit response: all eight addressed in `cbf60601`, plus one the audit did not name

Agreed on the theme. The fake was written from the transition matrix and dual-control rule, not from the operator surface's request validation, so it accepted what `model_admission_operator.go` refuses. Every request and verdict is now mirrored from that file, the fake enforces the same validation, and each negative test proves the runner does not accept a look-alike refusal. Runner tests 9 to 19; the other three BYOM python suites, governance and the PR declaration are green; CI re-ran green on the audited commit.

### HIGH

1. **Reason codes.** The four reasons are now `operator_experimental_disclosure` (step 5), `operator_catalog_binding_verified` (steps 6, 9), `operator_dual_control_settlement` (step 9) and `operator_matrix_probe` (step 11), asserted against `^operator_[a-z0-9_]{2,56}$` at import. `decide()` refuses any reason outside the grammar before the wire, so a runner defect can never be recorded as a coordinator verdict; the fake answers `400 invalid_request` to one, as the coordinator does. Test: `test_operator_reason_codes_are_inside_the_coordinator_grammar`.

2. **Approval body.** `approve()` now sends the closed `model_admission_decision_approve_request.v1` body: `pending_decision_id`, `expected_coordinator_event_id` set to the evaluated head the proposal returned (`coordinator_event_id` of the pending response, which the coordinator compares to `pending.EvaluatedHead`), and a fresh `idempotency_key`. Step 9 also asserts the proposal is `200` with the state unchanged and the evaluated head equal to the head the runner read. The fake refuses the old shape with `400 invalid_request`. Test: `test_approval_body_and_route_match_the_coordinator`.

   The one the audit did not name: the runner posted approvals to `/admin/model-admission/decisions/<id>`, but the coordinator only accepts `/decisions/<id>/approve` (`handleAdminModelAdmissionApprove`, `tail != "approve"` is `400`). Fixed in the same change; the fake now enforces the route and the 32-hex pending id.

3. **Self-approval verdict.** Step 9 requires exactly `code == 409 and error.code == "dual_control_required"` and that the approval event's `decided_by` differs from the proposal's. Tests prove `400 invalid_request`, `401 invalid_operator_token` and `409 no_pending_decision` on the same-actor approval all fail the step (`test_self_approval_refused_with_a_different_4xx_does_not_prove_dual_control`).

4. **Illegal-transition verdict.** Step 11 re-reads the head, requires exactly `409 invalid_transition`, and requires both the state and the `coordinator_event_id` to be unchanged afterwards. Tests prove `409 stale_head` and `400 invalid_request` do not set `transition_matrix_enforced` and publish no manifest (`test_illegal_transition_refused_for_another_reason_does_not_prove_the_matrix`).

### MEDIUM

5. **DSN in argv.** Gone from `psql`'s command line. Each ledger read writes the parsed DSN as a libpq service file, `0600` in a private `0700` directory that exists only for that read, and points `psql` at it with `PGSERVICEFILE`/`PGSERVICE`; every other `PG*` variable is dropped from the child environment. The parser accepts the `postgresql://` URI form and libpq `key=value` pairs, carries a closed allowlist of keywords (`host hostaddr port dbname user password sslmode sslrootcert application_name connect_timeout target_session_attrs`), and fails the run on anything else rather than dropping it. Tests: `test_ledger_dsn_never_reaches_psql_argv` (argv free of the DSN, user, host and password; file modes; file gone after the read), `test_libpq_parameters_accept_both_dsn_forms_and_fail_closed_on_unknown_keys`. Runbook "Secrets" section updated.

6. **Step 10 substitution.** `--gguf-ref` is required and `RigConfig.validate()` refuses a missing or duplicate reference. Step 10 reads the GGUF candidate only, requires `catalog_model_key == null` and `earning_path_class == "no_earning_path_in_v0_1"`; the opaque endpoint is no longer a fallback (it never reaches the coordinator, step 3). Test: `test_step_10_requires_the_gguf_candidate_to_have_no_earning_path`.

7. **Withdrawal reason.** The runner withdraws with `provider_requested` and asserts the document echoes it. `byom_journey_evidence.py` gains `ADMISSION_WITHDRAWAL_REASON_CODES` (the six SPEC-047 v0.1 values, mirroring `BYOMWithdrawalBuilder.reasonCodes`) and `_validate_admission_withdraw` closes `reason_code` to it. Test: `test_withdrawal_reason_is_the_closed_spec_047_enum` (valid document passes, `operator_withdrawal` is rejected by the validator and by the fake CLI).

8. **Step 12 surfaces.** The three redaction observations are set only after six named surfaces are clean: the runner log, the verbatim transcript of every CLI invocation and operator-surface exchange (`PhysicalRig` records stdout, stderr and every admin response), every captured document, the coordinator's own event listing (`GET /admin/model-admission/offers?provider_id=`), and the bytes the provider and coordinator logs gained during the run (`--provider-log`, `--coordinator-log`, both required; offsets taken at rig construction). Each surface is checked for both operator secrets, the DSN and its password, the shared `reject_secret_like_text` scanner, the synthetic probe prompt (`Reply with ok.`, the one prompt this journey causes to exist) plus chat request shapes, and chat completion shapes. A failure names the category and the surface, never the value. Test: `test_redaction_review_covers_every_surface` (six leak cases, one per category, each publishes nothing and withholds the value), `test_physical_rig_reviews_only_what_the_run_appended_to_the_logs`.

### Also in this change
Fake event ids are 64 hex and pending ids 32 hex, as on the wire, so the runner's grammar checks run against the fake too. The PR description is updated to match; the step 6 line there previously described the GGUF candidate, which step 6 does not touch, and now describes what the step asserts.

Ready for the re-run.

Augustas11:
## Re-audit R2 on `cbf60601` — request changes (0 CRITICAL / 4 HIGH / 2 MEDIUM)

Strong pass, Steven — all three lanes independently verified R1 #1–#6 genuinely resolved (exact status + error-code assertions, the fake mirroring `model_admission_operator.go`, and negative tests that reject look-alike refusals), and you caught the `/decisions/<id>/approve` route yourself. The bar isn't met yet because R2 reached the *next layer* of the same fake-vs-real theme, plus two gaps the new code left. Each finding below was temp-proved during the audit.

### HIGH
1. **The fake doesn't validate or route by `provider_id` / `candidate_id`.** It generates `candidate_id`s like `byom_mlxxcommunityx…`, but the real coordinator enforces `^byom_[a-z2-7]{52}$` in `modelAdmissionDecisionRequest.validate()` before any transition, and the fake's decision validator (`scripts/tests/test_admission_journey_runner.py:244-249`) skips provider/candidate grammar entirely. A body with `candidate_id: "c"` (or a wrong provider) gets `200` from the fake where production returns `400 invalid_request` — and nothing binds the approval to the same provider/candidate as the proposal. **Fix:** generate production-shaped fake BYOM ids; make the fake decision/approval validate provider + candidate grammar and require bound-field equality; add negative tests for malformed/mismatched provider/candidate.
2. **The fake approval commits without a stale-head check.** `test_admission_journey_runner.py:284` sets `settlement_capable` without comparing the current head to the pending record's `EvaluatedHead`; the real coordinator returns `409 stale_head` and invalidates the pending record when `head.CoordinatorEventID != pending.EvaluatedHead`. This is a money-path grant divergence — the fake grants where real hardware refuses after an intervening admission event. **Fix:** before committing the grant, compare the current fake event id to `record["head"]`; on mismatch return `409 stale_head` and delete the pending record. Add a negative test that appends an event between proposal and approval.
3. **Step-12 redaction runs only the credential-shaped subset.** `admission_journey.py:828` calls `reject_secret_like_text` but never scans for URLs, absolute paths, home paths, IP literals, or `localhost`. The transcript is guaranteed to contain the absolute `--config` path (`:240` logs `" ".join(args)`, `:479` includes `str(self.config.provider_config)`), so `http://127.0.0.1:…/admin`, `/Users/.../config.yaml`, and `localhost:…` pass Step 12 (temp-proved). **Fix:** redact transcript command arguments before storing, then run a URL/path/IP/`localhost` scanner over every Step-12 surface, with tests for URL/path leaks.
4. **The child environment leaks both operator credentials and the ledger DSN.** The libpq service-file fix strips `PG*` from `psql`'s env (`admission_journey.py:335`) but still passes every other variable through — including `MACPROVIDER_JOURNEY_LEDGER_DSN` and both operator bearer vars — and `cli_raw()` / `induce_drift()` (`:236`, `:355`) inherit the full runner env by omitting `env=`. Verified `T_OP_A`, `T_OP_B`, and `T_DSN` are present in the `psql` child env. This undermines the "service file only" secret boundary and dual-operator isolation. **Fix:** a scrubbed child-env helper that drops `operator_secret_a_env`, `operator_secret_b_env`, `postgres_dsn_env`, and `PG*`; for `psql`, add back only `PGSERVICEFILE`/`PGSERVICE` plus a minimal safe allowlist (`PATH`, `HOME`, `LANG`, `LC_*`).

### MEDIUM
5. **`typed_capture_validation=False` at the evidence pipeline** — `scripts/byom_journey_evidence.py:263`. The new `ADMISSION_WITHDRAWAL_REASON_CODES` validator exists and the runner calls it during capture, but `build_evidence(...)` still accepts a captured `model_admission_withdraw.v1` mutated from `provider_requested` to `operator_withdrawal` (temp-proved). So R1 #7 isn't closed at the actual handoff boundary. **Fix:** enable typed capture validation for `ADMISSION_CONTRACT` and add a `build_evidence` regression that mutates the withdrawal `reason_code` and expects rejection. (This is the `typed_capture_validation` flip — it's needed now, not deferred, for the validator to bite.)
6. **Step-2 "appended nothing" is not verified.** `admission_journey.py:599` checks only nonzero CLI exit + an `HTTP 409` substring, then accepts any coordinator-backed status with a non-null event id — it never snapshots the coordinator head before the duplicate offer and re-reads it after. **Fix:** read the head immediately before the duplicate, re-read after refusal, require the same `coordinator_event_id`; add a fake negative test that mutates the event while returning a 409-looking refusal.

### Not blocking
No CRITICAL. LOW/INFO: none. Verified resolved: R1 #1–#6 fully; #7/#8 at the assertion level (the remaining gaps are the pipeline typed-validation and the incomplete Step-12 scanner above).

Same loop: fix on the branch, I re-run the 3-lane audit on the full diff, and at 0/0/0 it gets antfleet-ops approval and merges. FYI the branch is now behind `main` (the #1499 SPEC-023 v0.12.0 / SPEC-005 v0.6.7 market-feeds spec merged) — a rebase onto current `main` before the next push is clean (no overlap with these files). The two hardware-capture handoffs from my side are still independent of this.

SmtTheSE:
## R2 response: all six addressed in `41c4692e`, rebased onto `main` @ `4c99f9d7`

Same method as R1: each finding checked against `model_admission_operator.go` and `model_admission.go` before the change, the fake now enforces the same rule, and a negative test proves the runner fails when the rule is broken. Runner tests 26 (from 19), evidence tests 117 (from 114); contract-lock, onboarding-oracle, spec-governance suites, `check_spec_governance.py`, the PR declaration and `git diff --check` are green; ASCII only.

### HIGH

1. **Provider / candidate validation and routing.** The fake's ids are now production-shaped (`byom_` + 52 base32, from `modelAdmissionCandidatePattern`; provider `^[a-zA-Z0-9_.-]{1,64}$`). Its decision and approval validators check both grammars (`400 invalid_request`), look the candidate up by `(provider_id, candidate_id)` and answer `404 no_offer` for an unknown pair, and an approval whose provider, candidate or evaluated head differs from the pending record is `400 invalid_request` (the coordinator's step (a), bound fields before anything else). The runner pins both grammars on the status document it reads, so a `byom_unstable_` id (unprovisioned namespace) stops the run before any decision. Tests: `test_fake_ids_are_production_shaped_and_the_runner_pins_the_grammar`, `test_fake_validates_and_routes_by_provider_and_candidate`.

2. **Stale head at approval.** The fake now performs step (e): after dual control, the candidate's current event id is compared to the record's evaluated head; on mismatch the record is deleted and the answer is `409 stale_head`, no grant. Test: `test_approval_after_an_intervening_event_is_stale_head_and_the_run_fails` (proposal, an intervening event on the same state, approval by the distinct actor is `409 stale_head`, the record is gone, the state is not `settlement_capable`).

3. **Step 12 scanner.** Every surface now goes through the shared evidence scanner minus only the DNS-shape rule (`reject_unredacted_text_except_hostname`): credential shapes, URLs, absolute and home-relative paths, IPv4 and IPv6 literals, `localhost`. So that the runner's own record cannot fail its own review for the wrong reason, the transcript stores CLI arguments with absolute and `~/` paths replaced by `<path>` (`redact_argument`) and admin exchanges as routes relative to the admin origin (no origin, no leading slash). Tests: `test_redaction_review_rejects_urls_paths_ips_and_localhost_in_any_surface` (five rules, each in a different surface), `test_physical_rig_transcript_redacts_path_arguments` (real `cli_raw` with the `--config` path; the path is absent, `--config <path>` is present, and the transcript passes the scanner). Runbook notes the consequence for the rig: the log files handed to the run must not gain a URL, path or IP literal while it runs; I checked that the admission operator surface logs `remote_addr` only on the ambiguous-bearer warning path, so a clean run at the default level is clean.

4. **Child environment.** `RigConfig.child_environment()` builds the environment for every subprocess: the two operator secret variables, the DSN variable and every `PG*` are dropped for the CLI and the drift hook (the CLI signs with its own keys and needs none of them); `psql` keeps only `PATH`, `HOME`, `LANG`, `LC_*`, `TMPDIR`, `TZ` plus `PGSERVICEFILE`/`PGSERVICE`. Tests: `test_child_environment_drops_secrets_and_pg_for_every_subprocess` (direct, and through `cli_raw` and `induce_drift` with `subprocess.run` intercepted), and `test_ledger_dsn_never_reaches_psql_argv` now also asserts the three variables are absent from the psql environment.

### MEDIUM

5. **Typed capture validation at the evidence boundary.** `ADMISSION_CONTRACT.typed_capture_validation` is `True`. The fixture's eleven stub captures (5-key status documents, 3-key withdraw) could not pass typed validation, so they are replaced by full-shape documents produced by the fake-coordinator run of the driver: synthetic, but shaped by exactly the typed validators the operator's real captures will have to pass, and every one passes `validate_captured_cli_document` today. The stale scope comments on the contract, `_digest_document` and `validate_captured_cli_document` are rewritten. Regressions through `build_evidence`: `test_admission_capture_rejects_a_withdrawal_reason_outside_the_enum` (the mutation you temp-proved, now rejected), `test_admission_capture_rejects_an_incomplete_status_document`, `test_admission_capture_rejects_a_settlement_claim_from_a_local_source`.

6. **Step 2 "appended nothing".** After the accepted offer the runner waits for the coordinator's own probe edge to land (`sandbox_probe_only`; reading before it lands would race a coordinator-origin event and produce a false failure on hardware), reads the `coordinator_event_id`, attempts the duplicate, requires `HTTP 409`, re-reads and requires the same event id. The fake gained `mutate_on_duplicate_offer`, which appends an event while answering the 409-looking refusal. Test: `test_step_2_requires_the_head_to_be_unchanged_by_the_refused_duplicate`.

### Also
Step 3's document id is `opaque-endpoint-refused-status`, matching the fixture manifest. PR description updated for the fixes and for the #1486 model change (no physical run from my side; the deliverable is this driver at 0/0/0).

Acknowledged on #1486: capture runs operator-side, nothing else needed from me for slice 7. Ready for R3.

antfleet-ops (approved):
R7 three-lane audit (code/security/architecture) converged 0 CRITICAL / 0 HIGH / 0 MEDIUM on the full fix diff; guidance/status output ground-truthed from the real modelAdmissionProviderGuidance + BYOMDiscovery.swift, validator pins next_action, label/meaning keys, ordered allowed_next_states, transition-reason nullability, and earning_path_class per coordinator state. Rebased clean onto main (55339a58). Approving for merge on green CI.

Augustas11 (approved):
Code-owner review (scripts/tests/** + Makefile): R7 three-lane audit converged 0C/0H/0M on the full fix diff; guidance/status ground-truthed from the real coordinator + CLI; rebased clean onto main (55339a58); CI fully green. Approving as code owner for merge.

PR #1496, opened 2026-09-12
Mint OpenRouter rate and demand proposals from one snapshot

### LINKED ISSUE #1484: One OpenRouter snapshot must mint rate-card and demand-rank (no small-dense $0.027 overlays)
## Premise

Price and demand must come from **one OpenRouter snapshot**. Today they do not.

Autotune on 16 GB still picks Llama 3.2 3B because:

- Llama 3.2 3B, Llama 3.1 8B, and Qwen3-8B all pay **$0.027/M** (operator small-dense lane)
- `demand_weight` is hand-typed **0.42 vs 0.45** while OpenRouter requests that day were **0.61M vs 73M**
- score is `payout × TPS × demand_weight`, so TPS wins

OpenRouter is already fetched by `scripts/openrouter_pricing_engine.py`. That tool **cannot apply**, its policy only lists four SKUs, and humans then overlay Entry 116 / P2-02 / yaml rows. That overlay is the bug.

Related: #1483 is a **safety rail** (16 GB must not pick the 8 GB onboarding SKU if 8B is eligible). This issue is the market model. Do not fold them into one PR.

## Current split (do not preserve)

| Surface | What it actually is |
|---|---|
| `scripts/openrouter_pricing_engine.py` | Snapshot + proposal. No apply. Policy: gpt-oss-20b, gemma-4-26b, nemotron-3-nano, qwen2.5-coder-32b only |
| Signed `rate-card.json` + `coordinator.yaml` `rewards.rate_card` | Parallel money tables. Small-dense trio hardcoded $0.027 (RESEARCH_227 cheapest 8B, Entry 116 3B-at-8B-parity, P2-02 Qwen3-8B match) |
| Signed `demand-rank.json` | Operator 0–1 weights, almost flat |
| Last archived snapshot | `docs/research/openrouter-snapshots/` 2026-08-10 — stale for pricing review (>48h rule in the runbook) |

Do **not** paste live OpenRouter cheapest prints 1:1. On 2026-09-10 the cheapest 3B endpoint was **$0.33/M** vs Llama 8B **$0.08/M**. Thin-book 3B would look more valuable and 3B would win harder. Liquidity-filter the book; do not add another human lane.

Evidence that scoring ignores volume: clean 1.8.123 probe on operator M3 16 GB (`/tmp/autotune-16gb-clean-20260911T113500Z.json`) — all three `completion_rate_usd_per_million_tokens=0.027`, shortage=1.0, demand 0.42/0.45/0.38, 3B wins 2:1 on TPS. Pearl jobs 40 (M4) and 37 (M5) same shape. #1483.

## Scope

One daily snapshot mints **both** signed money and demand feeds. Operator sign remains the apply gate.

### 1. Snapshot (engine)

- Map **every recommendable catalog key** in `openrouter_pricing_policy.json`, including `meta-llama/llama-3.2-3b-instruct`, `meta-llama/llama-3.1-8b-instruct`, `qwen3-8b`.
- Per mapped key: 30-day `completion_tokens`, 30-day `requests`, and **volume-filtered** paid endpoint prices (cheapest or volume-weighted median among endpoints that actually served the model). Drop `:free` and dust-volume quotes.
- That liquidity filter is the only pricing rule. It replaces small-dense / Entry 116 / P2-02 dollar overlays.

### 2. Prices (SPEC-005)

```
our_completion = floor(or_liquid_completion × (1 - undercut))
our_prompt     = floor(or_liquid_prompt × (1 - undercut))
```

Same `broad_fleet_undercut_fraction` (0.20) for all mapped rows. Coding premium stays a **policy flag on the mapping**, not a second handwritten lane.

**Signed `rate-card.json` is the only money table.** Coordinator loads that feed. Delete per-model rows from `coordinator.yaml` except `default`. Engine still has **no apply**. `catalog-release` is the sign/publish path.

### 3. Demand (SPEC-023)

From the **same** snapshot digest, write `or_completion_tokens_30d` / `or_requests_30d` into `demand-rank.json`. Stop shipping flat 0.42/0.45 weights as the scoring input.

```
eligible = hardware + hard-swap + recommendable + rate-card row
demand_share(row | mac) = tokens_30d(row) / sum(tokens_30d of eligible on this mac)
raw_score = completion_payout × measured_tps × demand_share
```

Share is **relative to this Mac’s eligible set**, not vs gpt-oss-120b. On 16 GB that set is {3B, Llama 8B, Qwen3-8B}; 8B is ~82% of those tokens, 3B ~1%.

`cold_start_floor` remains only for **new catalog keys with no OR history** (RESEARCH_229 FM-2). It must not flatten 8B vs 3B.

### 4. Release bind

Both signed JSON headers carry `source_snapshot.content_digest`. Rate card from snapshot A and demand from snapshot B is a release error. 48-hour stale rule stays. No installer live-fetch of OpenRouter. No unattended apply.

## Out of scope

- #1483 RAM-class implementation (separate PR; keep as belt)
- Auto-apply proposals
- Malibu `/v1` request counts as demand (circular; SPEC-023 Q1a unmatched keys stay intake-only)
- Qwen3.5-9B, LFM2.5, OpenRouter provider listing, BYOM v0.3
- Copying OR `$` with no liquidity filter

## Authority

SPEC-005 rate-card source of truth, SPEC-023 §4 scoring + demand-rank schema, RESEARCH_229 (OR 30-day tokens as v1 demand; coord stats later), RESEARCH_227 undercut lanes, `docs/runbooks/openrouter-pricing-engine.md` (no apply, 48h stale).

## Exit

- Policy maps all current recommendable keys.
- One snapshot compute emits a rate-card proposal **and** a demand-rank proposal bound to that digest.
- After a signed cut: coordinator has no per-model yaml prices except `default`; autotune 16 GB JSON shows Llama 8B `demand_share` ≫ 3B and 8B wins when 8B is eligible.
- 8 GB path still has an eligible paid row (3B).
- sku-econ / catalog-release tests fail if the two feeds’ snapshot digests diverge.

@erikHtoo engine mapping/compute is the contributor-shaped slice (still no apply). SPEC-005 yaml deletion and the signed cut stay operator/money-path.

erikHtoo:
Closes #1484 (engine + named market-peg ingest, proposal-only)

## What this does
- One OpenRouter snapshot mints both the rate-card proposal and demand-rank proposal.
- Both proposals share the snapshot digest; published FeedSchema-A bytes stay closed and contain no `source_snapshot` or `or_*` fields.
- Policy maps all 10 current recommendable catalog keys, including Llama 3.2 3B, Llama 3.1 8B, and Qwen3-8B.
- Numeric endpoint liquidity uses volume-weighted medians, paid `status==0` endpoints, throughput >= 1, uptime >= 0.90, and the 1M / 5%-of-model-volume floor.
- One 20% undercut applies to prompt and completion, including the coding lane; cache-hit uses the explicit policy fraction.
- Demand proposals use `total_tokens` for `or_completion_tokens_30d`; `or_requests_30d` is informational and always `0` because rankings-daily publishes no request count.
- Mapped recommendable keys absent from top-50 follow the cold-start/zero-demand demand proposal path, but rate-compute fail-closes with no proposal rather than retaining an old price.
- `catalog-release.py` has a closed unpublished `market-peg-bind.v1` validator and named four-input ingest that replays both proposals and checks published credit fields plus Score-A `demand_weight`.

## Tests
- `python -m unittest scripts.tests.test_openrouter_pricing_engine`
- `python -m unittest scripts.tests.test_openrouter_pricing_receipt`
- `python -m unittest scripts.tests.test_catalog_artifact_feed.RateCardSourceTest scripts.tests.test_catalog_artifact_feed.ReleaseSnapshotBindingTest`

## Notes
- No apply mode: the engine does not modify coordinator yaml, sign, or publish.
- MoneyTable-A yaml parity is unchanged and still owned by the existing release gate.
- First market-pegged cut must pass all four named inputs plus `market-peg-bind.json`; FeedSchema-B/Score-B remain later belts.

{
  "behavior_change": "yes",
  "contract_change": "yes",
  "issue": "https://github.com/Augustas11/macprovider/issues/1484",
  "specs": ["SPEC-005", "SPEC-023"],
  "requirements": ["SPEC-023-R008"],
  "authority_domains": ["billing-settlement-formula", "autotune-recommendation"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "python -m unittest scripts.tests.test_openrouter_pricing_engine",
    "python -m unittest scripts.tests.test_openrouter_pricing_receipt",
    "python -m unittest scripts.tests.test_catalog_artifact_feed.RateCardSourceTest scripts.tests.test_catalog_artifact_feed.ReleaseSnapshotBindingTest"
  ],
  "journeys": ["not-required"]
}


erikHtoo:
Small update after rebasing on current main: the new receipt provenance suite now passes too (18/18). I kept v5 snapshots/policies replayable so archived August evidence stays valid, while new snapshots are v6 and require request_count. Also fixed the earlier coordinator test flake was unrelated; this PR only touches scripts.

erikHtoo:
Rankings-daily works, but the row keys are only date, model_permaslug, and 	otal_tokens. 
equest_count is not emitted upstream. So the only open decision is the v1 request-count field: omit it for now, or point me at another source. The key blocker is closed.

erikHtoo:
Rebased onto current main and updated to the merged SPEC-023 v0.12.0 contract at 8884963b.

- `rankings-daily` parser is back to the real upstream shape: `{date, model_permaslug, total_tokens}`. `request_count` is no longer expected or fabricated.
- Every proposal demand row still carries `or_completion_tokens_30d`; `or_requests_30d` is always `0` because the source publishes no request count.
- Policy is `openrouter-market-feeds-v0.12.0`, one 20% undercut, and explicit `cache_hit_fraction: 0.25`.
- Snapshot endpoints now carry numeric 30-day completion tokens. Prices use the volume-weighted median with the 1M / 5%-of-model-volume liquidity floor, not a throughput proxy or cheapest print.
- Mapped recommendable keys absent from top-50 become zero-demand rows (cold-start path), not global failures.

Tests: engine suite 50/50; linked catalog-release suites 26/26. This supersedes my earlier two comments where I was still asking for a request-count decision — merged v0.12.0 resolved it.

erikHtoo:
Updated at da306e59.

- Repealed retain-on-absence: a mapped recommendable key missing from top-50 now fails the whole rate compute (no proposal). The demand proposal still carries the zero-token cold-start row.
- FeedSchema-A is closed again: published rate-card/demand/source bytes reject `source_snapshot` and `or_*`.
- Added the unpublished `market-peg-bind.v1` validator and named four-input release ingest (rate proposal, demand proposal, snapshot, policy).
- Release replay checks both proposals, engine bytes, policy bytes, snapshot digest/window, the three credit integers, and Score-A `demand_weight`.
- Preserved archived v5 receipt replay while keeping `cache_hit_fraction` mandatory for v0.12 policies.
- Updated the governance body to `SPEC-023-R008` / v0.12 and removed the stale request-count blocker text.

Focused tests pass locally: engine 50/50, receipt 18/18 (five Windows-only skips in the local symlink/commit binding cases), linked catalog-release suites 27/27.

erikHtoo:
Amended to 1ee78af8: archived v5 replay now stays a blocked-row replay instead of being reinterpreted under the v0.12 no-proposal rule. Focused engine/receipt suites pass locally.

erikHtoo:
Legacy replay fix is amended into `1ee78af8` (new SHA follows push): archived v5 evidence now preserves its original `retained` rows, blocked-row routing, and omits `source_snapshot` from the archived proposal shape. Current v6 snapshots still fail on recommendable-key absence and never emit retain-on-absence.

Focused local tests: engine 50/50, receipt replay suite passes (with the usual Windows-only skips), linked catalog suites 27/27.

erikHtoo:
Rebased onto current `main` and amended the legacy replay fix. The PR merge commit now exactly replays the archived v5 proposal as well; current v6 output still carries `source_snapshot`, while archived v5 output correctly omits it.

Focused tests after rebase: engine 50/50, receipt 18/18 locally on Windows-equivalent runs, linked catalog suites 27/27.

erikHtoo:
Ready for review at `1b70804c` (rebased on current `main`).

One correction to my earlier status comment: after the merged AC-MKT-12 wording, a mapped recommendable key absent from top-50 no longer lets rate compute complete with a zero-demand row. Rate compute fail-closes and emits no proposal; only the separate demand proposal carries zero-token cold-start rows. The PR body and tests reflect that rule.

What's covered:
- One snapshot mints both proposals; both bind to the same snapshot digest.
- Published FeedSchema-A stays closed (`source_snapshot` / `or_*` are not published).
- Liquidity uses the numeric volume-weighted median, not cheapest prints.
- `or_completion_tokens_30d` comes from rankings-daily `total_tokens`; `or_requests_30d` is informational and `0`.
- `market-peg-bind.v1` is unpublished and validates exact snapshot/policy/engine/proposal bindings.
- Release replay checks proposals, credit integers, and Score-A `demand_weight`.
- No apply mode, installer OpenRouter fetch, Malibu circular demand, or class fallback.
- Archived v5 evidence still exact-replays under legacy semantics.

All PR checks pass: engine 50/50, receipt replay, linked catalog suites 27/27, and the full CI matrix.

erikHtoo:
Final rebase commit is `0ae774a1` (after #1495 merged). Focused tests still pass: engine 50/50, receipt 18/18 with the normal Windows symlink skip, linked catalog suites 27/27. CI is rerunning on the updated tip.

Augustas11 (changes requested):
## 3-lane audit on the full diff (rebased to current main) — request changes (0 CRITICAL / 7 HIGH / 2 MEDIUM)

Thanks @erikHtoo — the structure (one snapshot → both proposals, proposal-only, closed FeedSchema-A, v5 legacy replay) is right and CI is green, but all three lanes independently found money-path defects that the current tests lock in rather than catch. These are SPEC-023 §3.3.2 / AC-MKT conformance gaps in the mint/price/gate logic. Bar is 0 C/H/M.

### HIGH
1. **Credit mint uses ROUND_HALF_UP + folds global_multiplier_ppm.** `openrouter_pricing_engine.py:1089` `int(credits_per_mtok.quantize(Decimal("1"), rounding=ROUND_HALF_UP))`, and the value at `:1085` divides by `(multiplier_ppm * usd_per_million_credits)`. SPEC-023 §3.3.2 rule 5 / AC-MKT-4 require `floor(or_liquid_usd_per_mtok × (1 − undercut) × 1_000_000 / usd_per_million_credits)` and explicitly "do not fold `global_multiplier_ppm` into the mint." An in-memory probe minted 2 where floor gives 1. **Fix:** `ROUND_FLOOR`, divide only by `usd_per_million_credits`, keep the multiplier out of the mint.
2. **Prompt price is not its own weighted median.** `:619` sorts by completion then `:628` reuses the selected endpoint's prompt price. Rule 4 / AC-MKT-3 require independent volume-weighted medians for prompt AND completion over the same eligible set. Probe: completion picked provider C's prompt 900000 while the prompt weighted median was 500000. **Fix:** compute two independent weighted medians.
3. **Illiquid mapped recommendable key still emits a proposal.** `:1351` records `no_active_priced_endpoint` as a `blocked` row and continues; `build_proposal` still returns a proposal at `:1439`. Rule 1/4 + AC-MKT-1 require the whole rate compute to fail closed (no proposal) for a v6 snapshot when any mapped recommendable row is absent/illiquid/no-provider/0-credit. **Fix:** raise before artifact write for v6; keep blocked/retained rows only for legacy v5 replay.
4. **Proposal omits `prompt_cache_hit_rate_per_mtok`.** `cache_hit_internal` is computed at `:1190` but `proposed_rates` at `:1382` emits only prompt/completion. The release validator expects the field at `catalog-release.py:789`, so the three-credit-field replay (rule 2 / AC-MKT-6) is not wired end-to-end. **Fix:** emit `prompt_cache_hit_rate_per_mtok` on every non-legacy proposal row, include it in change detection, and add tests.
5. **Market binding gate is optional.** `catalog-release.py:4195` `if market_peg is not None:` and `:4930` sets `market_args = None` when inputs are absent, so a market-pegged cut can be generated without checking snapshot digest / policy / engine bytes / proposal equality. AC-MKT-6 requires the named four-input ingest + `market-peg-bind.json` and fail-closed replay. **Fix:** require the market inputs for a market-pegged cut; add negative tests for omitted/tampered inputs.
6. **`default` row and demand source are not validated in replay.** `catalog-release.py:781` loops only over recommendable rows; it never checks the published `default` against the min-completion mapped row (AC-MKT-13 — must reject a stale `1000000` default), and `:389` still allows `macprovider_buyer_supply_deficit_v1` while `validate_market_peg` only compares `demand_weight` (AC-MKT-8 requires rejecting non-OpenRouter demand source). **Fix:** verify default minima from the replayed proposal rows and require the OpenRouter demand `source`.
7. **No 48-hour stale ranking-window cutoff.** v6 compute (`:1222`/`:1230`) checks only top-N coverage, not `ranking_window_end_date` age. Rule 6 / AC-MKT-1 + SPEC-005 §5.6 require no proposals from a snapshot older than 48h. **Fix:** compare `now` to the digest-bound `ranking_window_end_date` and fail closed for v6 > 48h.

### MEDIUM
1. **`or_*` demand fields emitted as strings.** `:1486`/`:1487` emit `or_completion_tokens_30d` / `or_requests_30d` as strings; AC-MKT-5 says both are integers, and the tests at `test_openrouter_pricing_engine.py:863/902` lock the string form. **Fix:** emit ints and assert `isinstance(..., int)`.
2. **Retained snapshot cannot replay the median/floor.** The snapshot stores only the selected endpoint's 30-day tokens (`:629`/`:636`/`:741`), not the eligible-endpoint population, so catalog-release cannot independently re-prove the volume-weighted median or the 1M / 5%-of-model-volume floor. **Fix:** retain a closed per-endpoint liquidity candidate list (or endpoint-response digests + bytes) sufficient to replay both.

The v5 legacy replay path and the closed FeedSchema-A boundary look correct and should not need changes — please keep those intact while fixing the above. Ping me when pushed and I will re-run the 3-lane audit.

erikHtoo:
Pushed at af32cc04.

- Credit mint now floors `or_liquid_usd * (1-undercut) * 1e6 / usd_per_million_credits`; `global_multiplier_ppm` is no longer folded into the mint.
- Prompt and completion are separate volume-weighted medians over the same eligible endpoint set. The snapshot retains that closed per-endpoint liquidity candidate list, so replay can reproduce the medians and 1M / 5% floor inputs.
- v6 now fail-closes with no proposals for mapped recommendable keys that are absent, illiquid, provider-empty, or zero/free-priced. Archived v5 retains its blocked/retained replay behavior.
- Non-legacy proposals carry `prompt_cache_hit_rate_per_mtok`, and change detection includes all three credit fields.
- Market ingest is explicit with `--market-pegged`, requiring all four named inputs; replay validates proposal bytes, OpenRouter demand source, default-row minima, snapshot/policy/engine digests, and ranking-window binding.
- v6 ranking windows older than 48 hours fail closed.
- Proposal demand `or_*` fields are integers, with requests always `0` because rankings-daily publishes no request count.

Focused: engine 50/50, archived receipts 18/18 (five usual Windows skips). Legacy v5 replay and closed FeedSchema-A boundaries remain intact.

erikHtoo:
Follow-up hardening at aac077e9:

- Presence of the unpublished `market-peg-bind.json` now makes the four named `--market-*` inputs mandatory, so the bind sidecar itself cannot coexist with an unbound release run.
- Added a negative test for that omitted-input case.
- Added a multi-endpoint fixture proving prompt and completion select independent volume-weighted medians, not prices from one endpoint.
- Focused suites pass: 80 engine/receipt tests plus linked rate-source tests.

Augustas11 (changes requested):
## R2 3-lane audit on the full diff (`99ad22cd..aac077e9`) — request changes (0 CRITICAL / 6 HIGH)

Thanks @erikHtoo — this is a big step from R1. Confirmed **resolved** and locked by tests: the ROUND_FLOOR mint with no `global_multiplier_ppm` fold (H1), independent prompt/completion weighted medians at fetch (H2 original reuse bug), fail-closed no-proposal on an absent/illiquid mapped recommendable row (H3), `prompt_cache_hit_rate_per_mtok` emitted + replay-compared (H4), OpenRouter-only demand `source` + `default` = min-completion mapped row (H6), and integer `or_*` demand fields (M1).

Three lanes (code / security / architecture) then independently found the following on the combined diff. Bar is 0 C/H/M.

### HIGH
1. **Median-selection crashes when prompt and completion medians come from different endpoints.** `openrouter_pricing_engine.py:631` does `next(item for item in priced if item[0] == completion and item[1] == prompt and item[5] == selected_tokens)`. `weighted_median` (`:579`) returns an existing endpoint's value, and the completion-sorted median endpoint is frequently **not** the prompt-sorted median endpoint (providers price prompt vs. completion independently), so no single row satisfies all three predicates → `StopIteration`, uncaught. The current fixtures happen to align the two, so the suite is green while a realistic multi-endpoint snapshot aborts the market-pegged compute. **Fix:** resolve the completion-median and prompt-median endpoints independently (carry each one's provider/tokens from its own `weighted_median` call); don't require one endpoint to own both.

2. **Catalog-release replay does not re-derive the median from the retained population.** `proposed_price` (`:1219`) reads the snapshot row's stored `completion_per_mtok` / `input_per_mtok`, and the `build_proposal` replay in `catalog-release.py:762` re-runs that same read — so the volume-weighted median is trusted from fetch time and never recomputed from `liquidity_filter.eligible_endpoint_liquidity`. The retained population is shape-validated (`:773`–`:790`) but never used to re-prove the median or the 1e6 / 5%-of-model-volume floor. §3.3.2 rule 2/4 require catalog-release to **independently** recompute both from the retained bytes and fail closed on mismatch; as written, a tampered stored median passes because both sides read the same trusted value. **Fix:** in replay, recompute `weighted_median` (and the floor) from `eligible_endpoint_liquidity` and byte-compare to the stored `or_liquid_*`.

3. **48-hour staleness is bypassed at the release boundary.** `catalog-release.py:764` passes `now = rate_proposal["generated_at"]` into the engine replay, so the 48h gate at `openrouter_pricing_engine.py:1278` (`now.date() − ranking_window_end_date > 2 days`) is evaluated against the proposal's **own** generation time. A snapshot/proposal generated weeks ago replays clean regardless of real release time. §3.3.2 rule 6 measures staleness against release-time now. **Fix:** keep `generated_at` for the byte-equal recompute, but add a separate staleness check of the digest-bound `ranking_window_end_date` against real wall-clock now before publishing.

4. **Zero-credit recommendable rows are not fail-closed.** `internal_rate` (`:1127`) returns `floor(value × 1e6 / usd_per_million_credits)` with no zero guard, and `proposed_price` checks `cache_hit_internal == 0` (`:1233`) but never `completion_internal == 0` or `prompt_internal == 0`. §3.3.2 rule 5 requires a 0-credit recommendable row to fail closed; a small-but-nonzero liquid price that floors to 0 credits currently ships. **Fix:** raise if `completion_internal == 0` or `prompt_internal == 0`.

5. **The market-bind negative test is broken and the suite is red.** `scripts/tests/test_catalog_artifact_feed.py:1977` `test_market_peg_bind_requires_named_market_inputs` fails with `AssertionError: CatalogError not raised`; the harness path table doesn't patch `MARKET_PEG_BIND_PATH`, so the test writes a temp sidecar while `generate()` checks the real repo path — the H5 gate never fires and is therefore untested. `python3 -m unittest scripts.tests.test_catalog_artifact_feed` = **230 run, 1 failed** at `aac077e9`. The production gate (`catalog-release.py:4200`) does look right, but it needs a test that actually exercises it. **Fix:** patch `MARKET_PEG_BIND_PATH` in the harness so the sidecar-present/inputs-absent case raises, and add direct `generate`/CLI/`verify` negative tests for omitted and tampered named inputs.

6. **Liquidity eligibility set deviates from the SPEC closed set.** `openrouter_pricing_engine.py:625` gates on `throughput < 1` and `uptime < 0.90` in addition to §3.3.2 rule 4's eligible set (`status == 0`, strictly positive paid completion price, not `:free`, volume gate). The extra gates narrow the eligible population and can move the weighted median off the SPEC-defined value. **Fix:** either drop the two extra gates, or — if they're intended as liquidity criteria — amend SPEC-023 §3.3.2 rule 4 first and cite the requirement (SPEC before code for a money-path rule).

The v5 legacy-replay path, the FeedSchema-A published-key boundary, and the sidecar exact-key/never-published enforcement look correct — please keep those intact. Ping me when pushed and I'll re-run the 3-lane audit on the full diff.

erikHtoo:
@Augustas11 pushed `f2e3f293`.

1. Medians are now resolved independently: completion and prompt each pick their own endpoint/tokens, so a provider split across the two no longer crashes.
2. Replay recomputes both weighted medians and the 1M / 5%-of-model-volume floor from `eligible_endpoint_liquidity`, and fails on mismatch.
3. Replay keeps `generated_at` for byte-equal recompute, but separately checks the digest-bound `ranking_window_end_date` against real release-time now.
4. Prompt/completion credits that floor to zero now fail closed.
5. The hermetic harness now patches `MARKET_PEG_BIND_PATH`, so the sidecar-without-inputs test actually reaches the production gate and passes. I also made the Windows-side OpenSSL probe safe so the gate can be run locally.
6. I dropped the throughput/uptime gates so eligibility matches SPEC-023 §3.3.2 rule 4 exactly. If you intended them as liquidity criteria, I think that should be a spec amendment first.

Focused tests: `test_openrouter_pricing_engine`, `test_openrouter_pricing_receipt`, `ReleaseSnapshotBindingTest`, `RateCardSourceTest`, and the fixed market-bind negative test all pass (99 tests). CI is running.

erikHtoo:
Pushed `1f7ab8c0 ` with the R2 fixes.

- Prompt and completion medians now resolve independently from each weighted-median endpoint, so different endpoints no longer crash selection.
- Replay recomputes both medians and the volume floor from \ligible_endpoint_liquidity\ using the named policy, then fails closed on mismatch.
- The 48-hour gate now checks the digest-bound \
anking_window_end_date\ against real release-time wall clock, not the proposal's own generation time.
- Zero-credit prompt/completion recommendable rates now fail closed, with a direct regression test.
- \MARKET_PEG_BIND_PATH\ is hermetic in the harness; added direct generate-CLI, verify, tampered-median, and stale-window negative tests.
- Dropped the throughput and uptime eligibility gates and pinned the SPEC rule-4 constants (\1000000\ tokens, \

Augustas11:
@erikHtoo heads up — we've been fixing this in parallel and I want to align before either of us lands it, so your in-progress work doesn't get clobbered.

On my side I ran a multi-round 3-lane audit (code / security / architecture) over the full diff and implemented fixes for the same market-replay/median finding set, reconciled onto current `main` and green (269 focused tests). I can see you've independently pushed comparable fixes (`Address PR 1496 market replay audit findings`, the median/replay/mint commits) and just merged `main` — so we've converged on the same problems from two directions. Rather than force one version over the other, let's sync on one.

Beyond the 7 HIGH in my earlier review, the audit surfaced a few more areas worth checking your branch covers — flagging them so you can confirm rather than rediscover:

1. **v6 must not demote a mapped recommendable key.** §3.3.2 rule 4 says "demotion from `recommendable` is a prior intake cut, not this compute" and partial rate cards are forbidden. The §5 hardware-fit gates (active/residency/TPS) and demand `rank<=50` were blocking `openai/gpt-oss-120b` and `qwen3-32b` (real policy) → a partial proposal. v6 should price every mapped recommendable key with the single undercut formula and fail the WHOLE compute if one is illiquid.
2. **`coding_specialist` does not select a second formula on a v0.12.0 cut** (rule 5) — but v5 legacy pricing (`min(premium_price, market_cap)` baseline/premium) MUST stay behavior-identical to base; keep them on separate helpers so the v6 change doesn't alter legacy.
3. **Eligibility = exactly rule 4**: `status==0`, positive paid prompt+completion, not `:free`, `completion_tokens >= max(1e6, ceil(0.05*model_tokens_30d))`. Throughput/uptime must not gate the median set, and the 5% floor must be a **ceiling** (an endpoint with `1_000_000` tokens at `20_000_001` model tokens should be excluded).
4. **Production must resolve the real recommendable catalog** (`phase3-binary/catalog/autotune/autotune-candidates.json`), never fall back to the test fixture.
5. **Catalog-release replay must independently re-derive** the weighted median + volume floor from the retained `eligible_endpoint_liquidity` (not trust the stored median), and 48h staleness must be checked at **release wall-clock** time in `generate()` (the `verify()` skip for historical releases is correct/by-design).
6. Minor: receipt runner should bind engine **and** runner bytes; the market-bind negative test needs to patch `MARKET_PEG_BIND_PATH` so it actually exercises the gate.

If it's useful I can hand you my reconciled branch (the exact diffs) or the per-finding audit notes — whatever gets you to a green, single version fastest. Ping me when you're at a stopping point and I'll re-run the 3-lane audit against your head and, if it's 0 C/H/M, approve + merge your version. Nice work converging on this independently.

erikHtoo:
Thanks — aligned. I rebased my head onto current `main` (`b577ca32`), which is the same commit GitHub's merge produced, so the PR head is now the single reconciled branch.

Confirmed against your six points:

1. **Mapped recommendable keys:** v6 already fails the whole compute when a mapped recommendable row is absent or illiquid; `qwen3-32b` / `openai/gpt-oss-120b` are no longer demoted by demand-rank or hardware-fit gates. The single undercut formula is kept.
2. **Legacy v5:** retains its separate `legacy` helpers and bucket/pricing path; v6 changes do not alter legacy replay.
3. **Rule 4 eligibility:** throughput/uptime are no longer eligibility gates. Now pushed `c2d93f71` changes the 5% floor to `ceil(...)` and adds the boundary regression test.
4. **Production catalog:** `c2d93f71` removes the fixture fallback, so the engine always resolves the real recommendable catalog; tests patch the production path explicitly.
5. **Replay:** already independently recomputes both medians + volume floor from retained liquidity, and generate checks 48h staleness at release wall clock (verify skip for historical releases remains by design).
6. **Runner bind + hermetic test:** runner bytes binding and the patched `MARKET_PEG_BIND_PATH` negative test are already on my side.

Validation: `scripts.tests.test_openrouter_pricing_engine` is 54/54 green, and the focused market-peg tests pass. Please re-run the 3-lane audit on `c2d93f71`; if 0 C/H/M, approve/merge your preferred single version.

Augustas11:
Superseded by #1534.

Thanks for the work here, @erikHtoo. Rather than iterate further on this branch, we're landing the audited version from our review session (#1534), which went through R2–R7 review rounds plus a 3-lane code/security/architecture audit.

The blocker on this branch: the latest commits added a `validate_policy` gate requiring the policy field `min_endpoint_completion_tokens` to equal exactly `1000000`, wired into the proposal-replay path. That rejects the historical 2026-08-10 legacy-policy replay and turns the `deploy tooling (check-deploy-config gate)` red (`test_validate_compute_exact_replay`). SPEC-023 §3.3.2 rule 4 defines `min_endpoint_completion_tokens = 1_000_000` and `floor_fraction = 0.05` as **defaults**, not mandatory exact policy fields.

#1534 applies the SPEC-correct liquidity floor `max(1_000_000, ceil(0.05 * model_tokens_30d))` via one shared helper across the engine and catalog-release replay, keeps the historical replay valid, and passes `make test-dist`. Closes the same issue (#1484).

Issue #1575, opened 2026-09-18
Fresh-Mac onboarding blocked by hard CLT/python3 dependency; installer hard-exits requiring a GUI click


Augustas11:
## Summary
On a **brand-new Mac** (no Xcode Command Line Tools yet), the installer hard-exits at the first step because it requires `python3`, which on a clean macOS is only Apple's CLT stub. This blocks the Terminal/curl path (multi-minute GUI detour + full re-run) and **fully blocks headless/SSH onboarding**.

## Behavior observed (macOS 26.4.1, fresh)
```
[macprovider-install] ERROR: This Mac needs Apple's Command Line Developer Tools to finish provider setup
(they include python3). A system installer has been opened — click Install, wait for it to finish, then
reopen Malibu.
```
- `/usr/bin/python3` exists but is the CLT **stub**; `xcode-select -p` errors; `/Library/Developer/CommandLineTools` absent.
- The installer runs `xcode-select --install` (opens a GUI dialog) and **exits rc=1**. It does **not** wait/poll for CLT to finish or auto-resume — the user must click a GUI dialog, authenticate as admin, wait ~5–10 min, then re-run the whole one-liner.

## Impact
- **Terminal/curl user:** aborts on first run; no resume.
- **Headless/SSH user (no GUI):** hard-blocked — the CLT dialog can't be clicked, and headless mode needs pre-configured passwordless sudo which a normal new host won't have.

## Gate location
`ensure_python3_usable` / `_python3_usable_or_die` (phase3-binary/dist/install.sh:7660–7681). The gate keys off the **first `python3` on PATH**; a real interpreter earlier on PATH satisfies it without CLT.

## Proof the sole blocker is python3 (not CLT itself)
Dropping a standalone `python3` (python-build-standalone) into `~/.local/bin` and prepending PATH lets the installer proceed with **no CLT and no admin** — it downloads/verifies the release and continues normally. So the CLI runtime does not need CLT; only the installer's python helpers do.

## Fix options (in leverage order)
1. **Remove the python3 dependency from install.sh** (port the few helpers to bash/awk, or bundle a small static python in the release). Removes a first-run wall for every new Mac and unblocks headless.
2. If python3 must stay, **poll/wait for CLT and auto-resume** instead of `die` rc=1 + full re-run.
3. Detect fresh-Mac + no-GUI early and print the exact non-interactive remediation up front.

## Caveat for any PATH-based workaround
[...]

PR #1582, opened 2026-09-18
Fix installer 404: paginate latest-release lookup and de-quadratic the parser (#1574)

### LINKED ISSUE #1574: install.sh 404s for all new hosts: no stable release found when >30 prereleases follow latest stable
## Summary
The public installer `curl -fsSL https://get.malibu.tech/install.sh | bash` currently **fails for every new host** with a 404. On a fresh Mac it errors:

```
[macprovider-install] ERROR: no non-prerelease macprovider-cli release (tag ^v[0-9]) found in recent GitHub Releases
[macprovider-install] Latest release:
curl: (56) The requested URL returned error: 404
[macprovider-install] ERROR: failed to download checksums.txt
```

## Root cause
`latest_release_tag()` (phase3-binary/dist/install.sh:9435) resolves "latest" by listing only the **30 most recent** releases and filtering out prereleases **client-side**:

```
api_url="https://api.github.com/repos/${GITHUB_REPO}/releases?per_page=30"
```

The recent release stream is dominated by **candidate prereleases (v1.8.124–v1.8.162, OpenRouter readiness builds)**. The latest **stable** release `v1.8.123` now sits at **non-draft index 39**, outside the 30-item window. So the client-side filter finds **0 stable `^v[0-9]` tags** → empty result → download 404.

## Evidence (today)
- `GET /releases?per_page=30` → 29 non-draft, **29 prerelease, 0 stable**.
- `v1.8.123` (latest stable) is at non-draft index **39**.
- GitHub's server-side `GET /releases/latest` (which ignores prereleases/drafts) correctly returns `v1.8.123` — the installer does **not** use it.

## Impact
- **Severity: Critical / live.** Every fresh terminal onboarding is broken right now. Only an operator who knows to pass `MACPROVIDER_VERSION=v1.8.123` can install.

## Fix options
1. **Preferred:** resolve the stable channel via `GET /releases/latest` (server-side, prerelease-aware) instead of `per_page=30` + client filter.
2. Or paginate / raise `per_page` until a stable `^v[0-9]` tag is found (bounded).
3. Stop candidate builds from burying stable: mark readiness candidates as **drafts** (excluded from the list) rather than prereleases, or publish them under a separate mechanism.
4. Add a scheduled **consumer-surface parity alarm** that fails loudly when the installer resolves no stable release (catches this within minutes, not when a user reports a 404).

## Repro
On any Mac with no `MACPROVIDER_VERSION` set:
```
curl -fsSL https://get.malibu.tech/install.sh | bash
```

---
_Filed from an end-to-end provider-onboarding test on a fresh Mac Studio (M3 Ultra, 256 GB, macOS 26.4.1)._

Augustas11:
## Summary

`curl -fsSL https://get.malibu.tech/install.sh | bash` currently **404s for every new host**. `latest_release_tag()` resolved "latest" from a single `/releases?per_page=30` page and filtered prereleases / non-`^v[0-9]+\.[0-9]+\.[0-9]+$` tags client-side. The recent release stream is a long run of consecutive candidate prereleases (OpenRouter readiness builds), so **zero stable tags land in the first 30** → `die 3` → `checksums.txt` download 404.

Verifying the fix surfaced a **second, coupled defect**: the awk parser was **O(n²)** and hung on realistic pages. Both are fixed here in one landing patch.

## What changed

**1. Paginate the latest-release lookup** — `phase3-binary/dist/install.sh`
- Walk `per_page=100&page=N` newest-first until a stable tag is found, an empty page ends the list, or a 10-page (1000-release) cap trips (then `die 3`).
- `/releases/latest` is still deliberately avoided, and the strict `^v[0-9]+\.[0-9]+\.[0-9]+$` + non-prerelease filter is preserved, so an unrelated release under the same repo (e.g. `macprovider-verify` under `verify-vX.Y.Z`) can't hijack the installer.
- Ordering invariant: GitHub returns releases newest-first within and across pages, and the parser keeps only the first match, so the first match walking pages ascending is always the newest stable.

**2. De-quadratic the JSON parser** — same function
- The previous parser scanned the concatenated page with `substr(data, i, 1)` and re-evaluated `length(data)` per iteration — **O(n²)** in the BSD awk shipped on macOS. A `per_page=100` page is multiple MB (each release carries its full asset list), so parsing hung for minutes (measured 320KB = 80s; the full 2.5MB page never completed). This also afflicted the old `per_page=30` path today (~3 min), matching the **"stuck at Starting installer…"** reports.
- Replaced with a single-pass, string-aware streaming tokenizer (per-line `split()`, O(1) indexing, no long-string accumulation). The full 2.5MB page now parses in **~0.6s**. Hijack-safety is preserved: braces/brackets/colons/commas or fake `tag_name`/`prerelease` text inside any string value (e.g. a release body) can't be mistaken for structure; nested author/assets/reactions fields are ignored.
- The parser prints the first match then **drains stdin to EOF** rather than `exit`ing, so the `printf %s | awk` producer never takes SIGPIPE (which under `set -euo pipefail` returned 141 and aborted the installer *after* the correct tag was found).

**3. Regression tests** — `scripts/test-install-version-pin.sh`
- Page-aware curl mock; new cases 23–28: pagination past page 1, cross-page and in-page newest-first ordering, empty-page early stop, body-injection hijack, nested-flag ignore, large realistic-page O(n) guard, and early-match SIGPIPE guard. Full suite: **106 checks**.

## Verification
- **Live, end-to-end** against the unauthenticated GitHub API: the patched function resolves **v1.8.123**, whose `checksums.txt` returns HTTP 200.
- `make test-dist` green; `bash -n` clean; shellcheck clean in the changed region.
- **Three-lane codex audit** (code / security / architecture) over the full fix diff: **0 CRITICAL / HIGH / MEDIUM**. (Round 1 code lane caught the SIGPIPE HIGH; fixed and re-audited to APPROVE.)

## Follow-ups (not in this PR)
- The `get.malibu.tech/install.sh` static copy is republished manually after merge (per the curl-channel parity runbook) — this fix only lands in `dist/install.sh`.
- **Recommend** a scheduled consumer-surface parity/health alarm that fails loudly when `install.sh` resolves no stable release (or its `checksums.txt` is unreachable). Flagged by the architecture audit; worth its own task.
- Issue #1575 (fresh-Mac CLT/python3 hard-exit) is **separate** and intentionally not bundled here.

{
  "issue": "https://github.com/Augustas11/macprovider/issues/1574",
  "behavior_change": "yes",
  "contract_change": "none",
  "specs": ["SPEC-003"],
  "requirements": ["SPEC-003-R002"],
  "authority_domains": ["provider-onboarding-identity"],
  "arbitration": ["CODE_BUG"],
  "tests": ["scripts/test-install-version-pin.sh", "Makefile:test-dist"],
  "journeys": ["not-required"]
}

Fixes #1574


antfleet-ops (approved):
Approving on behalf of review: installer latest-release pagination + O(n) parser fix for #1574. CI green, three-lane codex audit 0 C/H/M, governance declaration valid, verified end-to-end against live GitHub API (resolves v1.8.123).

PR #1596, opened 2026-09-19
Keep Pi streaming tool calls concat-safe like OpenRouter

Augustas11:
## Summary
- Live Pi vs OpenRouter: Malibu Qwen3 Coder is the right model (plain chat returns `pong`). Pi always streams; Malibu currently emits complete `{}` tool arguments, which Pi treats as a finished bash call. OpenRouter concatenates `{"command":"echo hello"}`.
- Coordinator now asks the current CLI for a non-stream completion when the buyer streams with tools, then synthesizes concat-safe SSE. `tool_choice=required` is rewritten to `auto` so agents do not get a 502.
- CLI XML emitter holds arguments until `</function>` so later CLI releases stay concat-safe. Tests cover the Pi extra-field body, empty-object JSON contract, and tool-result replay.

## Test plan
- [x] Live: Malibu non-stream/stream hello 200; stream+tools concat=`{}`; stream+tools+required 502; OpenRouter same Pi body concat=`{"command":"echo hello"}`
- [x] `go test ./internal/buyer -run 'TestStreamingTools|TestPiShaped|TestToolChoice|TestEmptyObject|TestPiMultiTurn'`
- [x] `swift test --filter NativeToolCallStreamEmitterTests`
- [x] `python3 scripts/check_spec_governance.py --base-ref origin/main`
- [ ] After merge: Pearl coordinator hotfix, then re-run live Pi tool turn

SPEC-GOVERNANCE-DECLARATION-BEGIN
{
  "schema_version": "spec-pr-governance-v1",
  "behavior_change": "yes",
  "contract_change": "yes",
  "specs": ["SPEC-018"],
  "requirements": ["SPEC-018-R002"],
  "authority_domains": ["agentic-tool-calling"],
  "arbitration": ["CODE_BUG"],
  "tests": ["phase4-coordinator/internal/buyer/streaming_tool_materialize_test.go:TestPiShapedStreamingToolRequestIsAdmittedAndMaterializesArgs", "phase3-binary/Tests/macprovider-cliTests/NativeToolCallStreamEmitterTests.swift:testFunctionXMLDoesNotEmitEmptyObjectBeforeClose"],
  "journeys": ["not-required"]
}
SPEC-GOVERNANCE-DECLARATION-END

PR #1597, opened 2026-09-19
Fix paged-KV attach gates so SPEC-038/039 continuous batching can engage on real MoE hardware

Augustas11:
## Summary

Fixes the gates that prevented SPEC-038/039 continuous-batching paged-KV attach from ever engaging on a live provider (Qwen3-Coder-30B-A3B on an M3 Ultra), plus the diagnostics needed to observe attach on-device. Default-off/inert; no config flip enables this — attach still requires a genuine on-device measurement.

### The two attach blockers (both inspection-provable, found via an adversarial verificator + source trace)

1. **Metallib path gate — the actual first failure on the live box.** `PagedKVMetallibGate.candidatePaths` probed only `default.metallib` / `mlx-swift_Cmlx.bundle/default.metallib`, but the release ships `mlx.metallib` and `mlx-swift_Cmlx.bundle/Contents/Resources/default.metallib` — the names `scripts/check-tier2-provider-artifact.sh:130` accepts. The gate matched neither on a real install, so `measurePagedKVRuntime` always returned nil and attach fell back to `paged_fallback_metallib`. Now probes the shipped names (legacy names kept as harmless fallbacks).
2. **Batched-decode attention mask — a real correctness bug the isolation probe catches.** `PagedKVBatchLayerCache.makeMask` passed pre-update per-row offsets as `createCausalMask`'s `lengths`; since it masks key `j` unless `j < lengths[b]` and row `b`'s own current token sits at key position `offset_b` post-update, unequal-length rows masked each row's own current token → the batched forward diverged from serial. Now passes post-update `lengths = offset_b + n`.

### Also in this PR
- **Tagged `[paged-kv]` diagnostics**: turns 8 silent nil-sites + 2 probe catch-blocks into per-gate stderr reasons, so one on-box load names the exact failing gate. Silent when the feature is off; EPIPE-safe.
- **Universal batched-isolation gate (from audit)**: the batched cross-row isolation probe now gates attach for **all** paging-eligible families (dense + MoE), not only MoE — closing the SPEC-038 FR-CB6 hole where a dense model could attach on the single-row parity probe alone and then serve an unproven batched path. `moeDispatchProven` stays MoE-specific.
- **Cache-class fix** (`forceSimpleKV`): the attach probe classifies the uncapped cache class, matching what the paged serve path builds; genuine windowed models still fail safe (SPEC-039 v0.1.1 FR-PKV12/AC-13 clarification).
- New env-gated real-model regression test (`MACPROVIDER_RUN_PAGED_MOE_ISOLATION=1`, Qwen3-8B) exercising the batched shared-forward isolation path.

### Testing
- `swift build` green; `swift test --filter "PagedKV|KVConversationColdTier"` → 102 tests, 0 failures.
- Local `swift test` cannot run real-GPU/MLX tests (mlx-swift Metal lib isn't compiled by plain `swift build`); the new probe test and AC-1/2/3 parity run in CI + on-device.
- `python3 scripts/gen_spec_index.py --check` → ok.

### Audit
3-lane review (code / security / architecture) on the full combined fix diff: **0 CRITICAL / 0 HIGH / 0 MEDIUM.** Carried LOW/INFO: metallib SHA fingerprint fidelity (attach stays proof-gated); the `makeMask` shape `assert` is release-stripped (future n>1 guidance); dense models incur a load-time probe only when the feature is opt-in enabled.

{
  "behavior_change": "yes",
  "contract_change": "yes",
  "specs": ["SPEC-038", "SPEC-039"],
  "requirements": ["SPEC-038-R006", "SPEC-039-R008", "SPEC-039-R012"],
  "authority_domains": ["paged-kv-attention", "continuous-batching-serving"],
  "arbitration": ["CODE_BUG", "SPEC_BUG"],
  "tests": [
    "phase3-binary: swift test --filter PagedKV|KVConversationColdTier (102 tests, 0 failures)",
    "phase3-binary: testBatchedSharedForward_InputIsolation_RealModel (env-gated MACPROVIDER_RUN_PAGED_MOE_ISOLATION=1, Qwen3-8B; runs in CI/on-device)",
    "scripts/gen_spec_index.py --check"
  ],
  "journeys": ["not-required"]
}


antfleet-ops (approved):
Approving: 3-lane audit (code/security/architecture) clean at 0 C/H/M on the full combined fix; 15/15 required checks green. The single red `check` is the pre-existing inherited req[16] (agentic-tool-calling, from #1596's squash), advisory and unrelated to this SPEC-038/039 change.

PR #1608, opened 2026-09-19
SPEC-038 FR-CB6: accept batched-vs-serial numeric ties so MoE continuous batching can attach

Augustas11:
## Summary

The SPEC-038/039 continuous-batching MoE isolation probe (`runMoEInputIsolationProbe`) required **bit-exact** greedy-argmax equality between each batched row and a serial reference. That is unachievable for MoE: a batched shared forward and a serial forward differ only in floating-point **accumulation order** (batched matmuls + MoE expert routing vs a single row), which flips greedy argmax between near-tied tokens. So the probe fail-closed on every MoE model regardless of correctness — **no MoE model could ever attach** continuous batching.

FR-CB6 already requires the batched output to match serial only **"within the accepted numerical tolerance"**, not bit-exactly. This PR conforms the probe to its own SPEC.

### On-device evidence (Qwen3-Coder-30B-A3B, M3 Ultra)
- Single-row paged-gather parity: **byte-exact** (3072/3072 gather calls) — the SPEC-039 gate, unchanged and still exact.
- Batched isolation, before: `crossRowDivergences=1` — row A returned **its own serial top-2** token `11201`, `0.25` logits behind its serial top-1 (`17.625` vs `17.375`); **not** a leak (≠ row B's token).
- Batched isolation, after this fix: **`proven=true, crossRowDivergences=0`** — the own-distribution near-tie is accepted; the leak guard still rejects any cross-row token.

### Change
- **SPEC-038 FR-CB6 (v0.2 → v0.2.1)**: makes the pre-existing "accepted numerical tolerance" concrete — a row's batched greedy token is conformant iff it is the row's **own** serial argmax, OR its **own** serial runner-up within a fixed small logit tolerance of its serial argmax; any other token, and **always** another row's serial argmax (a leak), is a divergence. Explicitly scoped to the batched-vs-serial comparison; **does not** relax the SPEC-039 single-row paged-gather parity gate (same accumulation order → stays exact).
- **Probe**: pure predicate `batchedTokenIsConformant(...)` + `SerialReference{top1,top2,logits}`. The exact-match and **leak guard run before** the tolerance, so admitting numerical ties never admits a cross-row leak. `batchedArgmaxLogitTolerance = 1.0` (observed ties are ~0.25; a leaked other-row token lands many logits below a row's own argmax).
- **Tests**: `PagedKVBatchedToleranceTests` pins exact-match, own near-tie within/beyond tolerance, out-of-top-2, the `gap == tolerance` boundary, out-of-vocab fail-closed, and the decisive leak case (other row's token, also this row's runner-up within tolerance → still rejected).

### Testing
- `swift build` green; `PagedKVBatchedToleranceTests` (8) pass; 159 fake-model paged-KV tests pass, 0 regressions; `gen_spec_index.py --check` green.
- On-device confirm on the real 30B (above) via a throwaway instrumented build (not in this diff).
- The env-gated real-model isolation fixture exercises the live path in CI.

### Audit
3-lane review (security / code / architecture) at **0 CRITICAL / 0 HIGH / 0 MEDIUM**. Security confirmed cross-row isolation is preserved (leak guard is exact and tolerance-independent). Carried LOW/INFO (non-gating):
- Leak guard checks a single "other row" — correct for the fixed 2-row probe; revisit if ever generalized to N>2 rows.
- Tolerance `1.0` is ~4× the observed `0.25` tie; could tighten to `0.5`. Not an isolation risk (leak guard is tolerance-independent); a wider bound only admits more genuine ties.
- `top1` derived via `argSort` last-entry vs `argMax` first-index could differ only on an exact fp tie over the vocab (effectively nil, absorbed by the tolerance branch).

{
  "behavior_change": "yes",
  "contract_change": "yes",
  "specs": ["SPEC-038"],
  "requirements": ["SPEC-038-R006"],
  "authority_domains": ["continuous-batching-serving"],
  "arbitration": ["CODE_BUG"],
  "tests": [
    "phase3-binary: swift test --filter PagedKVBatchedToleranceTests (8 cases incl. leak worst-case)",
    "phase3-binary: 159 fake-model paged-KV tests (PagedKVRuntimeBridge/KVConversationColdTier/ServingKnobs/PagedKVEngine), 0 regressions",
    "scripts/gen_spec_index.py --check"
  ],
  "journeys": ["not-required"]
}


antfleet-ops (approved):
Approving: 3-lane audit (security/code/architecture) clean at 0 C/H/M on the full diff. Faithful clarification of FR-CB6's pre-existing numerical-tolerance clause; cross-row isolation preserved (leak guard is exact + tolerance-independent); SPEC-039 single-row parity stays exact. Fix confirmed on-device (Qwen3-Coder-30B: batched-isolation proven=true, crossRowDivergences=0).

antfleet-ops (approved):
Re-approving after rebase onto latest main (picks up #1609 CoordinatorClient changes; resolves the stale-base merge artifact). Fix diff unchanged (probe + tests + SPEC-038 v0.2.1); 3-lane audit 0 C/H/M stands; on-device proven=true confirmed.

